PR #10, opened 2015-05-08
Add distance label while drawing occurs for immediate feedback.

ghutchis:
Gives the bond distance between the atoms.

Change-Id: Ia26ea632a4a32fd63d6c495f7b74206ca60949d5


cryos:
This looks great!

PR #15, opened 2015-05-09
Add gaussiancube reader to quantumio, initial commit

barrymoo:
Added gaussiancube reader to quantumio. I couldn't find an example where you take a <code>Core::Cube</code> and render the isosurface. It looks like this is done through <code>Molecule::addCube()</code>, but it isn't obvious to me how this works. Do I use <code>addCube</code> to generate the Cube and then populate the object?

Additionally, it would be great to get some feedback on my code style and pull request. I am excited to work on this project! @cryos @ghutchis

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,0 @@
-CMakeLists.txt.user*
-*.pro
-*.pro.user
--- a/.gitignore
+++ b/.gitignore
@@ -1,1 +1,3 @@
 CMakeLists.txt.user*
+*.pro
+*.pro.user
--- a/avogadro/quantumio/gaussiancube.cpp
+++ b/avogadro/quantumio/gaussiancube.cpp
@@ -64,21 +64,21 @@ bool GaussianCube::read(std::istream &in, Core::Molecule &molecule)
   std::string cubeName = line;
 
   // Read and skip field title (we may be able to use this to setCubeType in the future)
-  getline(in, line);   
+  getline(in, line);
 
   // Next line contains nAtoms and m_min
   in >> nAtoms;
   for (auto i = 0; i < 3; ++i)
     in >> min(i);
-  getline(in, line); //capture newline before continuing 
+  getline(in, line); //capture newline before continuing
 
   // Next 3 lines contains spacing and dim
   for (auto i = 0; i < 3; ++i) {
     getline(in, line);
     line = Core::trimmed(line);
     list = Core::split(line, ' ');
-    dim(i) = Core::lexicalCast<int>(list[0]);  
-    spacing(i) = Core::lexicalCast<double>(list[i + 1]);  
+    dim(i) = Core::lexicalCast<int>(list[0]);
+    spacing(i) = Core::lexicalCast<double>(list[i + 1]);
   }
 
   // Geometry block
@@ -95,9 +95,9 @@ bool GaussianCube::read(std::istream &in, Core::Molecule &molecule)
     a.setPosition3d(pos);
   }
 
-  // Render molecule  
+  // Render molecule
   molecule.perceiveBondsSimple();
-  
+
   // Get a cube object from molecule
   Core::Cube *cube = molecule.addCube();
 
--- a/avogadro/quantumio/gaussiancube.cpp
+++ b/avogadro/quantumio/gaussiancube.cpp
@@ -48,30 +48,29 @@ std::vector<std::string> GaussianCube::mimeTypes() const
 
 bool GaussianCube::read(std::istream &in, Core::Molecule &molecule)
 {
-  // Using the existing cube class
-  Core::Cube cube;
-
   // Variables we will need
   std::string line;
   std::vector<std::string> list;
 
   int nAtoms;
-  Avogadro::Vector3 min, spacing;
+  Avogadro::Vector3 min;
+  Avogadro::Vector3 spacing;
   Avogadro::Vector3i dim;
 
   // Gaussian Cube format is very specific
 
   // Read and set name
   getline(in, line);
-  cube.setName(line);
+  std::string cubeName = line;
 
   // Read and skip field title (we may be able to use this to setCubeType in the future)
   getline(in, line);   
 
   // Next line contains nAtoms and m_min
   in >> nAtoms;
-  for (auto i = 0; i < 3; ++i) in >> min(i);
-  getline(in, line);
+  for (auto i = 0; i < 3; ++i)
+    in >> min(i);
+  getline(in, line); //capture newline before continuing 
 
   // Next 3 lines contains spacing and dim
   for (auto i = 0; i < 3; ++i) {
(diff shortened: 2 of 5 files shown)


barrymoo:
Additionally, cube object now comes from <code>Molecule::addCube</code> (not <code>Core::Cube</code>). Next step would be to render the isosurface from the data, but I am unsure how to do this correctly.

cryos:
I will try to take a look at this soon. We need to hook up a little extra logic (easy enough to do) to show all available cubes, and create a surface mesh from one. You should call addCube, and then populate that with data.

.gitignore
```diff
@@ -1 +1,3 @@
 CMakeLists.txt.user*
+*.pro
```
cryos: Can we remove these changes, I am not sure we need the first line even, developers can add development environment specific excludes to their global configuration for git, rather than every project.

avogadro/quantumio/gaussiancube.cpp
```diff
@@ -0,0 +1,123 @@
+/******************************************************************************
+
+  This source file is part of the Avogadro project.
+
+  Copyright 2010 Geoffrey R. Hutchison
```
cryos: You should change this to your name, and the current year.

avogadro/quantumio/gaussiancube.cpp
```diff
@@ -0,0 +1,123 @@
...
+}
+
+bool GaussianCube::read(std::istream &in, Core::Molecule &molecule)
+{
+  // Variables we will need
+  std::string line;
+  std::vector<std::string> list;
+
+  int nAtoms;
+  Avogadro::Vector3 min;
```
cryos: You are in the Avogadro namespace, and so no need to add it explicitly here.

avogadro/quantumio/gaussiancube.cpp
```diff
@@ -0,0 +1,123 @@
...
+  // Read and set name
+  getline(in, line);
+  std::string cubeName = line;
+
+  // Read and skip field title (we may be able to use this to setCubeType in the future)
+  getline(in, line);
+
+  // Next line contains nAtoms and m_min
+  in >> nAtoms;
+  for (auto i = 0; i < 3; ++i)
```
cryos: We don't use auto, this is C++11 which is not a hard dependency yet. We should possibly rethink it, but not just for auto in loops.

avogadro/quantumio/gaussiancube.cpp
```diff
@@ -0,0 +1,123 @@
...
+    in >> min(i);
+  getline(in, line); //capture newline before continuing
+
+  // Next 3 lines contains spacing and dim
+  for (auto i = 0; i < 3; ++i) {
```
cryos: Auto

avogadro/quantumio/gaussiancube.cpp
```diff
@@ -0,0 +1,123 @@
...
+    getline(in, line);
+    line = Core::trimmed(line);
+    list = Core::split(line, ' ');
+    dim(i) = Core::lexicalCast<int>(list[0]);
+    spacing(i) = Core::lexicalCast<double>(list[i + 1]);
+  }
+
+  // Geometry block
+  Avogadro::Vector3 pos;
+  for (auto i = 0; i < nAtoms; ++i) {
```
cryos: Auto

avogadro/quantumio/gaussiancube.cpp
```diff
@@ -0,0 +1,123 @@
...
+    getline(in, line);
+    line = Core::trimmed(line);
+    list = Core::split(line, ' ');
+    short int atomNum = Core::lexicalCast<short int>(list[0]);
+    Core::Atom a = molecule.addAtom(static_cast<unsigned char>(atomNum));
+    for (auto j = 2; j < 5; ++j)
```
cryos: Auto

avogadro/quantumio/gaussiancube.cpp
```diff
@@ -0,0 +1,123 @@
...
+
+  // Get a cube object from molecule
+  Core::Cube *cube = molecule.addCube();
+
+  // Cube block, set limits and populate data
+  cube->setLimits(min, dim, spacing);
+  std::vector<double> values;
+  // push_back is slow for this, resize vector first
+  values.resize(dim(0) * dim(1) * dim(2));
+  for (auto i = 0; i < values.size(); ++i)
```
cryos: Auto

cryos:
I didn't test yet, but it looks good at first glance. We should avoid auto, especially for simple intrinsic types in loops as that would create a hard dependency on C++11.

barrymoo:
@cryos, everything should be set now. I was experimenting with Qt Creator, but decided I like just like `vim` better (hence the `.gitignore` stuff). I guess the next thing would be to display the cubes. I was digging around in the code before and I couldn't decide where the code needed to become aware of the cubes. Anyway, whenever you have time to check it out. Thanks!

avogadro/quantumio/gaussiancube.cpp
```diff
@@ -0,0 +1,123 @@
+/******************************************************************************
+
+  This source file is part of the Avogadro project.
+
+  Copyright 2015 Barry E. Moore II
+  Copyright 2013 Kitware, Inc.
```
cryos: Sorry to be a pain, you don't need to add Kitware if the entire file was written by you. As a side note I have been reading up a little on this, and we may want to consider removing these lines - all of this information is available (and properly tracked) in the git history. I am a fan of minimal headers, source releases would lack it, but can still refer to the git repository (with tagged releases).

avogadro/quantumio/gaussiancube.cpp
```diff
@@ -0,0 +1,123 @@
...
+    getline(in, line);
+    line = Core::trimmed(line);
+    list = Core::split(line, ' ');
+    dim(i) = Core::lexicalCast<int>(list[0]);
+    spacing(i) = Core::lexicalCast<double>(list[i + 1]);
+  }
+
+  // Geometry block
+  Vector3 pos;
+  for (unsigned int i = 0; i < nAtoms; ++i) {
```
cryos: This should be an int, or nAtoms should be an unsigned int.

avogadro/quantumio/gaussiancube.h
```diff
@@ -0,0 +1,66 @@
+/******************************************************************************
+
+  This source file is part of the Avogadro project.
+
+  Copyright 2015 Barry E. Moore II
+  Copyright 2013 Kitware, Inc.
```
cryos: Same as above, Kitware line is not necessary.

barrymoo:
@cryos Not a pain at all. At this stage `nAtoms` should really be an `unsigned int`. I believe a cube file in the "wild" can have a negative `nAtoms`, but it is non-standard.

cryos:
Great, this looks ready to merge. I would like to add a cube to the avogadrodata repository, and add a test reading the data file. We can work on this in a follow up commit, and avogadro/qtplugins/quantumoutput/ is where most of the code handling QM is right now. That is where we generate a cube, and then isosurface it. We can generalize there to look at existing cubes, and we probably need to think about how we can expose cubes in the GUI, I feel like we never got that right in Avogadro 1 either (they were quite hidden, often generated, but not always).

barrymoo:
Perfect. I will dig into the code sometime soon and start a new topic branch.

PR #19, opened 2015-06-04
Initial Render Gaussian Cube

barrymoo:
@cryos Hey Marcus, I had some today to implement the cube visualization piece. It's not quite done, but I wanted to submit a pull request early because I know you are very busy. If you have any time to make some comments, I would greatly appreciate it. Currently, I need to figure a few things:

* When I switch to `Cube` in the `surfaceCombo`, `moCombo` is grayed out
* Resolution means nothing in the `Cube` context, I suppose I should gray it out

I also found a units bug in `gaussiancube.cpp`. Gaussian cubes are in bohr units by default; I converted the molecular coordinates, but not the `min` and `spacing` of the cube. Does Avogadro2 just assume everything is in Angstrom units (or convert to it?)

### DIFF
--- a/avogadro/qtplugins/quantumoutput/quantumoutput.cpp
+++ b/avogadro/qtplugins/quantumoutput/quantumoutput.cpp
@@ -254,11 +254,9 @@ void QuantumOutput::calculateFinished()
   if (!m_cube)
     return;
 
-  // Barry: I have no clue what disconnect does, but if statement
-  // needed to not crash the cube calculation
-  if (!&m_concurrent->watcher()) {
+  // if statement protects from disconnecting from NULL
+  if (!&m_concurrent->watcher())
     disconnect(&m_concurrent->watcher(), 0, 0, 0);
-  }
 
   if (!m_mesh1)
     m_mesh1 = m_molecule->addMesh();
--- a/avogadro/qtplugins/quantumoutput/quantumoutput.cpp
+++ b/avogadro/qtplugins/quantumoutput/quantumoutput.cpp
@@ -103,10 +103,6 @@ QStringList QuantumOutput::menuPath(QAction *) const
   return path;
 }
 
-// Barry Moore: modifying function to allow cubes
-// to be readily apparent. If basisSet() we 
-// can set m_actions[0,1,2]->setEnabled, but 
-// with cubes only m_actions[2]->setEnabled
 void QuantumOutput::setMolecule(QtGui::Molecule *mol)
 {
   bool isQuantum(false);
--- a/avogadro/qtplugins/quantumoutput/quantumoutput.cpp
+++ b/avogadro/qtplugins/quantumoutput/quantumoutput.cpp
@@ -103,16 +103,26 @@ QStringList QuantumOutput::menuPath(QAction *) const
   return path;
 }
 
+// Barry Moore: modifying function to allow cubes
+// to be readily apparent. If basisSet() we 
+// can set m_actions[0,1,2]->setEnabled, but 
+// with cubes only m_actions[2]->setEnabled
 void QuantumOutput::setMolecule(QtGui::Molecule *mol)
 {
   bool isQuantum(false);
   if (mol->basisSet()) {
     m_basis = mol->basisSet();
     isQuantum = true;
+    m_actions[0]->setEnabled(isQuantum);
+    m_actions[1]->setEnabled(isQuantum);
+    m_actions[2]->setEnabled(isQuantum);
   }
-  m_actions[0]->setEnabled(isQuantum);
-  m_actions[1]->setEnabled(isQuantum);
-  m_actions[2]->setEnabled(isQuantum);
+  else if (mol->cubes().size() != 0) {
+    m_cubes = mol->cubes();
+    isQuantum = true;
+    m_actions[2]->setEnabled(isQuantum);
+  }
+
   m_molecule = mol;
 }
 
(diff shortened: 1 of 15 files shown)


cryos:
@barrymoo you should grey out resolution etc if they are not applicable. We assume everything is in Angstrom, you should convert to Angstroms to keep everything consistent.

barrymoo:
@cryos, all set. I will do some more testing tomorrow. I am not quite sure how the code handles multiple Meshes at this point (or Cubes for that matter). I will also fork and submit cube data/test in `avogadrodata`.

barrymoo:
`MeshGenerator` currently only works with square cubes, fixing now.

cryos:
A few comments in line, the basics all seem to work well here, and it is great to add back in cube rendering (which was present in the 1.x series). I think the dialog needs a little work to generalize, but it was already in need of improvement. Please don't add your name to comments - if we want to know who wrote something we can look in git blame, just add a brief explanation if you feel it is needed.

Some of these comments look more appriopriate for the log message, i.e. "modifying blah to do x" ceases to be relevant in code once this is merged, but is appropriate for the body of your commit message.

cryos:
There are also other sources of cubes, so simply calling them cubes in the user interface is probably more appropriate as there is nothing particularly special (other than the format) about Gaussian cubes.

barrymoo:
First, thanks again for the good notes here, and elsewhere, for future development.

Cleaning up the SurfaceDialog
===
Could we use a QDockWidget instead of the QDialog for the SurfaceDialog? I don't really know what options we have with Qt unfortunately. It would be nice to have the SurfaceDialog integrated into the GUI itself (instead of a separate QDialog).

More dynamic implementation
===
Inside the QDialog (or QDockWidget), add a list of all meshes available. When clicked, it would display the selected Mesh to the screen. Obviously, for an MO there are two Meshes made per orbital, therefore instead of `std::vector<Mesh *> m_meshes` we could have `std::vector<Isosurface *> m_isosurface` which is just two meshes. `m_meshes` could be used for densities to reduce the memory footprint.

cryos:
We could add support for making it dockable, I tend to work with multiple monitors, and so prefer dialogs (and think this to a degree is a user preference thing). It would be nice to add more support for managing/working with meshes/cubes once calculated, these things have been a little hidden in the past and are not easily saved either.

avogadro/qtplugins/quantumoutput/quantumoutput.cpp
```diff
@@ -221,13 +241,24 @@ void QuantumOutput::calculateElectronDensity(float isoValue, float stepSize)
...
+}
+
 void QuantumOutput::calculateFinished()
 {
   qDebug() << "The calculation finished!";
   if (!m_cube)
     return;
 
-  disconnect(&m_concurrent->watcher(), 0, 0, 0);
+  // Barry: I have no clue what disconnect does, but if statement
```
cryos: Don't need name, disconnect removes all signal-slot connections to the watcher. Your changes mean there is not always a valid watcher as no QtConcurrent job is initiated. I can take a pass and just clean up some of the comments/formatting as this is pretty close. You can't disconnect from a NULL, the debug statement is less valid now too, and we should probably just refactor the method for calculation and display as the two can now be entirely distinct.

barrymoo:
I don't mind cleaning everything up, unless you want it to be done ASAP. On Saturday, I should have time to:
* Clean up comments/code
* Add something to the QDialog for managing calculated and viewed cubes
* Separate calculation and viewing of meshes (MOs, cubes, densities, etc.)

I like the idea of every dialog being dockable (from a users perspective), but I think it would be best if you fixed that. Unless there is an example of this in the code already? I am more-or-less working from example right now.

cryos:
No rush, take care of the issues you suggest, happy to review. There are some other things I am trying to get cleaned up for a release. May not get to dockable, and would rather circle back after a release.

barrymoo:
I attempted to build a more flexible interface (building `surfaceCombo` on the fly etc.), but my attempt was thwarted. I get a strange error:

```
RCC Parse Error: '/home/barrymoo/src/openchemistry-fork/avogadrolibs/avogadro/qtplugins/quantumoutput/newsurfacedialog.ui' Line: 2 Column: 18 [unexpected tag: ui]
```

Just putting this up here, because Google is unhelpful and I need to focus on research. If I don't find anything I will edit `surfacedialog.ui` manually (although this seems like a stupid thing to do when qtcreator exists). Thanks!

barrymoo:
@cryos, I am getting somewhere with the new interface. Could you take a look at `newsurfacedialog.ui` from the most recent commit (just in qtdesigner is fine)?  I just want to know if you like the setup. Ultimately, I just want to separate the calculation and display of surfaces.

ghutchis:
I don't really like that dialog. I'm not sure what you're trying to accomplish - there should already be separation of calculation and display - this dialog should be about creating new surfaces, but the display is done by the engine.

barrymoo:
The goal here is to separate the calculation and display of surfaces. Currently, one needs to recalculate a surface every time you want to display it. This would add some flexibility in viewing.

barrymoo:
In the future I want to implement math with the calculated surfaces and display them. I suppose it makes more sense to put the new surfaces in the "Display" box. If you have any suggestions, I would gladly accept them.

ghutchis:
I would think tabs might be one way to go as far as math. (I'm sure many people would love to do Laplacians, etc. on calculated cubes.)  But as far as viewing.. I think we should emulate Avogadro v1, where the display types each had option windows. That's clearly a useful place for surface display UI (although a list might be more useful than the popup menu used in Avogadro v1).

barrymoo:
Alright, I will work on that.

PR #26, opened 2015-07-24
Newui rendersurfaces

barrymoo:
@cryos @ghutchis I put together the new UI which I think is more flexible. Additionally, I added a tab widget to extend the UI in the future. I will do some more testing and clean everything up this weekend. For now, there is a "NewQuantum" menu to test this UI out.

Note, I thought this would take me longer which is why I closed the other pull request.

### DIFF
--- a/avogadro/core/molecule.h
+++ b/avogadro/core/molecule.h
@@ -455,7 +455,8 @@ public:
   void clearCubes();
 
   /**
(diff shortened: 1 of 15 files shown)


barrymoo:
@cryos @ghutchis If you guys like the new surface UI, I can replace the "old" one.

PR #27, opened 2015-08-04
Newui rendersurfaces is done.

barrymoo:
@cryos @ghutchis NewQuantumOutput is done, sorry this took so long. All of this is working and ready to extend. If you guys like the new design I will merge it back to QuantumOutput.

### DIFF
--- a/avogadro/core/molecule.h
+++ b/avogadro/core/molecule.h
@@ -455,7 +455,8 @@ public:
   void clearCubes();
 
   /**
-   * Get the cubes vector set (if present) for the molecule.
+   * @brief Get the cubes vector set (if present) for the molecule.
+   * @return The cube vector for the molecule
    */
   std::vector<Cube *> cubes() { return m_cubes; }
   const std::vector<Cube *> cubes() const { return m_cubes; }
--- a/avogadro/core/molecule.h
+++ b/avogadro/core/molecule.h
@@ -455,6 +455,12 @@ public:
   void clearCubes();
 
   /**
+   * Get the cubes vector set (if present) for the molecule.
+   */
+  std::vector<Cube *> cubes() { return m_cubes; }
+  const std::vector<Cube *> cubes() const { return m_cubes; }
+
+  /**
    * Returns the chemical formula of the molecule
    * @todo This should eventually be an external algorithm, not a member of
    * Molecule.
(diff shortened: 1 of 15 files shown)


cryos:
Lumping everything together seems worse to me, so having the electronic surface lumped in with all the MOs makes it tougher to just calculate that when you want it. I also feel like you added a tabbed interface which takes up more space, offers redundant information, with the intent to add more tabs in the future?

What other tabs are you thinking of adding? I would like to overhaul this dialog, just not sure if this is really the right direction. I have started looking at the UI now that 0.8.0 has been tagged, and how it can be improved for end users. I would like to take the best from 1.x, and design some improved dialogs where cubes/other data should be easier to get at too.

barrymoo:
@cryos ultimately the goal was to introduce my CubeFile code into this interface. A new tab would be made to manipulate surfaces calculated in Avogadro. For example, one can compute approximate transition densities, take derivatives, etc. I wanted to get the base interface done before I advanced.

I looked at Avogadro 1.x interface and streamlined it (minor). "Surface Type" can be chosen based on the input file (opening gaussian.chk can compute MOs or Density, opening *.cube is a Cube).

I can gladly add more things, modify things, etc. I just don't know what's best for the community.

cryos:
That makes sense, I don't think lumping the electronic surface in with the molecular orbitals is the right thing to do. Same with cubes, I think there are so many MOs, and we necessarily want to put people in the middle with the HOMO/LUMO, then keeping them separate seems reasonable. We could just show the second level for MOs if MOs are available and selected. I will try to take a look at this again now your intent is a little clearer to me. Thanks for clarifying @barrymoo

barrymoo:
@cryos Absolutely, I was hoping to use the same interface but customize based on the input file. Let me know what you decide, I will write the code.

cryos:
Too much travel, I like your plans, and think we should continue down this path. We should replace the existing menu as it would be confusing to keep both.

barrymoo:
@cryos, hopefully for enjoyable reasons. I will replace the "old" interface as soon as I can.

ghutchis:
@barrymoo Are you still thinking of merging this branch? I think allowing import of surface/cube data is great, particularly since there are other formats (e.g., APBS generates OpenDX or some such.)

barrymoo:
@ghutchis Yes, I am having some trouble with my current forks. I will figure this out in the afternoon.

barrymoo:
Resurrected https://github.com/OpenChemistry/avogadrolibs/pull/62/commits/e9178090b83ff64abf961b85dd6f736eec81f509

ghutchis:
Can we close this in favor of #62 ?

barrymoo:
Yes.

PR #30, opened 2015-09-08
Fix build with const iterators.

cordlandwehr:


cryos:
LGTM - thanks for the patch.

PR #31, opened 2015-09-09
fix Gaussian FCHK parsing of unrestricted calculations

bryce-anderson:
Gaussian FCHK files have a propensity to not emit a `UHF` or `RHF` tag which is the only way the parser would check for the proper calculation type. This PR fixes this by double checking the calculation type if it finds a "Beta Orbital Energies" entry in the FCHK file and fixing the entries appropriately.


cryos:
Do you have any examples that you could share? This seems reasonable to make our code more robust, but it would be great to know where these might come from. Add an example to the avogradrodata repository if it can be shared, or over email if it can't be perhaps?

bryce-anderson:
I've put in a PR for adding gaussian calculations for H2O using restricted and unrestricted methods.

cryos:
Thanks, that is great and will be helpful in verification. Your changes look good to me, let's merge them  in.

PR #32, opened 2015-09-10
Fix linking to HDF5 libraries.

cordlandwehr:
Use library paths as provided by CMake's find module to correctly find
shared library in subdirectories, e.g.:
/usr/lib/x86_64-linux-gnu/hdf5/serial/libhdf5.so


cryos:
Looks good to me, thanks for your contribution!

PR #34, opened 2015-09-10
Fix the BasisSet.homo(..) functions

bryce-anderson:
Before this patch, basis.homo(basis.homo()) == false.

Added a unit test.

### DIFF
--- a/avogadro/core/basisset.h
+++ b/avogadro/core/basisset.h
@@ -95,10 +95,7 @@ public:
    */
   bool homo(unsigned int n)
   {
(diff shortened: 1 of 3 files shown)


bryce-anderson:
This is broken, but I feel the `BasisSet` class is currently unsuitable for use with open shell systems. In these systems the LUMO can be either alpha or beta depending on their relative energies yet `BasisSet` doesn't know anything about energies.

bryce-anderson:
Sorry, can we abort this? I need to fix the lumo functions as well.

PR #35, opened 2015-09-10
Fix the BasisSet.homo(..) and lumo(..) functions

bryce-anderson:
Before this patch, basis.homo(basis.homo()) == false and same for lumo.

Added a unit test.

This is a step in the right direction, but I feel the `BasisSet` class is currently unsuitable for use with open shell systems. In these systems the LUMO can be either alpha or beta depending on their relative energies yet `BasisSet` doesn't know anything about energies.

### DIFF
--- a/avogadro/core/basisset.h
+++ b/avogadro/core/basisset.h
@@ -95,10 +95,7 @@ public:
    */
   bool homo(unsigned int n)
   {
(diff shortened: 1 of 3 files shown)


bryce-anderson:
Sorry for the previous PR.

cryos:
No problem, there is some work to do for open shell systems and we appreciate your input here.

Issue #33, opened 2015-09-10
Support for beta orbitals


bryce-anderson:
Currently there doesn't appear to be the ability to plot anything but alpha (I presume they are alpha) orbitals. Is there a plan to add this functionality in the future? 

Currently the rendering appears to be under some flux (issue #27) but after that is done, I don't mind attempting to add that functionality.

cryos:
Yes, you are right. I think it would be great to add in that functionality, and would appreciate your help. Let's try and get this code merged in, and extend out our capabilities in dealing with electronic structure. I would be happy to give you any pointers you might need as you start looking into this.

ghutchis:
Long time coming, but @cryos has added this in #318

cryos:
Yes, I am sorry it took so long! It really wasn't that much code...

PR #38, opened 2015-09-11
WIP/openshell

bryce-anderson:
This is a work in progress. Input appreciated. Once this is ready for merging, I'll do a re-base/squash against master and open a new pull request.

This PR improves support for open shell electronic configurations. A number of changes are made to the internal representation of `BasisSet` and its sub-types.
The summary is that alpha and beta electrons are given equal status in both UHF and RHF configurations.

### DIFF
--- a/avogadro/command/qube.cpp
+++ b/avogadro/command/qube.cpp
@@ -30,6 +30,9 @@
 
 using Avogadro::Io::FileFormatManager;
 using Avogadro::Core::Cube;
+using Avogadro::Core::ElectronType;
+using Avogadro::Core::Alpha;
+using Avogadro::Core::Beta;
 using Avogadro::Core::Molecule;
 using Avogadro::Core::GaussianSetTools;
 using std::cin;
@@ -58,6 +61,7 @@ int main(int argc, char *argv[])
   // Process the command line arguments, see what has been requested.
   string inFormat;
   int orbitalNumber=0;
+  ElectronType spin(Alpha);
   string inFile;
   bool density = false;
   for (int i = 1; i < argc; ++i) {
@@ -77,6 +81,9 @@ int main(int argc, char *argv[])
     else if (current == "-orb" && i + 1 < argc) {
       orbitalNumber = atoi(argv[++i]);
       //cout << "plot orbital " << orbitalNumber << endl;
+    }
+    else if (current == "-beta" && i < argc) {
+      spin = Beta;
     }
     else if (current == "-dens" && i < argc) {
       density=true;
@@ -164,7 +171,7 @@ int main(int argc, char *argv[])
       printf("\n");
     }
     double value = m_tools->calculateMolecularOrbital(
-        m_qube->position(i),orbitalNumber);
+        m_qube->position(i),orbitalNumber,spin);
     printf("%13.5E",value);
     //line wrapping
     linecount++;
@@ -182,6 +189,6 @@ int main(int argc, char *argv[])
 
 void printHelp()
 {
-  cout << "Usage: qube [-i <input-type>] <infilename> [-dens] [-orb <orbital number>] [-v / --version] \n"
+  cout << "Usage: qube [-i <input-type>] <infilename> [-dens] [-orb <orbital number>] [-beta] [-v / --version] \n"
        << endl;
 }
--- a/avogadro/core/basisset.h
+++ b/avogadro/core/basisset.h
@@ -207,7 +207,7 @@ protected:
 
   /**
    * Energies of the beta molecular orbitals. If the system is restricted the
-   * the beta energies will not be defined.
+   * the beta energies will be the exact same as the alpha energies.
    */
   std::vector<double> m_betaMOEnergies;
 
--- a/avogadro/core/basisset.h
+++ b/avogadro/core/basisset.h
@@ -207,7 +207,7 @@ protected:
 
   /**
    * Energies of the beta molecular orbitals. If the system is restricted the
-   * the beta energies will be the exact same as the alpha energies.
+   * the beta energies will be unset.
    */
   std::vector<double> m_betaMOEnergies;
 
--- a/avogadro/core/basisset.h
+++ b/avogadro/core/basisset.h
@@ -26,6 +26,25 @@
 namespace Avogadro {
 namespace Core {
 
+/**
+ * Enumeration of the SCF type.
+ */
+enum ScfType {
+  Rhf,    // All orbitals doubly occupied.
+  Uhf,    // Alpha and Beta have separate orbitals and possibly different number of electrons.
+  Rohf,   // Some orbitals doubly occupied and some not.
+  Unknown
+};
+
+/**
+ * @brief The ElectronType enum describes the type of electrons being set or
+ * retrieved. Alpha and Beta both must be set regardless of shell type.
+ */
+enum ElectronType {
+  Alpha,
+  Beta
+};
+
 class Molecule;
 
 /**
@@ -44,7 +63,7 @@ public:
   /**
    * Constructor.
    */
-  BasisSet() {}
+  BasisSet(): m_scfType(Unknown) {}
 
   /**
    * Destructor.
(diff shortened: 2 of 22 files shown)


bryce-anderson:
This is about as much as I dare churn until the new rendering PR (#27) is merged. I have no experience with input types other than Guassian and there are a lack of examples in Avogadrodata so please read through changes there carefully.

After this is merged I'd like to go about updating the orbital plotting functionality to reflect the new information concerning alpha and beta orbitals (issue #33).

cryos:
This looks like you are duplicating data in the case of closed shell calculations. We talked about this at length before, and as so many calculations are closed shell we want to avoid duplicating data/calculations (which are often quite slow). I think we want to keep a design where only alpha is set/populated in closed shell systems, and beta is set/populated when necessary. In order to preserve the API I think it is reasonable to have Alpha as the default parameter value where before no value was passed, and those reading in open shell calculations will know to set alpha/beta.

bryce-anderson:
Indeed, I am duplicating some data for closed shell calculations but it should be a pretty trivial amount, on the order of kilobytes for medium sized calculations.

Are you referring to functions such as `homo` and `lumo` with regard to the default of `Alpha`? If so thats fair. I have never had good luck with default enum parameters in C++ because everything seems to implicitly coerce to a (compiler) acceptable type.

bryce-anderson:
I was in the process of adding default `Alpha` parameters and I don't think its a good idea for many of them: it will be extremely simple to write code that will be broken for open shell systems. Open shell systems are probably a minority but they are not uncommon and being explicit is a small price to pay to avoid generating incorrect results for open shell molecules down the road. 

Is there a reason to preserve the compile time API other than history that I'm overlooking?

ghutchis:
I'm fine with making things explicit. Marcus, this certainly breaks API (and ABI) but is anything linking to binaries yet? My feeling is that until Avogadro 2.0 is finally released, the API should be flexible.

bryce-anderson:
Depending on how much you're willing to change the code base, a more refined type model might make things more clear here. Currently, `BasisSet` would be better named `WaveFunction`. Refining the goal of `BasisSet` and making a real `WaveFunction` class with subtypes `Rhf` and `Uhf` would remove a lot of the implicit structure in the current model.
I'm not sure it will work out, but its a fair bit of code so I would like an opinion on it before diving in.

cryos:
We are using this in some projects that have deadlines, but it is open to being modified. I am fine with being explicit when it is necessary, just not convinced by some of the API changes proposed. I think we should work well with open shell systems, but in big calculations the proposed duplication is not insignificant, nor is the time that would be spent on duplicated calculations if things were blindly assumed to be open systems.

Also, things like electronCount already have a pattern I think we should expect - the default with no arguments gives you everything, or you can filter on a qualifier such as hydrogen to get the hygrogen count. I think this should be followed for electronCount in the case of alpha/beta. You have to change the way you deal with the system quitre a bit when it is open shell, saying you can write code that will be broken for open shell systems while being true I don't think is a good reason to not provide defaults for closed shell/nothing provided systems.

I am open to new approaches, but I want to be careful about duplicating data and changing defaults needlessly. You could easily keep the member functions with no arguments, and assert that they only be called for closed shell systems, add new methods that require an argument and not assert there. It would offer an easy to spot error in testing, without introducing too much of a speed penalty.

cryos:
I have been working with the NWChem developers too, and have seen a few weaknesses in our current model that I would like to spend some time addressing. It would be great to spend time getting open shell right too, Albert started on some of this too. As someone who seems very concerned by open shell it would be great to understand what you want from the API, and how you would like to work with these objects.

cryos:
Just locally squashing this down, and being explicit about changes that I would like to revisit (some may just be awaiting cleanup). General style - when defining functions/classes the opening brace gets its own line, other braces go on the end of the line for the conditional etc. I think moving open shell up if we can add support in slater types would be great. Adding tests is extremely welcome, and something I want to improve for the QM stuff.

In basisset.h I don't think you need to include iostream, and would rather avoid vector and the explicit alpha/beta energies there, but if we really want to store energies could live with it. You said you had trouble with default enum arguments, we can test them but I assure you they work well. I think we should default to Paired as we have. The iostream isn't worth it for the error message, and we should prefer to return 0 or indicate an error state in a way that an application can make sense of. We could add an assert.

gaussianset.h removal of defaults for Paired as above, I don't see how this helps. I agree on needing to write correct code, but I think there are better ways and think the approach being used worked reasonably. It would be good to see examples where these defaults really cause issues, and whether used function overloading and a few asserts might help more there.

avogadro/io/CMakeLists.txt - not sure why you changed this, it was a contributed patch recently, and seems to be needed due to changes in the way that HDF5 links. This should be moved to a separate patch if it is an issue.

bryce-anderson:
The CMakeLists.txt change was accidentally added and I intend to remove it. As an aside, that patch broke the build for me and, I believe, your CI system as well. It looks like it should work and maybe my Ubuntu env is the broken part. Unfortunately C++ doesn't foster an environment of reproducible builds.

The energies are necessary for determining the LUMO of open shell calculations. The current state of this branch doesn't reflect that because it requires one to be explicit about which spin you are interested in. However, to determine the 'global' homo they are necessary. 

I'm interested removing non 'basis set' properties from `BasisSet` and subtypes and making a new class, `WaveFunction` with `RHFWaveFunction` and `UHFWaveFunction` subtypes that contain the basis, MO coefficients (wavefunction eigenvectors), and energies (eigenvalues). Function overriding would make `homo` and `lumo` functions just as they are in master. Considerations that are RHF/UHF specific would live in the appropriate subtype. The work in this PR isn't representative of how that would look.

As far as the removal of the defaults, its not strictly necessary. In my mind, its a way for forcing people to at least think about the possibility of open shell system. For example, many of the tools in the `quantumio` package have given almost no consideration or just made a token gesture (such as [here](https://github.com/OpenChemistry/avogadrolibs/blob/master/avogadro/quantumio/gamessus.h#L100) which are never used or initialized...) at it but don't follow through because the type system doesn't make them think about it. Another example is that the MO visualization of beta orbitals is just missing because its not necessary to even think about. Unfortunately this is a natural result of defaults + humans. I think most of this would be alleviated by making the object model less flat as I propose above. However, I'm not going to sink the time into it if the project isn't interested in it.

cryos:
I think you misdiagnose the problem of humans + defaults, it is more limited time + priorities (and frankly open shell calculations have always been important, but not pressing). We are well aware open shell isn't working yet, Albert put some time into making a start on it. Forcing people to think about it won't make a consistent API, adding the feature and testing it will. I am willing to put some time into this, and think it is important. I would love to work with you on it.

If there are enough differences to justify it having a deeper hierarchy sounds reasonable to me. The fact of the matter is that the user interface needs to change when open shell calculations are present, and that will require a number of new interface elements and code paths. I don't think the simpler dialog for closed shell calculations should be crowded by this when it is not necessary. I have a tiny number of open shell examples, and so it would be great to flesh this out with someone who obviously has more.

I want to keep the API as intuitive as possible, and we have invested thought into some of the current design even if we didn't finish it. Open shell has been a missing feature since we started reading in QM data, and I would like to close that hole up. Most certainly interested in it, and I hope that is clear. Willing to look at API changes, but I would like to discuss them, and make reasonable changes to improve the API for all users. We have some current projects as I mentioned, and I am working on adding imrpoved support for NWChem output too.

HDF5 has two build systems, and I wonder if that is the main cause of that breakage... It can be built with autotools I think (many distros/system copies) or using CMake. I think CMake uses exported targets, and the autotools version does not. I will try to improve this so it works in both cases.

ghutchis:
The problem hasn't been "oh no one wants to think about open-shell calculations". The problem has, as Marcus mentioned, a matter of limited time and a matter of ranked priorities. You seem to have open shell calculations as high on your list, which is great - let's see if we can get them in soon. My priorities have been on the draw/edit side, etc.

Can we discuss this on the mailing list rather than the comments for a pull request? I think there are several people who may wish to weigh in on this. Bryce, you can outline (perhaps in pseudocode) some of the API you'd like to see, and it can be more broadly discussed.

How does that sound?

bryce-anderson:
@ghutchis, @cryos, sounds good to me.

bryce-anderson:
I was hoping to make a positive contribution to the project that doesn't appear to be as straight forward as I'd hoped.
Good luck, and sorry for wasting your time.

PR #54, opened 2016-03-09
Space Group Support

ghutchis:
Initial space group support from Albert DeFusco


cryos:
We shouldn't add submodules to avogadrolibs, ideally as said at the time spglib should be in the openchemistry repository, and found from here. It looks like this was later converted to not use the submodule, but the submodule was retained so perhaps we can simply delete the submodule that was added. I can spend some time testing that out if that is the case.

ghutchis:
I think this can be rebased. Since some of the GSoC proposals involve crystallography, I thought this WIP from @AlbertDeFusco would be useful background.

Agreed that adding spglib to the openchemistry repository would be ideal.

ghutchis:
I'm closing this, since @psavery has newer versions and can refer to the pull if needed (e.g., for the space group browser UI).

PR #56, opened 2016-03-14
Fixed a small misspelling

psavery:
This is just to demonstrate that I can submit pull requests...


cryos:
Looks good, it looks like you need to set the author name up correctly though as it is listed as psavery rather than your name. You can [amend that](http://stackoverflow.com/questions/3042437/change-commit-author-at-one-specific-commit), and use git config to set it more permanently.

psavery:
Did that fix it?

cryos:
It did - thank you. Merging.

PR #59, opened 2016-04-03
Provide copy constructor to create Molecule from base class.

cordlandwehr:
Make the interface symmetric to assignment operator.


ghutchis:
I know @cryos  had some specific thoughts on copy constructors.. Marcus?

cryos:
This was an oversight, there is already an assignment operator, so we should have added a copy constructor. Ideally one might use the other, but this looks good to me and thanks for spotting this.

PR #60, opened 2016-04-19
Workaround for systems with both python2 and 3

ktns:
Hi. 

Currently, ```cmake -DUSE_BOOST_PYTHON -DUSE_QT``` fails on systems with both python 2 and 3 due to discrepancy between minimum python versions for boostpython module and for qtgui module.

This patch at least workaround this issue, by finding later version of python first.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -31,6 +31,11 @@ endif()
 
 include_directories(BEFORE ${CMAKE_CURRENT_SOURCE_DIR}
   ${CMAKE_CURRENT_BINARY_DIR})
+
+if(USE_BOOST_PYTHON)
+  add_subdirectory(boostpython)
+endif()
+
 add_subdirectory(utilities)
 add_subdirectory(thirdparty)
 add_subdirectory(avogadro)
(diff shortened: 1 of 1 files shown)


cryos:
I don't see how this works around it, the add_subdirectory call is just made earlier. You should be able to set the Python version in cmake-gui, I have compiled this on systems with Python 2/3 and it works if setting the variables.

cryos:
The tougher problem, and I didn't find a good answer, is ensuring the Boost.Python version found matches the Python version found. Each Linux distro seems to have their own take on mangling the library name for Boost.Python, and so you have to manually match this up.

ktns:
> I don't see how this works around it, the add_subdirectory call is just made earlier.

Sorry, I've noticed it is a workaround only if /usr/bin/python is python2 and there is /usr/bin/python3.
In such system, the earlier call for python>3 finds /usr/bin/python3, but the earlier call for python>2 finds /usr/bin/python and fails at the later call for python>3.

> The tougher problem, and I didn't find a good answer, is ensuring the Boost.Python version found matches the Python version found.

Yes, it is the real problem and I don't know any solution either.

I'm closing the PR.

PR #61, opened 2016-04-26
Add a "sticky" drag for the draw tool to help prepare "normal" bonds.

ghutchis:
If a drag length is near (~0.1) a typical bond length, then
stick with the typical length. (This will become an option soon.)

Easier to experience than to describe in text.

### DIFF
--- a/avogadro/qtplugins/editor/editor.cpp
+++ b/avogadro/qtplugins/editor/editor.cpp
@@ -335,8 +335,7 @@ void Editor::bondRightClick(QMouseEvent *e)
   {
     Vector3 bondVector = atom1.position3d() - atom2.position3d();
     double bondDistance = bondVector.norm();
-    double radiiSum;
-    radiiSum = Elements::radiusCovalent(atom1.atomicNumber()) + Elements::radiusCovalent(atom2.atomicNumber());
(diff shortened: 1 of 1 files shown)


cryos:
This feels jerky on my machine, I feel like this is something that would be nice to support turning it on or off based on preferences. It may feel worse that it really should as the Ubuntu 16.04 update on my laptop seems to have triggered slow down in draw renders even for small molecules.

ghutchis:
Yes, my intent is for it to be an option (likely default to on). Once I can access hydrogen addition from a RWAtom, I'll also add angle tweaking.

I've also been working a bit on tweaking the "feel." The concept, I think is good - giving people slightly more accurate geometries without much work, but allowing distorted bonds as needed.

cryos:
You want to merge this, and continue working on it? I think it would be great to add an option to turn it off, but that doesn't need to be done before we merge this change. The changes look good to me.

ghutchis:
I'm not sure if I want to merge quite yet. It's fine to merge now, but it needs refinement and the option to turn off before I'm satisfied.

cryos:
OK, sounds good to me.

ghutchis:
Closing for now. I'll revisit the idea this summer.

PR #62, opened 2016-04-26
Resurrect newui-rendersurfaces

barrymoo:
This doesn't do beta orbitals yet, but I can work on that next.

### DIFF
--- a/avogadro/core/molecule.h
+++ b/avogadro/core/molecule.h
@@ -454,6 +454,13 @@ public:
 
   void clearCubes();
 
+	/**
+   * @brief Get the cubes vector set (if present) for the molecule.
+   * @return The cube vector for the molecule
+   */
+  std::vector<Cube *> cubes() { return m_cubes; }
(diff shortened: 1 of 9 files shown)


barrymoo:
One issue which existed before was: When the molecule is updated this interface breaks. I am working on trying to connect these processes, but I will be slow.

cryos:
This looks good, I will be traveling later and would like to fix up a few minor issues I see in a follow up commit (removed copyright headers, some formats removed, and double check the functionality still works. Thanks for bringing this up to date @barrymoo! I think we can help with some of the other issues you mention, but it will be good to integrate this and iterate on further features/updates.

barrymoo:
@cryos this is where https://github.com/OpenChemistry/avogadrolibs/issues/83 comes from. I can make a really nice interface with that functionality.

cryos:
@barrymoo I know, I got that.

cryos:
@barrymoo I was referring to copyright headers that had been removed above, and me adding them back in.

PR #65, opened 2016-05-11
Fixed a memory leak in an assignment operator.

psavery:
No need to check if the pointers are NULL since delete may be safely called on NULL pointers.

### DIFF
--- a/avogadro/core/molecule.cpp
+++ b/avogadro/core/molecule.cpp
@@ -59,7 +59,9 @@ Molecule& Molecule::operator=(const Molecule& other)
   if (this != &other) {
     m_graphDirty = true;
     m_customElementMap = other.m_customElementMap;
+    delete m_basisSet;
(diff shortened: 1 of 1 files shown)


cryos:
Good catch, although they must then be copied from other, or set to null in order to ensure we don't leave a dangling pointer (unless I am missing something).

psavery:
Thanks Dr. Hanwell! In the line following "delete m_basisSet", m_basisSet gets set to NULL, and in the line following "delete m_unitCell", m_unitCell gets set to NULL if other.m_unitCell is NULL, or a heap copy of other.m_unitCell is made if other.m_unitCell is not NULL.

cryos:
So we are missing a copy of the basis set, I missed the existing calls. This looks like a marked improvement, and we need to copy the basis set correctly in a follow up. Thanks, obviously not had enough coffee this morning.

PR #66, opened 2016-05-12
Added a clone function to BasisSet.

psavery:
Added a virtual copy constructor to BasisSet so that it may be copied.

### DIFF
--- a/avogadro/core/basisset.h
+++ b/avogadro/core/basisset.h
@@ -49,6 +49,11 @@ public:
   virtual ~BasisSet() {}
 
   /**
+   * Clone.
+   */
+  virtual BasisSet* clone() const = 0;
+
+  /**
    * @brief The ElectronType enum describes the type of electrons being set or
    * retrieved. If Paired, then Alpha and Beta cannot be set, if Alpha or Beta
    * then both must be set.
--- a/avogadro/core/gaussianset.h
+++ b/avogadro/core/gaussianset.h
@@ -66,6 +66,11 @@ public:
   ~GaussianSet() AVO_OVERRIDE;
 
   /**
+   * Clone.
+   */
+  GaussianSet* clone() const AVO_OVERRIDE { return new GaussianSet(*this); }
+
+  /**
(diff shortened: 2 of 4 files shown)


psavery:
Let me know what you think about this. I couldn't do a direct copy of BasisSet because it is an abstract class.

avogadro/core/basisset.h
```diff
@@ -49,6 +49,11 @@ class AVOGADROCORE_EXPORT BasisSet
   virtual ~BasisSet() {}
 
   /**
+   * Virtual copy constructor
```
cryos: It isn't a copy constructor, let's just call if a clone function.

avogadro/core/gaussianset.h
```diff
@@ -66,6 +66,11 @@ class AVOGADROCORE_EXPORT GaussianSet : public BasisSet
   ~GaussianSet() AVO_OVERRIDE;
 
   /**
+   * Virtual copy constructor.
+   */
+  GaussianSet* clone() const AVO_OVERRIDE {return new GaussianSet(*this);};
```
cryos: Don't need the final ;, please add a little white space such that

{ return new GaussianSet(*this); }

avogadro/core/slaterset.h
```diff
@@ -62,6 +62,11 @@ class AVOGADROCORE_EXPORT SlaterSet : public BasisSet
   ~SlaterSet() AVO_OVERRIDE;
 
   /**
+   * Virtual copy constructor.
+   */
+  SlaterSet* clone() const AVO_OVERRIDE {return new SlaterSet(*this);};
```
cryos: Space additions, and remove extra semicolon as above

psavery:
Will do. About to travel. Will update it later today.

cryos:
No rush, all fairly minor things. I will be leaving in about a half hour and less available until Saturday.

psavery:
Does that look fine now?

cryos:
Looks good to me.

PR #68, opened 2016-05-20
Added POSCAR format and clipboard import crystal.

psavery:
Added a POSCAR format reader and writer to the list of formats in Avogadro2.
Also added an "import crystal from clipboard" feature similar to that of
Avogadro1. It attempts to read the clipboard text using POSCAR format by
default. If a different format is to be used, the user can input that extension
and the crystal will be read for that format using obabel instead.

### DIFF
--- a/avogadro/io/poscarformat.cpp
+++ b/avogadro/io/poscarformat.cpp
@@ -1,0 +1,342 @@
+/******************************************************************************
+
+  This source file is part of the Avogadro project.
+
+  Copyright 2016 Kitware, Inc.
+
+  This source code is released under the New BSD License, (the "License").
+
+  Unless required by applicable law or agreed to in writing, software
+  distributed under the License is distributed on an "AS IS" BASIS,
+  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+  See the License for the specific language governing permissions and
+  limitations under the License.
+
+******************************************************************************/
+
+#include "poscarformat.h"
+
+#include <avogadro/core/elements.h> // for atomicNumberFromSymbol()
+#include <avogadro/core/matrix.h> // for matrix3
+#include <avogadro/core/molecule.h>
+#include <avogadro/core/vector.h> // for Vector3
+#include <avogadro/core/unitcell.h>
+#include <avogadro/core/utilities.h> // for split(), trimmed(), lexicalCast()
+
+#include <algorithm> // for std::count()
+#include <iomanip>
+#include <iostream>
+
+namespace Avogadro {
+namespace Io {
+
+using std::getline;
+using std::string;
+using std::vector;
+
+using Core::Array;
+using Core::Atom;
+using Core::Elements;
+using Core::Molecule;
+using Core::UnitCell;
+using Core::lexicalCast;
+using Core::split;
+using Core::trimmed;
+
+PoscarFormat::PoscarFormat()
+{
+}
+
+PoscarFormat::~PoscarFormat()
+{
+}
+
+bool PoscarFormat::read(std::istream &inStream, Core::Molecule &mol)
+{
+  size_t numLines = std::count(std::istreambuf_iterator<char>(inStream),
+                               std::istreambuf_iterator<char>(), '\n');
+
+  // There must be at least 7 "\n"'s to have a minimum crystal (including 1 atom)
+  if (numLines < 7) {
+    appendError("Error: POSCAR file is 7 or fewer lines long");
+    return false;
+  }
+
+  // We have to go back to the beginning if we are going to read again
+  inStream.clear();
+  inStream.seekg(0, std::ios::beg);
+
+  // We'll use these throughout
+  bool ok;
+  string line;
+  vector<string> stringSplit;
+
+  // First line is comment line
+  getline(inStream, line);
+  line = trimmed(line);
+  string title = " ";
+  if (!line.empty())
+    title = line;
+
+  // Next line is scaling factor
+  getline(inStream, line);
+  double scalingFactor = lexicalCast<double>(line, ok);
+
+  if (!ok) {
+    appendError("Error: Could not convert scaling factor to double in POSCAR");
+    return false;
+  }
+
+  Matrix3 cellMat;
+
+  // Next comes the matrix
+  for (size_t i = 0; i < 3; ++i) {
+    getline(inStream, line);
+    stringSplit = split(line, ' ');
+    // If this is not three, then there is some kind of error in the line
+    if (stringSplit.size() != 3) {
+      appendError("Error reading lattice vectors in POSCAR");
+      return false;
+    }
+    // UnitCell expects a matrix of this form
+    cellMat(0,i) = lexicalCast<double>(stringSplit.at(0)) * scalingFactor;
+    cellMat(1,i) = lexicalCast<double>(stringSplit.at(1)) * scalingFactor;
+    cellMat(2,i) = lexicalCast<double>(stringSplit.at(2)) * scalingFactor;
+  }
+
+  // Sometimes, atomic symbols go here.
+  getline(inStream, line);
+  stringSplit = split(line, ' ');
+
+  if (stringSplit.empty()) {
+    appendError("Error reading numbers of atom types in POSCAR");
+    return false;
+  }
+
+  // Try a lexical cast here. If it fails, assume we have an atomic symbols list
+  lexicalCast<uint>(trimmed(stringSplit.at(0)), ok);
+  vector<string> symbolsList;
+  vector<unsigned char> atomicNumbers;
+
+  if (!ok) {
+    // Assume atomic symbols are here and store them
+    symbolsList = split(line, ' ');
+    // Store atomic nums
+    for (size_t i = 0; i < symbolsList.size(); ++i)
+      atomicNumbers.push_back(Elements::atomicNumberFromSymbol(symbolsList[i]));
+    // This next one should be atom types
+    getline(inStream, line);
+  }
+  // If the atomic symbols aren't here, try to find them in the title
+  // In Vasp 4.x, symbols are in the title like so: " O4H2 <restOfTitle>"
+  else {
+    stringSplit = split(title, ' ');
+    if (stringSplit.size() != 0) {
+      string trimmedFormula = trimmed(stringSplit.at(0));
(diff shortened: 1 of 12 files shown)


psavery:
Let me know if I should change anything here. It's similar to Avogadro 1's import crystal from clipboard with one main difference: you can put in a different file extension and it will try to read the crystal using open babel.

Technically, POSCAR files are just ASCII files and don't have file extensions. However, in open babel we used "POSCAR" for the file extension, so I thought I could use that again here.

Also, I'm not sure what the mime type would be. I just put in "N/A".

avogadro/io/poscarformat.cpp
```diff
@@ -0,0 +1,337 @@
...
+  // We'll use these throughout
+  bool ok;
+  string line;
+  vector<string> stringSplit;
+
+  // First line is comment line
+  getline(inStream, line);
+  line = trimmed(line);
+  string title = " ";
+  if (!line.empty()) title = line;
```
cryos: Please use a new line after the if statement,

if (!line.empty())
  title = line;

avogadro/io/poscarformat.cpp
```diff
@@ -0,0 +1,337 @@
...
+    }
+    // UnitCell expects a matrix of this form
+    cellMat(0,i) = lexicalCast<double>(stringSplit.at(0)) * scalingFactor;
+    cellMat(1,i) = lexicalCast<double>(stringSplit.at(1)) * scalingFactor;
+    cellMat(2,i) = lexicalCast<double>(stringSplit.at(2)) * scalingFactor;
+  }
+
+  // Sometimes, atomic symbols go here.
+  getline(inStream, line);
+  stringSplit = split(line, /*delimiter=*/ ' ');
```
cryos: Most code editors will show you the arguments nowadays, this doesn't hurt but I am not sure it is needed either

avogadro/io/poscarformat.cpp
```diff
@@ -0,0 +1,337 @@
...
+  lexicalCast<uint>(trimmed(stringSplit.at(0)), ok);
+  vector<string> symbolsList;
+  vector<unsigned char> atomicNumbers;
+
+  if (!ok) {
+    // Assume atomic symbols are here and store them
+    symbolsList = split(line, /*delimiter=*/ ' ');
+    // Store atomic nums
+    for (size_t i = 0; i < symbolsList.size(); ++i)
+      atomicNumbers.push_back(Elements::atomicNumberFromSymbol(symbolsList.at(i)));
```
cryos: As we are doing the work to assure we don't go outside of bounds I would prefer [] to .at() to avoid the bounds checking cost.

avogadro/io/poscarformat.cpp
```diff
@@ -0,0 +1,337 @@
...
+  line = trimmed(line);
+  if (line.empty()) {
+    appendError("Error determining Direct or Cartesian in POSCAR");
+    return false;
+  }
+
+  bool cart;
+  // Check if we're using cartesian or fractional coordinates:
+  if (line.at(0) == 'K' || line.at(0) == 'k' ||
+      line.at(0) == 'C' || line.at(0) == 'c' )
```
cryos: When the if statement takes more than one line we prefer to use { } to make it clear where the statement is.

avogadro/io/poscarformat.cpp
```diff
@@ -0,0 +1,337 @@
...
+    cart = true;
+  // Assume direct if one of these was not found
+  else
```
cryos: Then it would spread here, as once a block needs {} they should be used for all branches for clarity

avogadro/qtplugins/crystal/crystal.cpp
```diff
@@ -152,11 +160,24 @@ void Crystal::updateActions()
...
+    m_importCrystalClipboardAction->setEnabled(true);
     m_toggleUnitCellAction->setEnabled(true);
     m_toggleUnitCellAction->setText(tr("Add &Unit Cell"));
   }
 }
 
+void Crystal::importCrystalClipboard()
+{
+  ImportCrystalDialog d;
+  Avogadro::Core::Molecule m;
```
cryos: You don't need the Avogadro:: qualifier as you are already in the Avogadro namespace

avogadro/qtplugins/crystal/importcrystaldialog.cpp
```diff
@@ -0,0 +1,94 @@
...
+{
+  delete m_ui;
+}
+
+bool ImportCrystalDialog::importCrystalClipboard(Avogadro::Core::Molecule &mol)
+{
+  QString text = QApplication::clipboard()->text();
+  m_ui->edit_text->setText(text);
+  // If the user rejected, just return false
+  if (this->exec() == QDialog::Rejected) return false;
```
cryos: Please use a new line after the if statement

avogadro/qtplugins/crystal/importcrystaldialog.cpp
```diff
@@ -0,0 +1,94 @@
...
+  // Use POSCAR format by default. If the extension was set, use that instead
+  std::string ext = m_ui->edit_extension->text().toUpper().toStdString();
+
+  // Update the text
+  text = m_ui->edit_text->toPlainText();
+  std::stringstream s(text.toStdString());
+
+  if (ext.empty() || ext == "POSCAR") {
+   Avogadro::Io::PoscarFormat format;
+   if (format.read(/*istream=*/s, /*Molecule=*/mol)) return true;
```
cryos: Please use a new line after the if statement

avogadro/qtplugins/crystal/importcrystaldialog.cpp
```diff
@@ -0,0 +1,94 @@
...
+   // Print out the error messages if we failed
+   else qDebug() << QString::fromStdString(format.error());
```
cryos: Please use a new line after the if statement

avogadro/qtplugins/crystal/importcrystaldialog.cpp
```diff
@@ -0,0 +1,94 @@
...
+  }
+  // Try using the extension
+  else {
+    // We don't need to set any of the info, so just put in empty strings for
+    // most arguments (except the extension)
+    OBFileFormat ob("", "", "", "", vector<string>{ext}, vector<string>{""});
+    if (ob.read(/*istream=*/s, /*Molecule=*/mol)) return true;
```
cryos: Please use a new line after the if statement

avogadro/qtplugins/crystal/importcrystaldialog.cpp
```diff
@@ -0,0 +1,94 @@
...
+    // Print out the error messages from the read if we failed
+    else qDebug() << QString::fromStdString(ob.error());
```
cryos: Please use a new line after the else statement

cryos:
Please develop in named branches, it makes it clear what you are doing. See inline comments on some style issues, looking good overall, just going to try building it. I don't see any examples on how I should use it, is there some avogadrodata example we could add to that repository, and a new test?

avogadro/qtplugins/crystal/importcrystaldialog.ui
```diff
@@ -0,0 +1,80 @@
...
+   </rect>
+  </property>
+  <property name="windowTitle">
+   <string>Import Crystal</string>
+  </property>
+  <layout class="QGridLayout" name="gridLayout">
+   <item row="0" column="0">
+    <widget class="QLabel" name="label">
+     <property name="text">
+      <string>File extension for open babel conversion (default: POSCAR):</string>
```
cryos: Is this using Open Babel, or Avogadro? Should use Open Babel capitalization, but it looks like the work above added a POSCAR format. It would be good to get a pointer to a file so I can see it in action.

psavery:
I'm out and about now. I'll fix these soon-ish. Here's an example that I just made up. You could test it to see that you get the same results for Avogadro 1 and 2 (you should). I can give you a real example like rutile soon. For the writer, maybe just perform an action on it like "rotate to standard orientation" and then write it to see if you get the same results. I did.

And it does use the Avogadro POSCAR format. So maybe I should word that differently.

Here's a sample:

O1Ti1
1.00000000
1.30000000 2.10000000 3.20000000
2.30000000 1.10000000 2.20000000
3.30000000 3.10000000 1.10000000
O Ti
1 1
Direct
0.00000000 0.75000000 0.75000000
0.50000000 0.50000000 0.50000000

psavery:
Okay. I fixed the changes you mentioned and sent in a pull request with a rutile POSCAR file to avogadrodata.

I also now said "default - Avogadro::POSCAR" instead of just "default: POSCAR" to clarify that it is using the Avogadro POSCAR and not the OpenBabel one.

The way that it works is that it will by default (or if the user inputs "POSCAR" for the file extension) use the Avogadro POSCAR reader . If the user inputs any other file extension, it tries to use OpenBabel to convert it.

The import crystal from clipboard feature is especially helpful for POSCAR files, though, because they are typically just named POSCAR (that's what they are supposed to be for VASP). People don't typically put a ".POSCAR" at the end, so a normal open would not work (they'd have to rename it). Plus, they are small enough that it's just easy to copy it to the clipboard and then just use the "import crystal from clipboard" feature.

I tested this with a couple of other file types, too (like GULP or GOT). When I used the GOT extension, it read it no problem.

psavery:
Oh yes, and I will add a test to tests/io soon

psavery:
Okay. I added a POSCAR test. It refers to the rutile POSCAR example I sent in a pull request for in avogadro/data.

ghutchis:
Do we really need a POSCAR reader, since there's already one in Open Babel? I agree that avogadrolibs needs support for some formats, but I'd like to minimize maintenance headaches..

psavery:
That'll be fine if I need to remove it. I think it'd be nice if we eventually moved away from dependence on OpenBabel just because of the licensing issues (and the fact that we have to use pipe calls), but I realize that might not happen because there's too much stuff to transfer over, and it's probably not worth it...

cryos:
As it is done I am fine with merging it, it is self-contained so if maintenance becomes an issue we can easily remove it from the repository. I don't know the pros and cons of this particular format, but am inclined to live with some duplication. I am going to merge as it takes care of issues I was concerned with.

I am not sure of the best balance, but there are several advantages to supporting formats we want to work with more closely including speed/scalability.

PR #69, opened 2016-05-26
Fixed a bug for transforming atoms.

psavery:


psavery:
When transforming atoms (rotate to standard orientation, Niggli reduction,
change volume, etc.), I noticed an issue with the transformations. Here's
an example:

Performing a "rotate to standard orientation" on this crystal:
O1Ti1
 1.00000000
   1.30000000   2.10000000   3.20000000
   2.30000000   1.10000000   2.20000000
   3.30000000   3.10000000   1.10000000
   O   Ti
   1   1
Direct
  0.00000000  0.75000000  0.75000000
  0.50000000  0.50000000  0.50000000


Should transform it to this (transpose removed):
O1Ti1
 1.00000000
   4.04227659   0.00000000   0.00000000
   3.05273519   1.42155121   0.00000000
   3.54255818   1.83286254   2.40850502
   O   Ti
   1   1
Direct
  -0.00000000  0.75000000  0.75000000
  0.50000000  0.50000000  0.50000000


Not this (transpose not removed):
O1Ti1
 1.00000000
   4.04227659   0.00000000   0.00000000
   3.05273519   1.42155121   0.00000000
   3.54255818   1.83286254   2.40850502
   O   Ti
   1   1
Direct
  -0.38703487  1.40193860  0.57312530
  -0.56795951  2.11810079  0.13622336



Note that removing the transpose gives us the same results as Avogadro 1.
Avogadro 1 performs the rotation correctly. If the transpose stays around,
it actually changes the crystal so that it is no longer the same crystal.

The same is true for a Niggli reduction. Removing the transpose fixes the
issue.

PR #72, opened 2016-05-31
Fixed an important bug in setCellParameters()

psavery:


psavery:
I knew something was wrong with this function for a few days... figured it out today

cryos:
Commit message seems a little overly mysterious, but the diff makes it quite clear. Merging, but in future something a little more descriptive like fixed typo in matrix value assignment or similar would make it clearer what was fixed.

psavery:
Will do

PR #73, opened 2016-06-03
Added basic spglib support to avogadrolibs

psavery:
spacegroupdata.h contains space group information based upon a hall number.
spacegroups.h provides the interface to access the space group information
in spacegroupdata.h. avospglib.h is a basic wrapper to obtain a hall number
with spglib functions. A test wals also added in tests/core/spglibtest.cpp.


psavery:
I may need to change a few things here as well. Let me know if I do. This code here is contingent on the openchemistry pull request I sent in just recently. That downloads, compiles, and sets up spglib for linking.

As I mentioned in the commit, spacegroupdata.h contains space group information based upon a hall number. It is similar to elementdata.h. It was generated by parsing the data in Albert's file that he copied from spglib. All data is accessed by using the hall number. There are 530 hall numbers total, so they are stored as unsigned shorts (unsigned char is a little too small). spacegroups.h is used to access the data.

avospglib.h is a basic wrapper to obtain a hall number with spglib functions. I think that one of the best ways to set up this model is to just use the hall number as an identifier (there is a unique hall number for every space group and setting combination - sometimes each space group may be represented in more than one way. That's why there are more hall numbers than there are space groups). Once the hall number is obtained, any of the data can be found by calling the functions in spacegroups.h. I'm thinking about caching the hall number in Avogadro::Core::Molecule and resetting it to 0 if any non-const function is called. I think we'll have to cache it somewhere anyways when we add a "setSpaceGroup" function.

A test was also added in tests/core/spglibtest.cpp. It creates a rutile crystal, finds the hall number using spglib, then checks to make sure the hall number is what it should be and that all of the lookups in spacegroupdata.h are what they should be as well. It worked just fine when I tested it. I tested this on a few systems and compared it to Avogadro1 results and got the same thing in the cases I tried.

Let me know what you think!

ghutchis:
I think Albert was suggesting something like this (i.e., keeping track of the Hall settings) because there's the space group number and then there's the "setting". I can't remember how the numbering works - would a Hall number of 0 correspond to a system without a unit cell (and thus space group)?

I'd ask @cryos about caching in Core::Molecule. It makes sense, but I'm still unclear on why RWMolecule can't be (isn't) a subclass of Molecule.

cryos:
I think this looks good, and we can look at caching in Core::Molecule in a future commit. Like I said before, RWMolecule contains no data, and will benefit from the cached data in Molecule -> there should be no duplication but we can add undo/redo commands there.

PR #74, opened 2016-06-05
Fix bug preventing keypress from changing bonds or elements.

ghutchis:
Now, typing 1,2,3,4 = bond order, 0 = automatic.
Typing element selects element.

Previous code would silently discard atomicnum from keypress through the tool widget.
Now, the code properly updates the selector.


cryos:
Looks good to me.

PR #75, opened 2016-06-06
Moved subdirectories after options in CMakeLists

psavery:
This fixes a compiling error if you compile the most updated
avogadrolibs from scratch. Since the options were not set before
adding the subdirectories, options such as USE_QT were set to be
false automatically, and then several needed libraries such as
qtgui would not be compiled.


psavery:
Interestingly, this is the way I was getting around this error before:

make -j4
COMPILE ERROR
make -j4

The options would be set in the first make -j4, so when I make again, it would work the second time.

So this update fixes this unnecessary issue.

cryos:
Good catch, those initial cache issues are easy to overlook when you are testing CMake code like this. Thanks!

PR #76, opened 2016-06-06
Auto-adjust draw hydrogens

ghutchis:
This is work-in-progress to adjust valence when editing.

The code is simple, but it exposes a really weird bug when draw-editing.
Perhaps the valence should only be changed at mouseRelease?

### DIFF
--- a/avogadro/core/CMakeLists.txt
+++ b/avogadro/core/CMakeLists.txt
@@ -29,7 +29,6 @@ set(HEADERS
   gaussianset.h
   gaussiansettools.h
   graph.h
-  hydrogentools.h
   matrix.h
   mesh.h
   molecule.h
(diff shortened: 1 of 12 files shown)


ghutchis:
Hydrogen geometries can be strange (because H atoms are re-used) but that's better addressed in a comprehensive hydrogen tools patch (e.g., flagging H atoms added by auto-adjust and/or removing all H-atoms to re-create new geometry).

ghutchis:
This is now finished - matching Avogadro v1 behavior as closely as I can. H positions will change when editing (i.e., will be removed and re-added in accurate positions).

cryos:
Feels nice and natural, massive improvements over the previous version, in my testing it was stable but I was sometimes able to orphan some of the added hydrogen atoms. It looks great to me, didn't spot any issues with the code.

You want me to merge, or please feel free to do so yourself. Great addition.

ghutchis:
Given all the debugging into this feature in Avogadro v1, I suspect there will be some snags still to catch. Is there a way to grab the hydrogens by unique ID instead of index?

Also, I'm having trouble with the "interactive editing" feature of RWMolecule so that edits can be merged into one undo.

PR #77, opened 2016-06-12
Moved hydrogentools test to qtgui and updated it

psavery:
This fixes a compiling error when we compile the tests. HydrogenTools was
recently moved from core to qtgui, so we should move the test as well. In
addition, Molecule was changed to RWMolecule and Atom to RWAtom to reflect
the changes in the in the hydrogentools functions (because they now take
RWMolecule instead of Molecule). The hydrogen tools test passes when it is ran.

### DIFF
--- a/tests/core/CMakeLists.txt
+++ b/tests/core/CMakeLists.txt
@@ -12,7 +12,6 @@ set(tests
   Eigen
   Element
   Graph
-  HydrogenTools
   Mesh
   Molecule
   Mutex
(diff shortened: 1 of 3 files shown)


psavery:
Unfortunately, it looks like some of the tests in qtgui (not hydrogentools) have been causing a segmentation fault (probably due to updates in the code without updating the tests). This does not fix that issue. I removed all tests except for hydrogentools when I ran the test, and it passed all of them. So the hydrogentools test is fine.

PR #78, opened 2016-06-14
Update scene radius and center to fix clipping.

psavery:
The description of resetGeomtry() in GLRenderer says "Reset the
scene geometry, this should be done when the scene geometry has
changed in order to ensure correct clipping." Unfortunately,
since caching was added to the Scene class, resetGeometry() does
not appear to do anything. This is because the radius and center
from Scene are cached, and it does not get set to "dirty" when
the group node is edited. Because they are being cached, they
do not get updated with the changes in geometry, and the radius
and the center do not get changed in GLRenderer::resetGeometry().
This results in clipping when a large molecule or crystal is
constructed. This update fixes that by forcing a re-calculation
of the center and radius of the scene before setting them in
GLRenderer.

### DIFF
--- a/avogadro/rendering/glrenderer.cpp
+++ b/avogadro/rendering/glrenderer.cpp
@@ -119,6 +119,7 @@ void GLRenderer::resetCamera()
 
 void GLRenderer::resetGeometry()
 {
+  m_scene.setDirty(true);
   m_center = m_scene.center();
   m_radius = m_scene.radius();
 }


psavery:
When I was looking into the clipping that you mentioned for the
super cell builder, I noticed that it was more generic than just
the super cell builder. It was an issue for making any crystal
larger. Then I noticed that it wasn't just crystals either: it
was an issue with molecules as well. If you zoom out and draw
a really long molecule, for instance, and then you rotate it,
you will notice that it gets clipped.

After spending a lot of time looking into this, I found out that
this is the reason: the scene center and radius are not being
updated when the node changes. This commit here added caching to
the center and radius:

https://github.com/OpenChemistry/avogadrolibs/commit/fdd653b0c603629fc905f533d8687b1e1c35972a

And that is probably a good thing. But notice that it is edited
by obtaining a reference to its root node:

https://github.com/OpenChemistry/avogadrolibs/blob/master/avogadro/rendering/scene.h#L116

The scene is apparently not being set to dirty somewhere after
it has been changed. I guess we could go in the code and make
sure that the scene is set to dirty any time the group node
is edited. However, it looks like this function is supposed to
be called anyways when the scene geometry changes:

https://github.com/OpenChemistry/avogadrolibs/blob/master/avogadro/rendering/glrenderer.h#L66

So I think it'd make sense to just set the scene to dirty and
update it when that function is called, right?

Let me know if you think there is a better way. But this fixes
the clipping issues we saw with building a super cell. It also
fixes the clipping issues I saw with building a large crystal
using any method and drawing a large molecule.

cryos:
Looks about right, let's give it a try. Good catch, thanks for digging into this.

PR #80, opened 2016-06-14
Update element symbols and names through 118 based on IUPAC

ghutchis:
Probably needs translations, esp. for element names.


ghutchis:
Ready to review and/or merge. I do think the element names need translation calls (e.g., iron needs localization, UK spells "aluminium", etc.)

PR #82, opened 2016-06-14
Change vector initialization to work with Clang

psavery:
Don't use vector<T>{list} to initialize vectors because
it doesn't work for every compiler.


ghutchis:
I can confirm this works for Clang / Mac, thanks.

PR #88, opened 2016-06-23
Newui rendersurfaces

cryos:


cryos:
This just adds back in some things that were removed, I think it is time to merge this and move forward. Thanks for updating this @barrymoo

PR #90, opened 2016-06-26
Added undo features for all crystal operations

psavery:
Added undo and redo commands for all crystal operations in RWMolecule.
They are also now all called so that the undo stack gets updated with
the changes.

### DIFF
--- a/avogadro/core/crystaltools.cpp
+++ b/avogadro/core/crystaltools.cpp
@@ -672,7 +672,6 @@ bool CrystalTools::setCellMatrix(Molecule &molecule,
                                  const Matrix3 &newCellColMatrix,
                                  Options opt)
 {
-
   if (opt & TransformAtoms && molecule.unitCell()) {
     const Matrix3 xform(newCellColMatrix
                         * molecule.unitCell()->cellMatrix().inverse());
--- a/avogadro/qtgui/rwmolecule.cpp
+++ b/avogadro/qtgui/rwmolecule.cpp
@@ -21,11 +21,15 @@
 #include <algorithm>
 #include <cassert>
 
+#include <avogadro/core/avospglib.h>
+
 namespace Avogadro {
 namespace QtGui {
 
 using Core::Array;
 using Core::AtomHybridization;
+using Core::CrystalTools;
+using Core::UnitCell;
 
 // Base class for all undo commands used by this class.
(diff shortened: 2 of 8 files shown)


psavery:
I’ve tested all of them and they seem to work fine. For testing this kind of thing, I found it especially helpful to have the text for the undo and redo commands displayed. I added that via this pull request:

https://github.com/OpenChemistry/avogadroapp/pull/29

Details:

I emit all the changes inside the RWMolecule functions, so I removed all emitting of changes outside it for these calls.

I added a very generic ModifyMoleculeCommand with a very generic modifyMolecule() function.  modifyMolecule() takes the new molecule as a parameter, and it takes the changes to be emitted and the undo text to be displayed as its other parameters. ModifyMoleculeCommand saves a copy of the previous molecule and the new one. For smaller changes in the molecule, it may not be very good to use because it takes sizeof(Molecule) space for both the old and new molecule (and if you’re just changing something like a bond, for instance, you don’t need to store everything else inside the molecule on the undo stack – and I know we already have commands for the bond anyways…). However, for big changes, it can probably be pretty effective at storing the changes by just storing the changes to the molecule as a whole.

In my case, I thought it was appropriate to use this for importCrystalFromClipboard(), because this feature overwrites whatever previous molecule you had with a new one. So it DOES need to store a copy of both the old and the new molecule.

I also ended up using modifyMolecule() for crystal operations that make many changes to the cell. Primitive reduction, cell conventionalization, Niggli reductions, and building a supercell change all of the following: number of atoms, atom positions, and the unit cell. I guess that we could potentially keep track of each one of these changes (old and new copies for atom positions, atomic numbers, and the unit cell), but I’m not sure if it is worth the effort since almost everything that can change does change. So I just have them using modifyMolecule() and storing a copy of the old and new molecules on the undo stack.

I also used modifyMolecule() for all the other cell operations that resulted in more than one thing being changed. Rotating a cell to standard orientation, editing a unit cell, and scaling a cell volume change both the atom positions and the unit cell. Let me know if you think it is worth the time for me to create a separate command for these instead of ModifyMoleculeCommand so that we don’t store the whole molecule.

For wrapping atoms to the cell, I just used the SetPositions3dCommand since only the atom positions are changing.

For adding and removing the unit cell, I made separate commands that store the new unit cell (for adding a unit cell) or the old unit cell (for removing a unit cell).

PR #92, opened 2016-07-07
Added fill unit cell and periodic distance calc

psavery:
Used a parser to retrieve and sort all space group transforms from
Open Babel at data/space-groups.txt. Added them to spacegroupdata.h.
Added some readers to spacegroups.cpp for these transforms. Then,
the fill unit cell features were added. Tests were added as well
to demonstrate that it works. Finally, a periodic distance calculator
was also added to unitcell.h (it is used in fillUnitCell()).


psavery:
I was originally thinking about using --fillUC in Open Babel. However,
I also need a function to reduce a crystal to its asymmetric form - which
is currently not available in Open Babel as far as I can tell. So I went
ahead and just got the information I needed for the functions - the transforms.
As I mentioned in the commit, I retrieved them from data/space-groups.txt in
Open Babel. I'll be able to use them for the asymmetric reduction function
as well.

The space groups were sorted just a little bit differently in Open Babel's
data/space-groups.txt. I wrote a Python parser to read the transforms in
space-groups.txt and sort them in the same order that we have hall numbers
in Avogadro2. I was able to get exactly 530 transforms (as I would expect
since there are 530 hall numbers - so I do not think any were left out).
I also used some Python parsers to make sure that there are no
multi-digit numbers used (my readTransformCoordinate() function in
avogadro/core/spacegroups.cpp assumes this). I guess that transform reader
would be the sketchiest function. I put a lot of assertions in it, though,
to make sure I am reading the transforms correctly, and every test I've tried 
has worked fine. So I think it's good.

Most cif files contain the transforms they want you to use - but they should
be identical to the ones used by the program. Perhaps I can add a reader for
the transforms in the cif files in the future...

To make sure I don't add an atom on top of another atom that is already in a
spot, I added a distance checker. Because of periodic conditions, however,
distance calculations can be a little bit more confusing. I added a
calculator for this in avogadro/core/unitcell.h. It may need some more
testing in the future, though...

I tested the fill unit cell on several systems, and it worked just fine - as
it did in Avogadro1. To be thorough, I added three tests for it. One is for
MgSiO3 post-perovskite, one is for dolomite, and one is for wollastonite. I
have the hyperlinks to the pages in the comments in the tests. fillUnitCell()
worked just fine for every case.

cryos:
Looking good, I may need more time to go over it all, but let's merge it, check for any issues, and iterate. Thanks for all the hard work!

PR #95, opened 2016-07-15
Added reduce to asymmetric unit function

psavery:


psavery:
I accidentally  kept this separate pull request's commit here:

https://github.com/OpenChemistry/avogadrolibs/pull/94

I can fix that if it is a problem. The second pull request is the main one here. Its commit says:

"A reduction to an asymmmetric unit basically means that
only one atom from each Wyckoff position will be kept - the rest
will be removed. Essentially, this takes the transforms of the atoms
checks to see if there is an atom lying on any of the transform
positions. If there is an atom there, it removes that atom.
It is essentially the opposite of filling a unit cell.

It seems to work fine. I added tests for it as well."

I should also mention that I already set it up for undo commands. So this one won't need a separate pull request with undo features. 

I also did a little bit of testing for my UnitCell::distance() function to make sure it takes into account periodic distances. It appears to work fine in my tests. So I'm happy about that!

Let me know if there are any issues with this!

PR #96, opened 2016-07-19
Added adjustHydrogens to RWMolecule

psavery:
Added adjustHydrogens to RWMolecule as an undo command.
Also added adjustHydrogens for multiple atoms as a single
undo command. In addition, added a new addAtom() that combines
adding the atoms and setting its position into one command. This
reduces the number of undo commands that appear while drawing. More
still needs to be done to reduce the number of undo commads further,
but I believe this is a good start.

### DIFF
--- a/avogadro/qtgui/rwmolecule.cpp
+++ b/avogadro/qtgui/rwmolecule.cpp
@@ -22,6 +22,7 @@
 #include <cassert>
 
 #include <avogadro/core/avospglib.h>
+#include <avogadro/qtgui/hydrogentools.h>
 #include <avogadro/core/spacegroups.h>
 
 namespace Avogadro {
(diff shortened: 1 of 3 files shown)


psavery:
This uses undo stack's "beginMacro()" and "endMacro()" features to combine some of the commands. I think this is the direction we want to be heading to combine more of them.

Now, if you left click on empty space with a carbon atom and "adjust hydrogens" checked, there are only two undo commands: one for adjusting the hydrogens and one for adding the atom. We can probably combine these two later.

Now, if you draw an ethane molecule, there is one undo command to remove all adjusted hydrogen atoms. Once again, we can do more to combine other commands later.

cryos:
Looks good to me, but has conflicts with your previous commits I assume.

psavery:
Okay, the conflict is fixed.

cryos:
Thanks!

PR #97, opened 2016-07-20
Fixed some bugs in the atomic coordinate editor

psavery:
The atomic editor wasn't working at all for me originally because
movePosition(QTextCursor::NextCharacter) was not moving it to the
end of the line. movePosition(QTextCursor::Down) followed by
movePosition(QTextCursor::StartOfLine) seems to work fine! And I
have it checking to make sure it only moves if it is not at the
end of the document. Additional bug fixes: program no longer freezes
if the user opens up the coordinate editor and hits "enter" or
"return" on a blank document. In addition, the input validator skips
over empty lines - this is helpful if the user deletes some lines but
forgets to delete white space or returns. Finally, I also set it up
so that the undo stack is updated when the coordinates are edited,
and it is now "undo-able."

### DIFF
--- a/avogadro/qtplugins/coordinateeditor/coordinateeditordialog.cpp
+++ b/avogadro/qtplugins/coordinateeditor/coordinateeditordialog.cpp
@@ -19,6 +19,7 @@
 #include "coordinatetextedit.h"
 
 #include <avogadro/qtgui/molecule.h>
+#include <avogadro/qtgui/rwmolecule.h>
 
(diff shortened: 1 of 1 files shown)


psavery:
Try this out if you get the chance! I think it really improves the usage of the atomic coordinate editor - no more having to worry about several small bugs. For me personally, it's helpful to be able to use the atomic coordinate editor when testing crystal features.

Let me know if there are any issues.

cryos:
Looks good

PR #99, opened 2016-08-11
Draws in edit mode are now properly undo-able

psavery:
Added some functions called beginDraw() and endDraw() in
RWMolecule. These functions use beginMacro() and endMacro().
In addition, when endDraw() is called, changes are emitted.
When a left click or right click is pressed, beginDraw() is
called. When the click is released, endDraw() is called. This
allows whole draw actions to be lumped into a single undo and
redo command.

Unfortunately, if there is only one undo on the undo stack, and
an undo and a redo is called, there seems to be a glitch where
all atoms are put at (0,0,0). This was not introduced by this
commit, but it needs to be fixed sometime in the future...

### DIFF
--- a/avogadro/qtgui/rwmolecule.h
+++ b/avogadro/qtgui/rwmolecule.h
@@ -502,7 +502,28 @@ public:
    * @param cartTol Cartesian tolerance for comparing atom positions.
    * @return True if the algorithm succeeded, and false if it failed.
    */
-  bool reduceCellToAsymmetricUnit(unsigned short hallNumber, double cartTol = 1e-5);
+  bool reduceCellToAsymmetricUnit(unsigned short hallNumber,
+                                  double cartTol = 1e-5);
+
+  /**
+   * Call this function when you wish to merge all undo commands.
+   * It turns on interactive mode to merge similar undo commands in a series
+   * (in order to save space), and it uses QUndoStack's beginMacro() to merge
+   * dissimilar undo commands together. You must call endMergeMode() to end
+   * the merging section (undo and redo are unavailable while merge mode is
+   * on).
+   * @param undoName The name of the undo command
+   */
+  void beginMergeMode(const QString& undoName = "Draw");
+
+  /**
+   * Call this function when you wish merge mode to end. This will turn off
+   * interactive mode, and it will call QUndoStack's endMacro(). All of the
+   * undo commands pushed between beginMergeMode() and endMergeMode() will be
+   * merged into one undo command. beginMergeMode() should have been called
+   * before this is called.
+   */
+  void endMergeMode();
 
   /**
    * @brief Begin or end an interactive edit.
@@ -742,6 +763,18 @@ RWMolecule::bondPairs() const
 inline std::pair<Index, Index> RWMolecule::bondPair(Index bondId) const
 {
   return m_molecule.bondPair(bondId);
+}
+
+inline void RWMolecule::beginMergeMode(const QString& undoName)
+{
+  m_interactive = true;
+  m_undoStack.beginMacro(undoName);
+}
+
+inline void RWMolecule::endMergeMode()
+{
+  m_interactive = false;
+  m_undoStack.endMacro();
 }
 
 inline void RWMolecule::setInteractive(bool b)
--- a/avogadro/qtplugins/editor/editor.cpp
+++ b/avogadro/qtplugins/editor/editor.cpp
@@ -103,27 +103,28 @@ QWidget *Editor::toolWidget() const
 QUndoCommand *Editor::mousePressEvent(QMouseEvent *e)
 {
   clearKeyPressBuffer();
-  if (!m_renderer)
+  if (!m_renderer || !m_molecule)
     return NULL;
 
   updatePressedButtons(e, false);
   m_clickPosition = e->pos();
-
-  if (m_molecule) {
-    m_molecule->setInteractive(true);
-  }
 
   if (m_pressedButtons & Qt::LeftButton) {
     m_clickedObject = m_renderer->hit(e->pos().x(), e->pos().y());
 
     switch (m_clickedObject.type) {
     case Rendering::InvalidType:
+      m_molecule->beginMergeMode(tr("Draw Atom"));
       emptyLeftClick(e);
       return NULL;
     case Rendering::AtomType:
+      // We don't know yet if we are drawing a bond/atom or replacing an atom
+      // unfortunately...
+      m_molecule->beginMergeMode(tr("Draw"));
       atomLeftClick(e);
       return NULL;
     case Rendering::BondType:
+      m_molecule->beginMergeMode(tr("Change Bond Type"));
       bondLeftClick(e);
       return NULL;
     }
@@ -133,9 +134,11 @@ QUndoCommand *Editor::mousePressEvent(QMouseEvent *e)
 
     switch (m_clickedObject.type) {
     case Rendering::AtomType:
+      m_molecule->beginMergeMode(tr("Remove Atom"));
       atomRightClick(e);
       return NULL;
     case Rendering::BondType:
+      m_molecule->beginMergeMode(tr("Remove Bond"));
       bondRightClick(e);
       return NULL;
     default:
@@ -148,14 +151,10 @@ QUndoCommand *Editor::mousePressEvent(QMouseEvent *e)
 
 QUndoCommand *Editor::mouseReleaseEvent(QMouseEvent *e)
 {
-  if (!m_renderer)
+  if (!m_renderer || !m_molecule)
     return NULL;
 
   updatePressedButtons(e, true);
-
-  if (m_molecule) {
-    m_molecule->setInteractive(false);
-  }
 
   if (m_clickedObject.type == Rendering::InvalidType)
     return NULL;
@@ -165,6 +164,12 @@ QUndoCommand *Editor::mouseReleaseEvent(QMouseEvent *e)
   case Qt::RightButton:
     reset();
     e->accept();
+    m_molecule->endMergeMode();
+    // Let's cover all possible changes - the undo stack won't update
+    // without this
+    m_molecule->emitChanged(Molecule::Atoms | Molecule::Bonds |
+                            Molecule::Added | Molecule::Removed |
+                            Molecule::Modified);
     break;
   default:
     break;
@@ -606,7 +611,8 @@ void Editor::atomLeftDrag(QMouseEvent *e)
         RWBond bond = m_molecule->bond(newAtom, clickedAtom);
         if (bond.isValid()) {
           int bondOrder = expectedBondOrder(newAtom, clickedAtom);
-          bond.setOrder(bondOrder);
+          if (bondOrder != bond.order())
+            bond.setOrder(bondOrder);
 
           changes |= Molecule::Bonds | Molecule::Modified;
         }


psavery:
There are names for Draw Atom, Remove Atom, Change Bond
Type, and Remove Bond. Unfortunately, features such as
Draw Bond and Change Atom are difficult to do because
they all start with a left click on the atom. This method needs
to know what action is performed as soon as the button is
pressed. However, I think a "Draw" will be fine for now...

To repeat the bug that I mentioned in the commit, open
Avogadro2 and left click to make an atom. When you undo
all the way back so there are no more undo's available and
then you redo, things start getting very whacky. So, something
is wrong when you undo all the way to the beginning and redo
the first time. I think it might be related to this:

https://github.com/OpenChemistry/avogadrolibs/blob/master/avogadro/qtgui/rwmolecule.cpp#L107

but I'm not completely sure. As I said in the commit, this bug
existed before this commit, so it is not new...

psavery:
And one final thing: draws on slow computers are very slow because
the undo stack functions are repeatedly being called while drawing.
This would be something I think we'd want to fix in the future. On
fast computers, it is not noticeable. However, not everyone is going
to be using a fast computer...

ghutchis:
My big problem is the new methods duplicate effort here. RWMolecule::setInteractive() which seems like it serves the same purpose.

I think the most obvious solution for the slowdown (and large # of undo items) would be to have some internal "immediate" methods for adding atoms, etc. used during interactive edits - obviously the last one would be preserved.

Can you tell if the undo stack is being compressed?

psavery:
A few problems with interactive mode from my point of view are:
1. It doesn't seem to be working right now. Interactive mode is set to be on when drawing starts, and it is set to be off when drawing ends. But the commands are not being merged.
2. It seems to be an unnecessarily complicated feature. There is a separate class commands that use it here: https://github.com/OpenChemistry/avogadrolibs/blame/master/avogadro/qtgui/rwmolecule.cpp#L73. Commands that use it have to inherit from it and have their own mergeWith() functions for merging such as the one here:
https://github.com/OpenChemistry/avogadrolibs/blame/master/avogadro/qtgui/rwmolecule.cpp#L376

Using the beginMacro() and endMacro() features seems to be a much more straightforward way of doing this. Notice that I only really added 6 lines of code to editor.cpp to get it to do this.

http://doc.qt.io/qt-5/qundostack.html#beginMacro

beginMacro() and endMacro() are for lumping undo commands together. I'm not sure if the actual commands get compressed down to take up less memory (I doubt it, but I doubt interactive mode does that as well); however, it DOES from the user's perspective (as in, it only takes one CTRL-Z to undo an entire draw command instead of multiple ones).

psavery:
About reducing the slowdown, I think that sounds right. So either we need to stop using RWAtom and RWMolecule while drawing, or we need to make them NOT update the undo stack while interactive mode is on.

psavery:
I performed a little test. I went to this function here: https://github.com/OpenChemistry/avogadrolibs/blob/master/avogadro/qtplugins/editor/editor.cpp#L415

The "drag" function, and I changed all the RWAtoms and RWBonds to just Core::Atoms and Core::Bonds, and I changed all calls to RWMolecule to calls to its qtgui molecule. Now, the stack does not get updated at all while drawing.

My results were surprising to me. It looks like removing the calls to update the undo functions had no noticeable impact on the speed. So I'm thinking that these calls to the undo functions may not actually be the problem - it's probably something else. I'm not sure what, though. Perhaps there are ways we can speed up the atomLeftDrag() function...

ghutchis:
I think you need some profiling. I don't see the issue (which makes me think it might be a Qt platform issue). But, if you use a sampling technique, you can see what methods are spending the most time:

* [oprofile](http://oprofile.sourceforge.net/examples/)
* [poor man's profiler](https://poormansprofiler.org)
* [callgrind](http://stackoverflow.com/questions/375913/what-can-i-use-to-profile-c-code-in-linux/10945113#10945113)

avogadro/qtgui/rwmolecule.h
```diff
@@ -744,6 +758,19 @@ inline std::pair<Index, Index> RWMolecule::bondPair(Index bondId) const
...
+
+inline void RWMolecule::endDraw()
+{
+  m_undoStack.endMacro();
+  // Let's cover all possibilities
+  emitChanged(Molecule::Atoms | Molecule::Bonds |
+              Molecule::Added | Molecule::Removed);
+}
+
 inline void RWMolecule::setInteractive(bool b)
```
ghutchis: Can't you just modify this, e.g.

setInteractive(bool b, const QString &undoName = "Draw")
{ 
  if (b) m_undoStack.beginMacro(undoName);
  else // ..

ghutchis:
I agree that setInteractive() isn't being used as much as it should. I'm not sure the undo calls in Avogadro are merging as they should. That's a separate bug.

My point is that we don't need to add new methods. Instead, you could add the beginMacro() and endMacro() calls in setInteractive(). Then, when undo commands are properly merging, the undo stack will be smaller, and calls will work. 

My problem with adding specific "draw" methods is that I can imagine other edit features (e.g., auto-optimize) that aren't draw..

psavery:
For drawing slowness:
I tried out some profilers. From callgrind, all I see are LLVM functions - so that isn't all that informative. From doing a few poor man's profiling, I was able to see that the slowness is mostly due to rendering functions (inside avogadro/rendering). So perhaps there is an issue with rendering with Ubuntu.

However, I should clarify that this is being done on my virtual machine. I tried it on my modern Ubuntu desktop, and it appears to be quite fast (unless I'm running other things in the background). We might not consider it a big issue because it is more important for it to be fast on native linux than on virtual machines...

psavery:
For interactive mode:

I've done a lot of research on interactive mode today, and I think things will be a lot clearer (although this comment will likely be very long).

It looks like interactive mode DOES work! This is good news. I figured this out when I used the manipulate tool to move an atom around. It does indeed merge all of the move commands together into one (saving space).

So why doesn't it work for the draw tool? Because of this: https://github.com/OpenChemistry/avogadrolibs/blame/master/avogadro/qtplugins/editor/editor.cpp#L609

Interactive mode can only merge QUndoCommands that have the same dynamic type (see https://github.com/OpenChemistry/avogadrolibs/blob/master/avogadro/qtgui/rwmolecule.cpp#L376 for instance - note the dynamic cast). It wouldn't make sense to combine two different types of QUndoCommands together. It's impossible with the current setup - different undo classes have different data members to save space (a bond undo doesn't store the positions of the atoms, for instance, because it would take up unnecessary space). It should probably be kept that way. But that implies that interactive mode CANNOT combine different types of undo commands - it can only combine several of the same type of command into one.

So how does this relate to our draw situation? Well, every time the atomLeftDrag() function is called, an alternating series of  changeAtomPosition and changeBondOrder get called (note that changeBondOrder gets called every time - not just when the bond order is different - so I changed that in this PR). This means that changeAtomPosition functions can't possibly merge when drawing because they are all separated by changeBondOrder functions - and changeBondOrder functions can't merge because they are all separated by changeAtomPosition functions. So I just updated my commit to include an if() statement before it that ensures the bond order has actually changed before setting it. This allows interactive mode to merge all of the changing atom positions (so long as the bond order isn't changing): line 617 https://github.com/psavery/avogadrolibs/commit/2b4ca175d2e995b2c88168015af19a9900e5d554#diff-2b33a8d62f0a4a77205caae88971abe6R617

Interactive mode is designed to lump together several QUndoCommands that have the same dynamic type. It's useful because it reduces the size of the undo commands. However, as I mentioned, it can only do so for similar commands. Drawing consists of several different types of undo-able commands: creating an atom, creating a bond, moving the atom, changing the bond order, changing an atom type, etc. Thus, interactive mode cannot be used to combine all of these. QUndoStack's beginMacro() and endMacro(), on the other hand, can combine different types of undo commands into one.

Thus, I am now proposing this: if we want to make a single draw action have a single undo command from the user's perspective (I think we do - it makes it more user-friendly to me at least), we need to use both interactive mode AND beginMacro() with endMacro(). Interactive mode allows us to lump together similar commands to save space - and beginMacro() and endMacro() allows us to lump together different commands. The other option that I can think of is to take a snap shot of the whole molecule before and after the draw. This could take up a lot of memory, however, if the molecule is big.

I'm starting to think, though, that having beginMacro() and endMacro() be called separately from interactive mode turning on or off might be a good idea as well. Some features (like the manipulation tool) only perform actions that can be combined with interactive mode (and there is no need for beginMacro() and endMacro()). So I think having a separate function for beginMacro() and endMacro() is not a bad idea. I left it that way in the PR above. We CAN call it something more generic than "beginDraw()" and "endDraw()", however. But something different from interactive mode might be a good idea. 

To me, using beginMacro(), endMacro(), and turning on interactive mode seems to be the best move.

cryos:
I suspect a large source of the slowdown in editing (although I didn't expect it to be this much of an issue, and at points in the past it wasn't as apparent) is that edits currently cause a deletion of the scene, and wholesale regeneration. I think we could reap the largest speed up by being cleverer with updating the scene, and possibly in optimizing the generation of new scenes.

My other question is whether you are seeing these slowdowns in Release/RelWithDebInfo, or just in Debug builds (where I see them too). It is certainly a bug that change bond order got called every time, but it would be good to know if you see this in release mode too. There are a number of functions that get much faster in release mode, knowing more about the build you are talking about would be good. Current master with a small 20-30 atom molecule seems pretty interactive here (maybe I need to go bigger too?)

psavery:
I'm about to update my branch here - I think it'll be a lot cleaner with my update.

cryos:
Sounds good, I would still like to better understand if slowness in editing is observed for stuff in Release mode, I am using RelWithDebInfo and things look good, but in Debug it is pretty slow.

psavery:
For the observed slowness: it has indeed been in debug mode. I should try it in release mode to see if it is faster.

Like I said, though, it's on my virtual machine. And my laptop itself is starting to be more and more aged. So I'm hoping it won't be an issue for most users.

cryos:
Most users will use a released binary, and that is compiled in release mode. If we are only seeing slowness in debug mode then I would advise that you consider RelWithDebInfo, maybe we should change the default. Debug is useful, but when using modern APIs it can get very slow due to all the extra checks added in a number of the calls we make (asserts + other stuff). There is stuff to optimize, but I would hate to divert too much effort to trying to make Debug builds faster.

psavery:
Okay, I updated this. This commit, however, is not directly related with the speed - it's more related to making all draw actions undo-able.

I added a beginMergeMode() function that uses both beginMacro() and turns interactive mode on. Thus, other calls for interactive mode are not necessary when using beginMergeMode() and endMergeMode(). It seems to work nicely - it lumps similar undo commands together with interactive mode (saves space), and it lumps dissimilar undo commands together with beginMacro(). This allows for a single undo command in the GUI for undoing a draw action.

cryos:
I could see that from this branch, but wanted to comment as it was discussed at length above. The changes you have made look great, and are about what I had in mind when I said we could use begin/end macro. :+1: from me, I will let @ghutchis comment in case he still has any remaining concerns.

psavery:
FYI, I think I found the bug I was referring to in this commit. The bug is here:
https://github.com/OpenChemistry/avogadrolibs/blob/master/avogadro/qtgui/rwmolecule.cpp#L108

If we undo all the way down to zero atoms, and then redo once, it doesn't push_back() to the positions vector. I asked Dave about this since he was the one who wrote it. He said this was supposed to be a check to see if we are using positions3d() or not (like in the case of SMILES or other formats where we don't use coordinates). However, it sounds like he missed that size being 0 doesn't necessarily mean we aren't using positions3d - it could also mean we don't have any atoms. So I'll add some other kind of data member in there to fix it (or find another way of doing it) later. I think that's it, though!

cryos:
Merging as this seems reasonable, and we can fix in follow up branches. Trying to catch up and get outstanding PRs merged.

PR #101, opened 2016-08-11
Added undo commands for OpenBabel features.

psavery:


psavery:
However, there does seem to be a problem with the OpenBabel addHydrogens. The hydrogens don't seem to have the right geometry around the atoms...

ghutchis:
Before or after the patch?

psavery:
I'll need to double check when I get back, but I'm pretty sure it is independent of this update. It puts the hydrogens all in the same plane for carbon. Looks like it does trigonal planar and then puts another hydrogen in between two of them

ghutchis:
I suspect the hybridization isn't being set properly then - weird, that worked before. Even if the new hydrogen addition makes this obsolete, I'd like to figure out what's going on (e.g., does it cause problems with other OB integration).

psavery:
I updated OpenBabel to make sure it wasn't an old version - problem still happens. To reproduce the problem: start up Avogadro2. Uncheck "adjust hydrogens." Click on the screen to create a carbon atom, or draw one. Create as many as you want. Go to Extensions->OpenBabel->addHydrogens (or addHydrogensForpH - you run into the same exact problem), when you add the hydrogens, they will all be in the same plane. Strangely enough, if you then go into navigation mode and rotate the camera slightly, then add another carbon and hit "Adjust Hydrogens", it will add them in a proper tetrahedral fashion. So, perhaps it does not realize at first that we are in a 3-D system or something?

So the problem is either here:
https://github.com/OpenChemistry/avogadrolibs/blob/master/avogadro/qtplugins/openbabel/openbabel.cpp#L583

and here (since it is a problem for AddHydrogensForPh as well):
https://github.com/OpenChemistry/avogadrolibs/blob/master/avogadro/qtplugins/openbabel/openbabel.cpp#L583

perhaps it is an issue with the MDL generation.

or it is in OBProcess::convert():
https://github.com/OpenChemistry/avogadrolibs/blob/master/avogadro/qtplugins/openbabel/obprocess.cpp#L199

psavery:
Also, there is a problem with "Optimize Geometry". It appears to assume that there is a correct number of hydrogens in the molecule.

As in, if you have 2 carbon atoms with no hydrogens and you hit "Optimize Geometry", an error will pop up saying that the result it got from OpenBabel has 10 atoms and it should only have 2 (it has 10 atoms because the optimize geometry added 8 hydrogen atoms so that the carbon atoms have a proper valence shell).

cryos:
LGTM

PR #103, opened 2016-08-15
Cclib script

Schamnad:
The cclib script created as a ScriptFileFormat which calls the cclib parser script from the python path. Presently, for the script to work cclib needs to be installed on the host PC.

Usage:
The cclib script is programmed to be available as a reader to any .out, .log,. adfout, .g09 output file. The script generates a standardized cjson file which is parsed internally by avogadro.


cryos:
This branch has all of cclib in it right now as far as I can tell, as I said over email this should be integrated in the superbuild or installed by the user. I want to avoid duplicated and possibly old versions of libraries in Avogadro. This will need to be removed from the commits in the pull request before it is ready to be merged.

I will take a look at the rest of the code soon, but any pull request with over 10,000 lines of changes is a lot to review - I suspect that number will be far lower with cclib removed.

Schamnad:
This commit doesn't include all of cclib, only the files pertinent to generating the cjson format. But you're right, the number of lines of changes in the pull request is extremely large due to cclib, the only file which is new/exclusive to avogadro is the cclibScript.py under scriptfileformats->formatscripts->cclib . I'll create a new Pull request with only this file.

cryos:
Sorry, it doesn't matter if it is all or just pertinent parts - we don't want a copy of cclib in the plugin as it will make the job of maintaining an updated cclib more difficult.

Schamnad:
Closing this one as I have created a new PR based on the comments provided here. 
Haven't updated this one as I did not wish to pollute the commit history.

New PR: https://github.com/OpenChemistry/avogadrolibs/pull/106

PR #105, opened 2016-08-15
Visualize properties

Schamnad:
The data parsed from the cjson has been displayed in the form of two tables. The previously existing 'Molecular Properties...' dialog box has been converted to a drop down list of two items:
1. General Properties
2. Energy Properties


cryos:
I am going to close this, I think we should  extend the properties, and will loop back to this, but don't think we will merge it in its current form.

ghutchis:
Can we at least leave this PR open so it isn't forgotten?

cryos:
If you really want to, but I would have just tracked it as an issue.

ghutchis:
Okay, I'm closing this particular pull request now that there's an issue #329 to track this.

@Schamnad - if you have any feedback, please comment - your thoughts would be *very* welcome after working on this. 😄

PR #106, opened 2016-08-22
Cclib format script

Schamnad:
The cclib integration script created as a ScriptFileFormat which calls the cclib library from the python path. Presently, for the script to work cclib needs to be installed on the host PC.

Usage:
The cclib script is programmed to be available as a reader to any .out, .log,. adfout, .g09 output file. The script generates a standardized cjson file which is parsed internally by avogadro.

To Do: Incorporate the cclib package in the Avogadro superbuild.


cryos:
This looks much better, I will work on getting this ready to test locally before merging.

cryos:
#147 contains these changes

PR #117, opened 2016-11-15
Update element data, fixing element names (translator), mass, radii

ghutchis:
Updated covalent and VdW radii with consistent sets:
Covalent: Pyykko doi: 10.1002/chem.200800987

VdW: Alvarez Dalton Trans., 2013,42, 8617-8636
DOI: 10.1039/C3DT50599E

Updated masses from IUPAC 2015:
http://www.chem.qmul.ac.uk/iupac/AtWt/


cryos:
Does this mean we can't simply parse the bodr data anymore? There are some fixes made by Wibe de Jong that I would like to make sure we get too. I am fine with merging this, but would really like to ensure bodr has what we need, and can be parsed. VTK now has a similar file that should be updated there too.

ghutchis:
I do not think the BODR is maintained. It's great in concept, but I'd much prefer this approach (i.e., we maintain our copy). If you can point me at Wibe's changes, I'll integrate them. I'm also happy to merge a copy to VTK if that's needed.

ghutchis:
This is the same source of covalent radii as he mentioned (https://github.com/egonw/bodr/pull/1), and it's the same set of VdW suggested here (https://github.com/egonw/bodr/issues/8) as a more consistent, comprehensive set. I also updated masses.

cryos:
Sounds good, I was hoping to use a shared resource but if it isn't maintained we can switch. We could always switch back if things improve, I really liked the idea and hadn't realized it was considered unmaintained.

PR #120, opened 2016-11-17
Fix "paste SMILES" - now generates 3D coordinates correctly.

ghutchis:
Previously OB file format would only look for ".smi" instead of
".smiles"


cryos:
Looks good, thanks!

PR #121, opened 2016-11-17
Fix to use HTTPS, new File->Import menu path, and set molecule name.

ghutchis:


cryos:
Looks great, I had been meaning to add this.

PR #124, opened 2016-11-18
Paste to insert

ghutchis:
Test by drawing something, performing Cut, then draw something more and paste.

Need to finish off selection for this to be really worthwhile.


cryos:
Looks good to me, tested locally.

PR #129, opened 2016-11-21
Move our last few Q_WS_ tests to Q_OS_

cryos:
Minor fixes to ensure we execute the right code paths.

### DIFF
--- a/avogadro/qtplugins/coordinateeditor/coordinateeditordialog.cpp
+++ b/avogadro/qtplugins/coordinateeditor/coordinateeditordialog.cpp
@@ -61,10 +61,10 @@ using Avogadro::Vector3;
 namespace {
 
 // Ensure a cross-platform monospaced font
-#ifdef Q_WS_X11
-static const QString EDITOR_FONT = "Monospace";
-#else // Windows and Mac
+#if defined(Q_OS_WIN) || defined(Q_OS_OSX)
 static const QString EDITOR_FONT = "Courier";
+#else // Linux and other OSes
+static const QString EDITOR_FONT = "Monospace";
 #endif
 
 // Various integer constants.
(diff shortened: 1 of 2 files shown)


ghutchis:
This seems to restrictive, since there's also:
Q_OS_FREEBSD
Q_OS_UNIX

I *think* the latter is the best choice?
http://stackoverflow.com/a/34166706/131896

cryos:
To be honest X11 was the better define, none of these are essential but led to better rendering on Linux. I have no idea what these look like on the BSDs, and worried that UNIX would be defined on Mac OS X so I went with the more conservative option.

ghutchis:
Agreed that X11 would be a better define. I'm fine with merging for Linux and letting BSD maintainers decide if we need to declare other defines in a bug fix. (I'd assume it's similar.)

cryos:
I was thinking about this on the drive to work earlier, and I might flip this around - we know that Windows and Mac are good with the more specific font, and then everything else would likely be best using the more generic font specification (including Linux, BSD, etc). I will edit this branch and push an update.

cryos:
@ghutchis I was just fixing this, and noticed that Q_OS_MAC is not defined according to http://doc.qt.io/qt-5/qtglobal.html, and we should use Q_OS_OSX. I see we have a number of Q_OS_MAC checks and wanted to check if this was behaving as expected for you? I will use Q_OS_WIN and Q_OS_OSX in this update.

ghutchis:
Seemed to work OK. I looked in qglobal.h on my computer, and I see things like:
`#ifdef Q_OS_MAC`
So Qt uses the _MAC variant (significantly).

It turns out there are a bunch of compatibility defines:
<pre>
// Compatibility synonyms
#ifdef Q_OS_DARWIN
#define Q_OS_MAC
#endif
#ifdef Q_OS_DARWIN32
#define Q_OS_MAC32
#endif
#ifdef Q_OS_DARWIN64
#define Q_OS_MAC64
#endif
#ifdef Q_OS_OSX
#define Q_OS_MACX
#endif
</pre>

cryos:
That is good to know, weird that they wouldn't document them even as deprecated/compatibility defines.

PR #130, opened 2016-11-21
WIP: Enable POV-Ray export command. Only writes to std::cout

ghutchis:
At a minimum, this needs to pop up a save dialog and write to a POV-Ray file.

I'd like to see the POV-Ray visitor class move from rendering/ to the extension (e.g., so an SVG or VRML variant can use the code as a template).

It's also not clear how complete the POV-Ray visitor class is.

### DIFF
--- a/avogadro/qtplugins/povray/povray.cpp
+++ b/avogadro/qtplugins/povray/povray.cpp
@@ -37,7 +37,7 @@ namespace QtPlugins {
 POVRay::POVRay(QObject *p) :
   Avogadro::QtGui::ExtensionPlugin(p),
   m_molecule(NULL), m_scene(NULL), m_camera(NULL),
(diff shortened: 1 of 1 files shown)


cryos:
It was about half working, so I added the code but didn't expose it. The camera stuff wasn't working as expected, and there were other issues hence me not exposing it, but I merged it so the approach didn't get lost. I am not sure how much value it offers to any end user at this point.

ghutchis:
I think it's worth leaving this here - if only as a flag for "this needs to be finished before 2.0"

cryos:
Certainly, I would like to get it finished up but wanted to let you know where it was and why I didn't expose it in the interface.

cryos:
Closing, merged the other branch, it looks better than I remember. I would like to add a dialog to run POV-Ray at some point, but this is an improvement, and we should make it easy to export these things.

PR #131, opened 2016-11-22
Fix minor issue with GCC 6

barrymoo:
I compiled avogadro today with GCC 6. Built just fine, minus the regex match below.


cryos:
Yeah, I am using GCC 6 too, we need to remove that CMake file, and bump the CMake requirement, but this works in the short term.

PR #132, opened 2016-11-30
Initial version of a selection tool

cryos:
This only supports atom selection at this stage, with some support for
rendering a selection box that is disabled until the support is added to
the scene to determine what primitives are in the box area. This just
toggles atoms, and saves them to a selected atom array. Some basic
support was added to ball and stick to render selected atoms as blue and
a little bigger.

This is intended as a preview, but is far from complete - sorry it took
so long, hopefully I will have a more complete version by year end!


cryos:
I am going to merge this, and get this release tagged shortly. This is more of a preview, but as Geoff said it provides some basic API we can start working with (and still make changes to if necessary).

ghutchis:
One snag with this version - if a hydrogen is selected, the renderer still might not show it because of options. I think if the atom is selected, it should always be visible.

cryos:
I would have expected them to be hidden selected or not, but it wouldn't be hard to adapt the code. Right now it doesn't do much of anything useful, but I can take a look at making a change, perhaps adding another setting for show all selected atoms would also be appropriate. I think there is one more small patch related to changes in Visual Studio 2015 runtime being split, and then I will tag so that we can move forward (there is a bunch of redundant code I also want to remove as it is no longer needed in the rendering).

PR #136, opened 2016-12-07
Importpqr

cryos:


barrymoo:
There is an issue with merging this into master. Some of the commits aren't carried causing my build to fail. Compare https://github.com/OpenChemistry/avogadrolibs/pull/136/files#diff-949aae7ae8ca947cdb655f436bdd5f63R218 with https://github.com/OpenChemistry/avogadrolibs/blob/2c633cb89dc14773a1d5e4c5aa29bcceb0ccb998/avogadro/qtplugins/importpqr/pqrrequest.cpp#L218

cryos:
The only difference I see is the nullptr keyword, which was added in a recent commit. We now require C++11 for master, and nullptr is one of the new keywords. I take it your compiler doesn't support C++11, and is failing?

cryos:
I see what happened now, the code was incorrectly checking for null on a function that returns an int. My laptop machine doesn't have webview available (Ubuntu doesn't seem to package it) and so I missed this. Sorry, one line patch incoming, everything else seemed to work.

barrymoo:
I didn't diff the files, so that might be the only thing missing. Using `nullptr` isn't valid there because `string[index]` returns a character. If I change it to `== NULL` or `== '\0'` it works. I am using GCC 6.2.1.

barrymoo:
The PR fix must be in github limbo. Thanks @cryos

cryos:
https://github.com/OpenChemistry/avogadrolibs/blob/master/avogadro/qtplugins/importpqr/pqrrequest.cpp#L218 shows checking for 0, string arrays are 0 terminated, which is all NULL evaluated to in that instance. This is one of the major reasons nullptr was added, to avoid the ambiguity between a true null pointer and a 0 or false value. This should work fine now, thanks for reporting it.

PR #139, opened 2016-12-16
An initial set of selection commands.

ghutchis:


cryos:
This works well locally, but I end up with a select menu at the top level that looks like this:

image

This is with avogadroapp at master, things get worse as I said in the review of the menu builder branch when switching to that branch (or stay the same in the case of this menu).

cryos:
I hadn't appreciated you want a separate selection menu, so this seems fine although I think the QAction with a - looks odd - does this look different in Macs? The functionality worked as expected for me.

ghutchis:
It shows up as a separator. I'll try to sort out the UI issues.

I wanted a selection menu, because I think there are a bunch of useful tools for selections:
* Enlarge selection by X Angstroms
* Select all Fe atoms
* Select all Tyr residues
* Select all water molecules
* Save selection (e.g., for QM fragment calculations)

Chris Wilmer here was pointing out that CAD programs have a lot of useful tools for operations on selections - adding, subtracting, intersections, etc.

cryos:
Thought I had commented on this, it sounds great and he makes a good point.

ghutchis:
I'm not 100% sure this commit fixes the problem or if I need to handle some with the menubuilder branch in avogadroapp.

ghutchis:
Part of the problem in the screenshot you give, is that menu builder will sort menu items alphabetically, which is a disaster. Plugins should be able to supply actions in an ordered list, which is the point of that branch.

cryos:
That is fair enough, but the duplicates it creates at present on Linux are pretty disastrous too. I can try debugging that further, but I think the root cause is that Qt doesn't guarantee that OS plugins won't change the user visible text and that it can't be used as a unique identifier. I will give this branch a try now.

PR #140, opened 2016-12-20
Enhance for selections

ghutchis:
Add support in various features for selection:

* Bond or unbond only selected atoms
* Move only selected atoms

### DIFF
--- a/avogadro/qtplugins/bonding/bonding.cpp
+++ b/avogadro/qtplugins/bonding/bonding.cpp
@@ -96,14 +96,16 @@ void Bonding::bond2()
       radii[i] = 0.0;
   }
 
+  bool emptySelection = mol->isSelectionEmpty();
+
   // Main bond perception loop based on a simple distance metric.
   for (Index i = 0; i < m_molecule->atomCount(); ++i) {
-    if (!m_molecule->atomSelected(i))
+    if (!emptySelection && !m_molecule->atomSelected(i))
       continue;
 
     Vector3 ipos = m_molecule->atomPositions3d()[i];
     for (Index j = i + 1; j < m_molecule->atomCount(); ++j) {
-      if (!m_molecule->atomSelected(j))
+      if (!emptySelection && !m_molecule->atomSelected(j))
         continue;
 
       double cutoff = radii[i] + radii[j] + tolerance;
--- a/avogadro/qtplugins/bonding/bonding.cpp
+++ b/avogadro/qtplugins/bonding/bonding.cpp
@@ -33,11 +33,14 @@ namespace Avogadro {
 namespace QtPlugins {
 
 using Core::Elements;
+using Core::Array;
+
+typedef Avogadro::Core::Array<Avogadro::Core::Bond> NeighborListType;
 
 Bonding::Bonding(QObject *parent_) :
   Avogadro::QtGui::ExtensionPlugin(parent_),
-  m_action(new QAction(tr("Bond perception"), this)),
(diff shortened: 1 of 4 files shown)


cryos:
This works well locally, but I end up with a select menu at the top level that looks like this:

image

avogadro/qtplugins/bonding/bonding.cpp
```diff
@@ -95,8 +98,14 @@ void Bonding::bond2()
 
   // Main bond perception loop based on a simple distance metric.
   for (Index i = 0; i < m_molecule->atomCount(); ++i) {
+    if (!m_molecule->atomSelected(i))
```
cryos: This means I will only get any bonding if all atoms are selected? It seems like if none are selected you might just prefer global bond perception (rather than nothing happening and assuming it is broken). Maybe storing if selection is empty, and only skipping if not empty and the atom is selected would make more sense.

avogadro/qtplugins/bonding/bonding.cpp
```diff
@@ -121,7 +130,30 @@ void Bonding::bond2()
 
 void Bonding::clearBonds()
 {
-  m_molecule->clearBonds();
+  // remove any bonds connected to the selected atoms
```
cryos: This felt asymmetric when I tested. Removing bonds remove from all if nothing was selected but would not add back to all if nothing was selected. Also, removing from selected removed from the any bond that had a selected atom at either end, whereas adding back only added bonds where both ends were selected.

avogadro/qtplugins/manipulator/manipulator.cpp
```diff
@@ -122,19 +124,39 @@ QUndoCommand * Manipulator::mouseReleaseEvent(QMouseEvent *e)
 QUndoCommand * Manipulator::mouseMoveEvent(QMouseEvent *e)
 {
   e->ignore();
+  if (!(m_pressedButtons & Qt::LeftButton))
```
cryos: This feels much more intuitive here

ghutchis: The manipulate code needs  a bunch of work - no zoom or rotate, but this is a start

avogadro/qtplugins/bonding/bonding.cpp
ghutchis: Sorry about that. That's why I added the isSelectionEmpty() method.

avogadro/qtplugins/bonding/bonding.cpp
ghutchis: The first behavior is a bug (fix coming). I have to think a bit about the UX for adding bonds back if only one atom is selected. To me, it makes sense that bonds to a selected atom should disappear. Does it make sense to re-bond if only one atom is selected?

To me, it made the most sense that you would need to select a bunch of atoms to bond only in that set.

Keep in mind that the bond add/remove is undoable.

avogadro/qtplugins/bonding/bonding.cpp
```diff
@@ -96,14 +96,16 @@ void Bonding::bond2()
       radii[i] = 0.0;
   }
 
+  bool emptySelection = mol->isSelectionEmpty();
```
cryos: I think you mean m_molecule - this doesn't compile for me

ghutchis: Grr, sorry about that - pushed before compile (didn't) finish.

cryos:
Looks good to me now.

PR #144, opened 2016-12-21
Save and read atom selections to CJSON.

ghutchis:
As part of the "atoms" dictionary, e.g.

    "selected" : [ 1, 1, 0, 0, 0, 0, 0, 0 ]

This will allow input generators and other scripts to get the selected atoms if they want/need them.

### DIFF
--- a/avogadro/io/cjsonformat.cpp
+++ b/avogadro/io/cjsonformat.cpp
@@ -143,6 +143,18 @@ bool CjsonFormat::read(std::istream &file, Molecule &molecule)
     return false;
   }
 
+  Value selected = atoms["selected"];
+  if (!selected.empty()
+    && selected.isArray()) {
+      if (selected.size() && atomCount != static_cast<Index>(selected.size())) {
+        appendError("Error: number of selected atoms != number of atoms.");
+        return false;
+      }
+      for (Index i = 0; i < atomCount; ++i) {
+        molecule.setAtomSelected(i, selected.get(i,0).asBool());
+      }
+    }
+
   Value coords = atoms["coords"];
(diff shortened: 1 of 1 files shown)


ghutchis:
I think a more substantial CJSON spec will be needed soon. I will make an initial after Christmas.

cryos:
What form do you want a spec in? I am working on a paper with Bert documenting our work on JSON for chemical representations, and it would be great to know what you mean so that might be more useful. We are certainly working on fleshing out the details, and I have talked with Stuart about chemical JSON and we set up the repository as a project neutral place to flesh out such a spec.

ghutchis:
I think most people would be fine with a set of Markdown (or similar) expressing:

- All required attributes
- Optional attributes

I'm thinking of something similar to http://opensmiles.org/opensmiles.html

For example, in the Python workflow code, we realized some molecules only have fractional coordinates. Similarly, we discussed the optional 3DTuple attribute..

* Another example - if something returns atoms, but no bonds section, is that OK? 
* Should all bond orders be specified?
* How do you specify an overall charge on a molecule (e.g., benzene anion)?

This came up during GSoC and again with Paul and the workflow code. Having more concrete documentation about what's required, what's common, what's optional is very useful.

cryos:
This is great feedback, and something I can work on in the chemicaljson repo, and we could edit/patch up there. Thanks Geoff, I will see if I can get something basic together soon. We want to have this paper finished up by the end of the calendar year, and this would be a good document to author.

ghutchis:
Seems like you're already writing the same bits for the manuscript. Thanks.

PR #145, opened 2016-12-21
Initial pass, bump to CMake 3.3 minimum

cryos:
This removes the local copy of the generate_export_header code, and
moves to more modern method of specifying default visibility. Also
require C++11, and turn off any compiler specific extensions.


cryos:
@ghutchis this along with the other PR for avogadroapp would require (and enable) C++11 support everywhere. I thought I would do some of the conversion in a separate pull request once we have tested this out but it works well for me on Windows, Mac and Linux.

PR #146, opened 2016-12-27
Add workflow scripts and refactor InputGenerator

ghutchis:
Currently does not properly update the QtGui molecule for changes or use the undoMolecule. I'll add a comment.

### DIFF
--- a/avogadro/qtgui/interfacescript.cpp
+++ b/avogadro/qtgui/interfacescript.cpp
@@ -14,16 +14,19 @@
 
 ******************************************************************************/
 
-#include "inputgenerator.h"
+#include "interfacescript.h"
 
 #include <avogadro/core/coordinateblockgenerator.h>
 #include <avogadro/core/molecule.h>
 
 #include <avogadro/io/fileformat.h>
 #include <avogadro/io/fileformatmanager.h>
+#include <avogadro/io/fileformatmanager.h>
 
 #include <avogadro/qtgui/generichighlighter.h>
 #include <avogadro/qtgui/pythonscript.h>
+#include <avogadro/qtgui/molecule.h>
+#include <avogadro/qtgui/rwmolecule.h>
 
 #include <QtCore/QJsonDocument>
 #include <QtCore/QJsonArray>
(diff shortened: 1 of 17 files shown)


ghutchis:
The 2nd half of the project (i.e., refactor the quantum scripts to use the new base class) is hard for me to do, since Molequeue won't link on Mac 10.12.

The new widget provides the central "form" and the updated quantum code will need to handle the preview text, etc.

avogadro/qtplugins/workflows/scripts/test.py
```diff
@@ -0,0 +1,171 @@
+"""
```
ghutchis: Obviously I need a real workflow-centered test.

avogadro/qtgui/interfacescript.cpp
```diff
@@ -145,7 +161,74 @@ void InputGenerator::reset()
...
+    Io::FileFormatManager &formats = Io::FileFormatManager::instance();
+    QScopedPointer<Io::FileFormat> format(formats.newFormatFromFileExtension(
+                                            "cjson"));
+    // convert the "cjson" field to a string
+    QJsonObject cjsonObj = obj["cjson"].toObject();
+    QJsonDocument doc(cjsonObj);
+    QString strCJSON(doc.toJson(QJsonDocument::Compact));
+    if (!strCJSON.isEmpty()) {
+      result = format->readString(strCJSON.toStdString(), mol);
+      // TODO: need to indicate changes to the QtGui Molecule
```
ghutchis: mol is a Core::Molecule as a parameter, and I cannot seem to cast to a QtGui::Molecule here - I always get NULL.

Ideas or fixes are *very* welcome

cryos:
> The 2nd half of the project (i.e., refactor the quantum scripts to use the new base class) is hard for me to do, since Molequeue won't link on Mac 10.12.

I updated my Mac to Sierra to try and fix that issue, but it builds/links fine here.

cryos:
This fails to compile here in the molequeue directory:

/home/marcus/src/openchemistry/avogadrolibs/avogadro/molequeue/inputgeneratorwidget.h:22:28: fatal error: inputgenerator.h: No such file or directory
 #include "inputgenerator.h"

cryos:
This builds when turning off MoleQueue, if you want to merge this earlier then I would suggest simply leaving the inputgenerator.* files in place and refactoring in a future branch. I was able to compile by turning off MoleQueue, but I don't see any entries to test the functionality. 

I turned on INSTALL_TEST_SCRIPTS, but then spotted that the CMake variable was incorrect. I just pushed a commit to your branch that fixes it, and the scale operation seems to work as expected. It looks really nice.

ghutchis:
I was intending to re-factor the input generators in a separate push - not sure why the files were missing, but it was easy to revert.

PR #151, opened 2016-12-29
Automatically select all appended atoms

ghutchis:
This ensures that a paste or insert will always select new atoms.


ghutchis:
If you're getting menubuilder bugs, please file them, thanks.

PR #155, opened 2016-12-30
Move to build menu

ghutchis:
Minor UI adjustments (e.g., standardize the Edit menu with UI guidelines)


ghutchis:
I'm *not* sure if we're including too many commands with the same & characters - I can't see them on Mac.

PR #160, opened 2017-01-02
Fixed compiler error

cryos:
This code should never have used NULL, but checked for 0 as it is not
dealing with pointers.

Change-Id: If0247a2a00c0bcdcecd3df5573a0df8a7746f2bb


cryos:
@barrymoo this fixes your compile error.

PR #168, opened 2017-02-17
Added move semantics to Avogadro::Core::Molecule

psavery:
Added move copy constructor and move assignment operator
to the Molecule class.

### DIFF
--- a/avogadro/core/molecule.cpp
+++ b/avogadro/core/molecule.cpp
@@ -54,6 +54,33 @@ Molecule::Molecule(const Molecule &other)
   }
 }
 
+Molecule::Molecule(Molecule &&other) noexcept
+  : m_graph(std::move(other.m_graph)),
+    m_graphDirty(std::move(other.m_graphDirty)),
+    m_data(std::move(other.m_data)),
+    m_customElementMap(std::move(other.m_customElementMap)),
+    m_atomicNumbers(std::move(other.atomicNumbers())),
+    m_positions2d(std::move(other.m_positions2d)),
+    m_positions3d(std::move(other.m_positions3d)),
+    m_coordinates3d(std::move(other.m_coordinates3d)),
+    m_hybridizations(std::move(other.m_hybridizations)),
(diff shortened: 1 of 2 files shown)


psavery:
Hi Marcus!

I have been reading about C++11, and I think the move semantics are quite interesting.

I remember that you said you were going to switch Avogadro2 over to completely use C++11. And I think I can see that from the use of nullptr.

I was wondering, have you considered adding some functions like move constructors and move assignment operators in? This can automatically speed up the program when a temporary object needs to be copied, and it can make it easier to transfer data if a move needs to be performed.

For example, I did one here for Molecule. It may not be 100% perfect, but I think it is the kind of thing we would go for.

For the move constructor, it basically calls the move constructor for every data member except for the pointers (if there is not a move constructor, it will call the copy constructor instead). Pointers get "moved" manually.

For the move assignment operator, it cleans up what needs to be cleaned and then, similar to above, "moves" all of the data members via their move assignment operators to "this".

This can speed the program up if we ever generated a temporary Molecule (something like Molecule mol = generateMolecule(), or there may be some cases already in Avogadro2 where a temporary is generated and copied), and it can make it easier to transfer data from one molecule to another (mol = std::move(oldMol)).

Anyways, just thought I'd get your thoughts.

psavery:
Oh yes, and I just remembered the area where it would probably matter most. If we had a vector full of objects (such as Molecule) and the vector needed to resize (such as push_back() being called and it is already at full capacity), then it can be a lot more efficient to "move" all of the items to the newly created vector than to do the traditional copy. And it can "move" all the items if there is a noexcept move constructor in the class...

cryos:
Yes, as stated on the mailing list I moved to require C++11 last year after the 1.90.0 tag was created. I then did some simple replacements for the macros used for override, nullptr, etc. I think we should add move semantics to classes where it makes sense, as they are new I would like to make sure we get it right (I have not written many).

psavery:
Indeed, the move functions were declared 'noexcept', so I think noexcept should be true.

PR #169, opened 2017-02-20
Copy and delete cubes in Molecule correctly

psavery:
This potentially fixes a memory leak in the Molecule
class for when the destructor is called and the cubes
are never deleted.

It also fixes any issues with copying the cubes via
the copy constructor or assignment operator.

### DIFF
--- a/avogadro/core/molecule.cpp
+++ b/avogadro/core/molecule.cpp
@@ -48,9 +48,15 @@ Molecule::Molecule(const Molecule &other)
     m_unitCell(other.m_unitCell ? new UnitCell(*other.m_unitCell) : nullptr)
 {
   // Copy over any meshes
-  for(Index i = 0; i < other.meshCount(); ++i) {
+  for (Index i = 0; i < other.meshCount(); ++i) {
     Mesh *m = addMesh();
     *m = *other.mesh(i);
+  }
+
+  // Copy over any cubes
+  for (Index i = 0; i < other.cubeCount(); ++i) {
+    Cube *c = addCube();
+    *c = *other.cube(i);
   }
 }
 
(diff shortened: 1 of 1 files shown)


psavery:
Also, I noticed that the copy constructor and assignment operator do not copy over all of the data... I wasn't sure if this was intentional or not.

Here are the members that I noticed are not copied:
m_graph
m_data
m_coordinates3d
m_vibrationFrequencies
m_vibrationIntensities
m_vibrationLx

avogadro/core/molecule.cpp
```diff
@@ -52,6 +52,12 @@ Molecule::Molecule(const Molecule &other)
     Mesh *m = addMesh();
     *m = *other.mesh(i);
   }
+
+  // Copy over any cubes
+  for(Index i = 0; i < other.cubeCount(); ++i) {
```
cryos: Should be a space after the for keyword

avogadro/core/molecule.cpp
```diff
@@ -79,6 +85,14 @@ Molecule& Molecule::operator=(const Molecule& other)
       Mesh *m = addMesh();
       *m = *other.mesh(i);
     }
+
+    clearCubes();
+
+    // Copy over any cubes
+    for(Index i = 0; i < other.cubeCount(); ++i) {
```
cryos: As above

cryos:
We should look at the others, I don't think it was intentional.

psavery:
Okay. I will submit that very soon.

Are you okay with the extra commit that was a part of this one? I was trying to get rid of it... it seems to be unnecessary

cryos:
It would have been good to remove the extra commit, but it is merged now and not critical.

PR #171, opened 2017-02-25
Render molecule previews as PNG data instead of QWebEngineView

etp12:
@ghutchis 
Removes QWebEngineView/Chromium as a necessity for importing molecules via PQR by rendering molecule previews as a PNG.

Screenshot:

### DIFF
--- a/avogadro/qtplugins/importpqr/pqrrequest.cpp
+++ b/avogadro/qtplugins/importpqr/pqrrequest.cpp
@@ -51,15 +51,7 @@ void PQRRequest::sendRequest(QString url)
 */
 void PQRRequest::sendRequest(QString url, QString mol2)
 {
-  QUrl httpRequest(url);
-  QNetworkRequest request;
-  request.setRawHeader("Accept", "text/html,application/xhtml+xml,application/xml;q=0.9,image/webp,*/*;q=0.8");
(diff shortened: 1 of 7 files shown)


avogadro/qtplugins/importpqr/pqrrequest.cpp
```diff
@@ -65,6 +65,24 @@ void PQRRequest::sendRequest(QString url, QString mol2)
...
 
 /**
+* @brief Sends a network request to download a png form PQR
+* @param url The url to send the request to
+*/
+void PQRRequest::sendPNGRequest(QString url)
+{
+  QUrl httpRequest(url);
+  QNetworkRequest request;
+  request.setRawHeader("Accept", "text/html,application/xhtml+xml,application/xml;q=0.9,image/webp,*/*;q=0.8");
```
ghutchis: Why is all of this here? I don't think we need to add all these headers.

etp12:
@ghutchis Should be fixed now.

cryos:
This looks like a much simpler approach. Could you squash away all the 'Merge pull request' commits if possible too? In general this shouldn't be needed, if you are working locally I would suggest a rebase, and if you see conflicts before the final merge then a single merge at the end should be more than enough.

etp12:
Sure, I also went ahead and squashed together my three formatting commits into one to remove some of the clutter.

PR #175, opened 2017-03-13
Improved Array assignment operator efficiency

psavery:
In the Array assignment operator, an unnecessary
copy of d->data was being made when d->ref() >= 2
because of the use of the detach() function. This
copy of d->data was almost immediately over-written
in the assignment operator.

Adding a detach() function that does not copy the data
fixes this problem. It is detach_noCopy(). This function
is now called instead of detach() when d->data() is about
to be cleared or overwritten since the extra copy is
unnecessary.

Note: we could also do this by adding a parameter with a
default to detach() (such as detach(bool copy = true)).
However, the detach() is inline anyways, so it probably
produces less machine code with a separate, shorter
inline function.

### DIFF
--- a/avogadro/core/array.h
+++ b/avogadro/core/array.h
@@ -154,12 +154,18 @@ public:
    * normally necessary, but can be useful when you want to ensure you have a
    * copy of all data.
    */
+  void detachWithCopy();
+
+  /**
+   * Explicitly detach from any other implicitly shared containers. This
+   * version does not copy the data.
+   */
   void detach();
 
   /** Retrieve a pointer to the underlying data. */
   T* data()
   {
-    detach();
+    detachWithCopy();
     return &d->data[0];
   }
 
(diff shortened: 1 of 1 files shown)


psavery:
Let me know what you think about this.

psavery:
I just fixed it to call d->data.clear() right after detach_noClear() in Array::clear() because I realized that if d->ref() == 1, d->data.clear() still needs to be called...

cryos:
I think keeping it inline should be preferred, and agree it would be good to avoid copies that will never be used.

psavery:
Okay, I updated it.

Indeed, this will only have an efficiency improvement when d->ref() > 1, so that may not be all that often. Plus it probably wouldn't be noticeable on small vectors. But it may be on large vectors.

I ran into this when I was looking at this class and trying to figure out how to add move operations. I think move operations may be helpful also.

PR #181, opened 2017-03-30
Write Pov-Ray to file

etp12:
This is related to #130 and enables users to export the current molecule as a Pov-Ray file. This adds a QFileDialog for users to select an output file/directory that defaults to the user's home directory:

It was mentioned that povrayvisitor wasn't working correctly, specifically there were issues with the camera. However it seems to work well for spheres and cylinders. For example: 

As well as when I rotate the molecule:

I tested this with a number of molecules, but if I missed a specific case that has issues I can take some time to look into it.

Originally I was planning on passing a stream for povrayvisitor to write to, but in the end I figured it would be easier to change PovRayVisitor::end() to simply return a string and allow the caller to deal with writing to the file. I can change this if it's not preferred.

### DIFF
--- a/avogadro/qtplugins/povray/povray.cpp
+++ b/avogadro/qtplugins/povray/povray.cpp
@@ -36,8 +36,8 @@ namespace QtPlugins {
 
 POVRay::POVRay(QObject *p) :
   Avogadro::QtGui::ExtensionPlugin(p),
-  m_molecule(nullptr), m_scene(nullptr), m_camera(nullptr),
-  m_action(new QAction(tr("Render with POV-Ray"), this))
+  m_molecule(NULL), m_scene(NULL), m_camera(NULL),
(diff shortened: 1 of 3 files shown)


avogadro/qtplugins/povray/povray.cpp
```diff
@@ -36,8 +38,8 @@ namespace QtPlugins {
 
 POVRay::POVRay(QObject *p) :
   Avogadro::QtGui::ExtensionPlugin(p),
-  m_molecule(nullptr), m_scene(nullptr), m_camera(nullptr),
-  m_action(new QAction(tr("Render with POV-Ray"), this))
+  m_molecule(NULL), m_scene(NULL), m_camera(NULL),
```
cryos: This should remain nullptr, NULL is deprecated

cryos:
What can I say, I am a perfectionist...

PR #186, opened 2017-04-15
Dalton plugin

cryos:


ghutchis:
Why did you push this? I didn't have time to do it, but I'd much rather solve this on the Python side (e.g. a method that applies Dalton style to the atoms after sorting).

ghutchis:
Oh, I see, this supersedes the other branch. Still, I'd rather handle in Python.

cryos:
I think there is a decent argument for doing this on the C++ side, although not in the way chosen. It can be handled on the Python side, but I wasn't aware of any other work, and want to move it forward. Even if we refactor, I wanted to understand their needs better and test it out.

ghutchis:
I'm closing this - it's solved by #200

Issue #185, opened 2017-04-15
Symmetry plugin attempts to detect symmetry for every molecule loaded


cryos:
This needs to be converted to do so lazily when the dialog will be shown, and we should audit other plugins to confirm they are lazily evaluating the molecule only when they will be shown/do something. On large molecules this hangs the application.

cryos:
This is now resolved,

PR #187, opened 2017-04-16
Symmetry tweaks

cryos:


cryos:
This addresses issue #185, primarily aiming to only evaluate symmetry when someone will look at it. Also some other changes related to only initializing classes when they will be used.

cryos:
@ghutchis any objections to any of these changes?

PR #190, opened 2017-04-24
Export to VRML

etp12:
This PR allows for spheres and cylinders to be exported to VRML (.wrl) format, similar to the Pov-Ray renderer. It is available under File->Export->VRML Render.

Example molecule in the Avogadro scene: 

After exporting to VRML:

Unlike the Pov-Ray renderer, this currently doesn't change the camera orientation. In the Pov-Ray format it looked as simple as listing the cameraX/Y/Z coordinates, but in VRML it uses a single quaternion orientation. I was able to calculate this [here](http://cgvr.informatik.uni-bremen.de/teaching/vr_literatur/Calculating%20VRML%20Viewpoints.html) after printing out the cameraY/Z vectors. I wasn't able to find the original C version they used (looks like it was written in 1995), although I was able to look over the Javascript version they use on their site. 

Is the camera orientation important for this feature? I wouldn't think it would be necessary for simply making 3D models, but if it's going to be used similarly to the Pov-Ray renderer then I can look into recreating how the site calculates the orientation.

### DIFF
--- a/avogadro/qtplugins/vrml/vrml.cpp
+++ b/avogadro/qtplugins/vrml/vrml.cpp
@@ -1,0 +1,98 @@
+/******************************************************************************
+
+  This source file is part of the Avogadro project.
+
+  Copyright 2014 Kitware, Inc.
+
+  This source code is released under the New BSD License, (the "License").
+
+  Unless required by applicable law or agreed to in writing, software
+  distributed under the License is distributed on an "AS IS" BASIS,
+  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+  See the License for the specific language governing permissions and
+  limitations under the License.
+
+******************************************************************************/
+
+#include "vrml.h"
+
+#include <avogadro/qtgui/molecule.h>
+#include <avogadro/rendering/camera.h>
+#include <avogadro/rendering/scene.h>
+#include <avogadro/rendering/vrmlvisitor.h>
+
+#include <QtCore/QTextStream>
+#include <QtGui/QClipboard>
+#include <QtGui/QIcon>
+#include <QtGui/QKeySequence>
+#include <QtWidgets/QAction>
+#include <QtWidgets/QApplication>
+#include <QtWidgets/QFileDialog>
+#include <QtWidgets/QMessageBox>
+
+#include <string>
+#include <vector>
+
+namespace Avogadro {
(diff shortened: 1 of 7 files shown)


avogadro/rendering/CMakeLists.txt
```diff
@@ -40,6 +40,7 @@ set(HEADERS
   transformnode.h
   visitor.h
   volumegeometry.h
+	vrmlvisitor.h
```
cryos: This should be two spaces, not a tab

avogadro/rendering/CMakeLists.txt
```diff
@@ -69,6 +70,7 @@ set(SOURCES
   transformnode.cpp
   visitor.cpp
   volumegeometry.cpp
+	vrmlvisitor.cpp
```
cryos: As above

avogadro/qtplugins/vrml/vrml.cpp
```diff
@@ -0,0 +1,98 @@
...
+{
+  m_scene = scene;
+}
+
+void VRML::setCamera(Rendering::Camera* camera)
+{
+  m_camera = camera;
+}
+
+void VRML::render()
```
cryos: I think this should be called export, or write, but it looks reasonable otherwise.

etp12: Am I able to change this if it's an extension plugin? If so, where do I specify which function should be called when it's selected from the menu?

etp12:
Here's a screenshot of a generated .wrl file with meshes as viewed on [Shapeways](https://www.shapeways.com).

ghutchis:
Ping @cryos - is the formatting okay now? It'd be nice to merge this, so there are a few examples of exporting graphics (e.g., POV-Ray, VRML, eventually SVG)...

avogadro/qtplugins/vrml/vrml.cpp
cryos: Sorry, I didn't see the comment. You can call it whatever you want - it is an internal slot. It is only the virtual methods you are overriding that must match, unless I am missing something. This works for now though, we can get it merged. Sorry I missed the notification, I would have replied much sooner!

cryos:
Sorry @ghutchis I had missed that the PR had been updated, merged. Thanks for the reminder.

PR #191, opened 2017-05-27
Fixed a compiling issue with array.h on Windows

psavery:
For some reason, I was experiencing the following error
for some template types for Array on Windows:

"formal parameter with requested alignment of 16 won't be aligned"

and the error occurred on the altered line.

I don't fully know why, but some sources online suggested to use a
reference to fix this issue. So I changed the function to pass the
type by a const reference, and it seems to fix it. It looks like it
might have something to do with it not working on 32-bit systems.

As far as I can tell, it seems like a reasonable fix. Let me know
if you think otherwise, though. This fixed the problems.

### DIFF
--- a/avogadro/core/array.h
+++ b/avogadro/core/array.h
@@ -187,7 +187,7 @@ public:
     d->data.reserve(sz);
   }
 
-  void resize(const size_t& sz, ValueType t = ValueType())
+  void resize(const size_t& sz, const ValueType& t = ValueType())
   {
     detachWithCopy();
     d->data.resize(sz, t);


cryos:
What compiler are you seeing this with? I haven't noticed it, to be honest I think dropping support of 32 bit Windows is reasonable at some point too. I don't have any specific objection to this patch, but would like to know a little more. I would like to require VS2015 on Windows as that implements pretty much all of C++11.

psavery:
Oh sorry. I was using VS2015.

Is there anything special that you have to do to make it compile 64-bit instead of 32-bit? I didn't specify anything so it probably went with the default.

cryos:
You need to select the 64 bit generator, it will default to 32 bit. Several projects have dropped 32 bit support, and we have analyzed download numbers noting that 32 bit downloads were very small for most projects in the last year or two.

PR #193, opened 2017-07-15
Fixed incorrect header guard

foowhiz:


cryos:
Looks good, can you amend your commit message please. General advice is to use a short email subject type line, no full stop/period, such as

Fixed incorrect header guard

I think that would be enough in this case, the diff would show me the rest. If you need more details leave a blank line and then write it much as the body of an email

The header guard was named incorrectly, which could cause issues with...

Keep to 80 characters a line in the main text, I think about 50 is advised for the subject. Good spot, sometimes these minor things sneak through.

foowhiz:
Got it, fixed the commit message. I'll keep this in mind for my next commit.

cryos:
I don't see an updated commit message, and on inspection note that your author information is also incorrect:
```
commit 48becbf101275097c999bb08cb10f41b6cfe94d9
Author: foowhiz <foowhiz@users.noreply.github.com>
Date:   Sat Jul 15 22:49:26 2017 +0530

    Fixed endif comment from AVOGADRO_IO_XMLFORMAT_H to MDL and XYZ respectively
```
You should run through a few git tutorials,

https://git-scm.com/book/en/v2/Customizing-Git-Git-Configuration
https://chris.beams.io/posts/git-commit/

You need to use git commit --amend, probably with set author/email, and then git push -f to force. You should work on topic branches and push them too in future, i.e. not master, but git checkout -b branch-name.

foowhiz:
Working on updating the commit message. Something went wrong and I seem to have reverted my repo to an older version. Closing the pull request, I will update the repo and create a new PR. Thank you for the git tutorials.
If by author info you mean the email address, I did that on purpose to not show my email address in commits.

cryos:
We need real name and email address in commit messages, set up one for commit messages if you wish. I don't think spammers really harvest these, and gmail is set up well to protect it. The name/email address define an author, and we require them for commits.

PR #194, opened 2017-07-15
Pybind11

cryos:


python/main.cpp
```diff
@@ -0,0 +1,21 @@
...
+
+namespace py = pybind11;
+
+void exportCore(py::module& m);
+void exportIo(py::module& m);
+
+const char* hello()
+{
+  return "Ey up...";
+}
```
cjh1: Might want to remove this northern greeting :smile:

PR #196, opened 2017-08-11
Add name property to basis set

cryos:
The name of a basis set is useful if it is a simple basis set applied to
all atoms in a calculation. Add a name property, and set it to the name
of the uniformly applied basis set, or to custom if more than one basis
set is observed in a given file.

### DIFF
--- a/avogadro/quantumio/nwchemjson.cpp
+++ b/avogadro/quantumio/nwchemjson.cpp
@@ -113,7 +113,6 @@ bool NWChemJson::read(std::istream& file, Molecule& molecule)
         if (xcFunctional == "B3LYP Method XC Potential") {
           xcFunctional = "b3lyp";
         }
-        cout << "functional found: " << xcFunctional << endl;
       }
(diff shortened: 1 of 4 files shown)


avogadro/quantumio/nwchemjson.cpp
```diff
@@ -102,6 +104,23 @@ bool NWChemJson::read(std::istream& file, Molecule& molecule)
...
+      if (calcSetup.isMember("exchangeCorrelationFunctional")) {
+        Value functional = calcSetup["exchangeCorrelationFunctional"].get(
+          Value::ArrayIndex(0), "");
+        if (functional.isObject()) {
+          xcFunctional = functional["xcName"].asString();
+        }
+        if (xcFunctional == "B3LYP Method XC Potential") {
+          xcFunctional = "b3lyp";
+        }
+        cout << "functional found: " << xcFunctional << endl;
```
cjh1: debug?

cryos: Yes, let me yank that...

PR #197, opened 2017-10-19
Parser for PDB files

foowhiz:
Reads in ATOM and CONECT records, assigns residue bonds based on distance between atoms.


ghutchis:
It's a nice start.

Generally, the solution for residue bonds is to have a dictionary of known bonds in residues:
https://github.com/openbabel/openbabel/blob/master/data/resdata.txt

I think the other point of reading in a PDB file directly, would be to get the residue data, which obviously can't exist until there's a residue class.

ghutchis:
So I think I'll hold off on merging this for a few weeks until I can get a fragment/residue class finished.

foowhiz:
But wouldn't reading the dictionary for each residue take up a lot of time?

ghutchis:
Performing a dictionary lookup (i.e., a hash table) is a constant, so looping through the atoms to assign residue bonds this way is O(N).

Assigning bonds based on atomic distances requires O(N^2) because for each atom, you need to check its neighbors.

In short, on large molecules like proteins, the dictionary lookup is *guaranteed* to be faster than bond perception.

foowhiz:
I see, thank you very much for the answer!

badarsh2:
@ghutchis @cryos I have added a basic residue class implementation, which holds the residue name, residue ID, names of atoms in the residue and also a handler for getting bond information for the residue. I've entered data for ALA alone for now. If the approach sounds good, I'll add the data for rest of the residues. 

P.S. I've signed off the commit but I'm still getting a DCO check failure.

ghutchis:
As I discussed in some of my pull requests yesterday, the DCO requires *every* commit be signed-off. Obviously, that's a pain for existing patches.

avogadro/io/pdbformat.cpp
```diff
@@ -132,7 +159,7 @@ bool PdbFormat::read(std::istream& in, Core::Molecule& mol)
       }
     }
   } // End while loop
-  mol.perceiveBondsSimple();
+  // mol.perceiveBondsSimple();
```
ghutchis: You still need to do this, although CONECT records help, you still want to bond HETATM. The main thing would be to ensure that perceiveBondsSimple() ignores the bonds and bond orders that you've added already.

avogadro/core/residue.h
```diff
@@ -0,0 +1,76 @@
...
+  virtual ~Residue();
+
+  std::string residueName();
+  void setResidueName(std::string& name);
+
+  size_t residueId();
+  void setResidueId(size_t& number);
+
+  void setResidueAtomName(std::string& name, size_t id);
+  std::vector<std::string> residueAtomNames();
```
ghutchis: It would be useful to have a method that returns a vector of the atoms in the residue. For example, this would make it easy to select all atoms in a given residue.

avogadro/core/residue.h
```diff
@@ -0,0 +1,76 @@
...
+
+  std::vector<std::tuple<size_t, size_t, size_t>> resolveResidueBonds();
+
+protected:
+  std::string m_residueName;
+  size_t m_residueId;
+  AtomNameMap m_atomNameMap;
+};
```
ghutchis: We also need to track the chain ID of the residue. Some PDB files contain multiple chains.

avogadro/core/residue.cpp
```diff
@@ -0,0 +1,110 @@
...
+  std::vector<std::tuple<size_t, size_t, size_t>> res;
+  if (m_residueName == "ALA") {
+    bondSeq = {
+      std::make_tuple("CA", "N", 1),   std::make_tuple("CA", "CB", 1),
+      std::make_tuple("CA", "C", 1),   std::make_tuple("C", "O", 2),
+      std::make_tuple("N", "H", 1),    std::make_tuple("N", "HN", 1),
+      std::make_tuple("HA", "CA", 1),  std::make_tuple("HB1", "CB", 1),
+      std::make_tuple("HB2", "CB", 1), std::make_tuple("HB3", "CB", 1)
+    };
+  }
```
ghutchis: I feel like there has to be a better way. First off, can we separate this out into a separate header? Something like a struct with char* char* int, and then an array of that struct. Or maybe you keep this tuple form, but use a macro to simplify the vector<tuple> and move it into the separate file.

This method would just need to index the residue name into that header.

avogadro/io/pdbformat.cpp
ghutchis: Basically, our experience in the Open Babel project (and others) is that CONECT records can't always be trusted, so still run the perceiveBondsSimple at the end here.

cryos:
> As I discussed in some of my pull requests yesterday, the DCO requires every commit be signed-off. Obviously, that's a pain for existing patches. 

Any existing patches can be exempted, if I wasn't clear I think that is reasonable rather than requiring rewriting branches (not that hard, but I don't think we need require it).

avogadro/core/residue.h
```diff
@@ -0,0 +1,76 @@
...
+  /** Assignment operator */
+  Residue& operator=(Residue other);
+
+  /** Destroys the residue object. */
+  virtual ~Residue();
+
+  std::string residueName();
+  void setResidueName(std::string& name);
+
+  size_t residueId();
```
cryos: Should prefer Index for index/id types, as used in other classes.

avogadro/core/residue.cpp
cryos: I am not entirely clear on what this is supposed to do, it looks like a big departure from previous techniques where a lot of objects are being created for Alanine I guess, using a standard sequence.

cryos:
I don't know why you are tying the residue class to this PR, it is probably better to do it in a separate PR, and we can always rebased this PR, or take a PDB reader in a fresh direction. I would develop this with some tests showing the intent of the new residue class.

badarsh2:
Okay, I'll do the same and come back to this branch later then.

badarsh2:
I've now implemented residue instance addition. 
Resolving of atoms with multiple position indicators still needs to be tackled. For now, all position possibilities get rendered but bonds get assigned only to the first occurring indicator atoms as per current 
 residue bond implementation.
I've still kept `perceiveBondsSimple()` disabled because it seems to be adding bonds on top of the bonds derived from residue data. So in some cases, the renderings seem to look quite ambiguous.

ghutchis:
That should be marked as a bug for `perceiveBondsSimple` - certainly it shouldn't add extra bonds to atoms that are bonded.

ghutchis:
I think this is a good start. The changes in #336 will ensure that addBond() and perceiveBondsSimple() do not create bonds where others exist.

I'll leave for @cryos to review this weekend before merging.

ghutchis:
I'm going to merge this. I think it's a useful step forward and can be improved more later.

Thanks to @foowhiz and @badarsh2 for working on this.

PR #198, opened 2017-11-24
Add cpp Travis-CI build

psavery:


psavery:
Hi Marcus!

I thought I would try to assist a little bit at getting the Travis-CI set up.This is by no means a completed .travis.yml script, but I think it is a step in the right direction! It at least compiles avogadro2 successfully!

[Here ](https://travis-ci.org/psavery/avogadrolibs/builds/306561882) is a link to one of my builds with it.

A few things to note:

1. The log is a little too long to display in the regular UI on Travis-CI. You have to click the "Raw Log" to look at the rest. This is partially because of warnings and partially because I am compiling some extra things here such as libarchive (I couldn't get cmake to correctly find libarchive with find_package, so I settled with compiling it for now).

2. It only compiles with GCC 4.8. We can add more if needed in the future.

3. Qt 5.3 is required by one of the submodules (I think Molequeue - which has a part in it that AvogadroApp needs). Qt 5.2 is the newest one on the Ubuntu 14.04 package manager, so I needed to add a ppa to install Qt > 5.3 (ppa:beineri/opt-qt542-trusty).

4. AvogadroApp requires cmake 3.3. Travis-CI only comes with cmake 3.2, so I downloaded (via wget) and extracted a newer cmake for us to use.

5. I needed to update AvogadroApp to the latest commit because I would get a compilation error otherwise.

6. I used "USE_SYSTEM_OPENBABEL=ON" to keep it from compiling Open Babel. However, Open Babel is not on the computer, and if we run tests that require obabel, they will probably fail.

7. No tests are being run. It is just a compilation. But we can add tests in the future I think.

Anyways, let me know what you think.

ghutchis:
I think this is great - having both Travis GCC + Clang builds and Appveyor Windows builds would *hugely* improve pull request reviewing.

ghutchis:
I'm inclined to accept this as-is, and work towards a broader build matrix, e.g.
https://docs.travis-ci.com/user/languages/cpp/#Build-Matrix

I can imagine:
- Trusty GCC-4.x, 5, 6, 7
- OSX Clang-X

cryos:
I think adding env: TASKS="clang-format" and env: TASKS="ctest" would make it easier to know what the tasks are doing, see https://github.com/OpenChemistry/tomviz/blob/master/.travis.yml. So far the macOS builds have generally timed out, and been of no use. I didn't realize the Linux images were so old, we have been using CircleCI for the Linux builds.

I think with the tiny addition to the envs for each build I would merge this in, we can always improve it later as Geoff said.

cryos:
I would also say that while it is great that GCC 4.8 works, we shouldn't shy away from using all of C++11, and that may mean requiring at least GCC 4.9. The experimental PyBind11 needs pretty complete C++11 support, and if expanded out further would require MSVC 2015, can't remember the GCC version off the top of my head.

psavery:
I added the TASKS environment variables to the two runs. I also noticed that "tar -xzvf" on cmake was producing ~5000 lines of output, so I got rid of the "v", and now all of the output shows up in the Travis-CI web UI.

And yes, that makes that we can require GCC 4.9 if we find out that we need it.

cryos:
I moved the submodules forward, so the hack for avogadroapp submodule should not be needed. I assume you were saying we can simply increment the dependency for a more recent GCC, any -v option is usually going to waste space/bandwidth on CI, glad you caught it as it can be hard spotting real errors when things get too noisy.

cryos:
One question that arises, should we even be building avogadroapp for the avogadrolibs repo? It is a dependency of it, but I don't think I would build it by default. I think there are lots of tweaks for future, and maybe that just gets left for then. My take is that avogadroapp repo would build everything, including avogadrolibs, but it is a short build, and possibly a useful sanity check.

psavery:
Reply to the first comment:

Sounds good! I updated it to remove the hack for the avogadroapp submodule.

And sorry, I meant to say "that makes _sense_ that we"

cryos:
There is another reason we might want to bump to a very recent GCC, old GCC's shadow warning was broken, and warns about many things that are not really shadow declarations. We could just do a release build, but if we want compiler warnings worth listening to, then they only fixed it in pretty recent versions.

ghutchis:
I think it makes sense to build avogadroapp here, because most of the development happens here.

There's benefits to having both repos build each-other, kind of like a bootstrapping procedure. My only concern is when there's a need to change API or build with separate PR's for both projects.

PR #199, opened 2017-11-26
Use multiple Linux compilers with Travis-CI

psavery:
Also move the majority of the Travis-CI install and
build scripts to shell files.

### DIFF
--- a/scripts/travis/build.sh
+++ b/scripts/travis/build.sh
@@ -1,0 +1,32 @@
+#!/bin/bash
+
+if [[ $TASKS == "clang-format" ]]; then
+  cd $TRAVIS_BUILD_DIR
+  ./scripts/travis/run_clang_format_diff.sh master $TRAVIS_COMMIT
+else
+  # First, get the super module dir
+  cd ..
+  git clone https://github.com/openchemistry/openchemistry
+  cd openchemistry
+  git submodule init avogadroapp avogadrodata molequeue
+  git submodule update
+
+  # Move the trial avogadrolibs into the open chemistry dir
+  mv ../avogadrolibs .
+
+  mkdir build
+  cd build
+
+  ${CMAKE_EXE} -DCMAKE_BUILD_TYPE=Debug \
+    -DENABLE_TESTING=OFF \
+    -DUSE_SYSTEM_EIGEN=ON \
+    -DUSE_SYSTEM_GLEW=ON \
+    -DUSE_SYSTEM_GTEST=OFF \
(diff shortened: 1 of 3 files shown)


psavery:
A couple of things to note here:

1. I found that it is easier to add more compilers if we move the install and build scripts outside of the include matrix. And I ended up moving them from the .travis.yml to actual shell scripts (like those in [tomviz](https://github.com/OpenChemistry/tomviz/tree/master/scripts/travis)). Note that I am also running them in the current shell (instead of a subshell) because some variables are set in the install script that are needed in the build script.

2. I'm not sure how many compilers would be considered sufficient. Here it uses gcc 4.8, 5.4, 6.3, and clang 3.9. Let me know if that is too many.

And let me know if there are any other concerns with this.

scripts/travis/build.sh
```diff
@@ -0,0 +1,33 @@
...
+  git submodule update
+
+  # Move the trial avogadrolibs into the open chemistry dir
+  mv ../avogadrolibs .
+
+  mkdir build
+  cd build
+
+  # CMake flags
+  export CMAKE_FLAGS="$CMAKE_FLAGS -DENABLE_TESTING=OFF"
```
cryos: There is no need to export these, and using a more compact format would make things more readable if doing this in a script, i.e.
```
${CMAKE_EXE} -DUSE_SYSTEM_EIGEN:BOOL=ON \
  -DUSE_SYSTEM_GLEW:BOOL=ON \
  ..
make -j2
```

cryos:
I think that we need to be careful to choose what is most needed, and so I would skip any mid-point compilers. Having the oldest we think we want to support, and the newest available seems great. Adding latest available clang is nice.

I prefer using scripts as they are more familiar, and we can add in ctest etc more easily in future.

scripts/travis/build.sh
cryos: The type specification is optional, but I use it in most scripts.

psavery:
Okay. I made the modifications that were suggested. It now compiles with GCC 4.8, GCC 6.3, and Clang 3.9.

I removed the CMAKE_FLAGS variable and unnecessary usages of "export". I had the CMAKE_FLAGS variable there because sometimes I've used different flags for different builds. But since we aren't doing that here (at least right now), it's not necessary.

I added the explicit cmake build type of "Debug". I tried it once with "RelWithDebInfo", and the GCC compilations took 3-4 minutes longer.

I also simplified the .travis.yml file a little bit.

PR #201, opened 2017-12-07
Run AvogadroTests for testing

psavery:


psavery:
We might be able to run some tests like this. But notice that a few of the tests are currently failing.

ghutchis:
I'm merging this, so pull requests now require passing compile + tests.

PR #202, opened 2017-12-09
Update element info in tests

psavery:
A few elements were renamed in PR #80, and some element
info (masses, radii) was updated in PR #117. This commit
updates those changes in the tests so that they pass.


psavery:
This update makes it so all the tests in PR #201 pass.

psavery:
I see that the clang-format failed. Do you think that is worth changing? I'll do it if you do.

ghutchis:
No, this looks fine.

psavery:
Thanks Geoff! Looks like the tests in #201 pass now!

Issue #153, opened 2016-12-30
spacegroupdata.h has two identical entries and two invalid caracters


bonfus:
I was trying to (re)use the nice spacegroup data of this project in another code when I noticed that:

- there are two illegal '#' characters in the symmetry operation entries of space_group_transforms
- the symmetry operations of hall numbers 506 and 515 are identical. I suspect that 506 is wrong.
- I also suspect that there is a mistake in the entry 450 (see [bilbao](http://www.cryst.ehu.es/cgi-bin/cryst/programs//nph-trgen?gnum=160&what=gp&trmat=2/3a+1/3b+1/3c,-1/3a+1/3b+1/3c,-1/3a-2/3b+1/3c&unconv=R%203%20m%20:R&from=ita) )

Hope this helps,
best wishes for the new year,
Pietro

ghutchis:
We have been using spglib but we'll definitely check these, thanks.

@psavery - do you have any comments?

psavery:
It looks like he is probably right. Please note that I got this data from OpenBabel (https://github.com/openbabel/openbabel/blob/master/data/space-groups.txt), so anything that is wrong here is also wrong there (I can confirm that the illegal '#' characters are also in OpenBabel).

I will try to fix this soon. I'm probably going to have to write some parsers to aid me (I'm not sure if it matters for the parsers that Bilbao does things like "x+1/2" instead of OpenBabel doing "1/2+x", but I should probably get it to do that just in case).

ghutchis:
@psavery - I'm going through old issues. It seems like Open Babel has updated space-groups.txt that everyone is "happy" with? (At least it hasn't been changed in a few months.)

Can we fix this one?

psavery:
Okay. I need to re-write the script that will parse the space groups file in Open Babel and generate this file from it. I'll try to do that soon.

psavery:
Alright. I think this issue should be fixed now.

ghutchis:
Thanks @psavery

PR #204, opened 2017-12-28
Updated spacegroupdata.h to match current Open Babel data

psavery:


psavery:
This is submitted in response to comments in #153. I wrote a script to find all Hall symbols that we use here in the Open Babel space-groups.txt file and copy over their transformations to the appropriate locations.

Thanks to ClangFormat, it is easy to view the changes.

I've noticed that some of the lines are just small re-orderings. Although there are also several sign changes here and there.

Please let me know if there appear to be any problems with this.

psavery:
Nevermind, I just found a bug. Give me a second to correct it.

psavery:
Okay. As far as I can tell, I think my script worked this time around. Let me know what you think.

ghutchis:
Maybe we can commit the script as well?

ghutchis:
Otherwise, this looks fine.

psavery:
Sorry for the delay. I was traveling around the holidays.

Yes, I will make the script a little more user-friendly and add it in soon.

psavery:
Okay. I added the script. Let me know if there are any other changes needed.

ghutchis:
Looks okay to me.

PR #209, opened 2018-02-02
Added theoretical XRD pattern calculations

psavery:
I added (as a plugin) the ability to calculate and plot a theoretical XRD pattern given a crystal.

This uses an external, static `genXrdPattern` executable to perform the calculation. The source for this program can be found [here](https://github.com/psavery/genXrdPattern).

The majority of the work comes from [ObjCryst++](https://github.com/vincefn/objcryst), though.

VTK is used to generate the plot. As such, this plugin will only be compiled if USE_VTK=ON (which also means that Travis-CI will not currently be testing it...).

Some images from Ubuntu:

image

image
(*Note: All of the dialog options have tooltips)

image

The example shown is diamond. Note the [literature](http://rruff.info/Diamond/R050204) (although we have different x ranges):
image

Note that compiling avogadroapp with -DUSE_VTK=ON only fully worked for me with [this pull request](https://github.com/OpenChemistry/avogadroapp/pull/66).

Also, the ability to interact with the VTK plot is pretty cool! But I think it blocks the Avogadro2 event loop (Avogadro2 will gray out in the background while the VTK plot is up, and it won't respond again until the VTK plot is closed). Maybe I should spawn the VTK plot in a separate thread? Let me know if there is a better option.

The static binaries are downloaded from [here](https://github.com/psavery/genXrdPattern/releases/tag/1.0-static), and they seem to work from the tests I've done on them.

Here are some outputs from some dynamic dependency analysis programs:
For Linux:
```
$ ldd linux64-genXrdPattern 
	not a dynamic executable
```

For Mac (compiled with backward compatibility to OS X 10.9):
```
$ otool -L osx64-genXrdPattern 
osx64-genXrdPattern:
	/usr/lib/libc++.1.dylib (compatibility version 1.0.0, current version 120.1.0)
	/usr/lib/libSystem.B.dylib (compatibility version 1.0.0, current version 1226.10.1)
```

For Windows:
```
>dumpbin /dependents win64-genXrdPattern.exe
Microsoft (R) COFF/PE Dumper Version 14.12.25834.0
Copyright (C) Microsoft Corporation.  All rights reserved.


Dump of file win64-genXrdPattern.exe

File Type: EXECUTABLE IMAGE

  Image has the following dependencies:

    KERNEL32.dll
    msvcrt.dll
    USER32.dll
```

If the cmake option `USE_SYSTEM_GENXRDPATTERN` is set, the executable will not be downloaded.

When Avogadro2 is running, it searches for the `genXrdPattern` executable in the same directory as the `avogadro2` executable, and then it searches in `../bin/`. If the user has set the environment variable `GENXRDPATTERN_EXECUTABLE`, it will use that instead.

Please let me know what you think about all of this. Feel free to question any of my design decisions. I'm sure I can change several things around if needed.

### DIFF
--- a/avogadro/qtplugins/plotxrd/xrdvtkplot.cpp
+++ b/avogadro/qtplugins/plotxrd/xrdvtkplot.cpp
@@ -1,0 +1,101 @@
+/******************************************************************************
+
+  This source file is part of the Avogadro project.
+
+  Copyright 2018 Kitware, Inc.
+
+  This source code is released under the New BSD License, (the "License").
+
+  Unless required by applicable law or agreed to in writing, software
+  distributed under the License is distributed on an "AS IS" BASIS,
+  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+  See the License for the specific language governing permissions and
+  limitations under the License.
+
+******************************************************************************/
+
+#include <vtkAxis.h>
+#include <vtkChartXY.h>
+#include <vtkContextScene.h>
+#include <vtkContextView.h>
+#include <vtkFloatArray.h>
+#include <vtkPlot.h>
+#include <vtkRenderWindow.h>
+#include <vtkRenderWindowInteractor.h>
+#include <vtkRenderer.h>
+#include <vtkSmartPointer.h>
+#include <vtkTable.h>
+#include <vtkTextProperty.h>
+
+#include "xrdvtkplot.h"
+
+namespace Avogadro {
+namespace QtPlugins {
+
+void XrdVtkPlot::generatePlot(
+  const std::vector<std::pair<double, double>>& data)
+{
+  // Save the axes titles
+  const char* xTitle = "2 Theta";
+  const char* yTitle = "Intensity";
+
+  // Create a table and add two columns
+  vtkSmartPointer<vtkTable> table = vtkSmartPointer<vtkTable>::New();
+
+  vtkSmartPointer<vtkFloatArray> arrX = vtkSmartPointer<vtkFloatArray>::New();
+  arrX->SetName(xTitle);
+  table->AddColumn(arrX);
+
+  vtkSmartPointer<vtkFloatArray> arrY = vtkSmartPointer<vtkFloatArray>::New();
+  arrY->SetName(yTitle);
+  table->AddColumn(arrY);
+
+  // Put the data in the table
+  table->SetNumberOfRows(data.size());
+  for (size_t i = 0; i < data.size(); ++i) {
+    table->SetValue(i, 0, data[i].first);
+    table->SetValue(i, 1, data[i].second);
+  }
+
+  // Set up the view
+  vtkSmartPointer<vtkContextView> view = vtkSmartPointer<vtkContextView>::New();
+  view->GetRenderer()->SetBackground(1.0, 1.0, 1.0);
+  view->GetRenderWindow()->SetSize(600, 600);
+  view->GetRenderWindow()->SetWindowName("Theoretical XRD Pattern");
+
+  // Add the chart
+  vtkSmartPointer<vtkChartXY> chart = vtkSmartPointer<vtkChartXY>::New();
+  view->GetScene()->AddItem(chart);
+
+  vtkAxis* bottomAxis = chart->GetAxis(vtkAxis::BOTTOM);
+  vtkAxis* leftAxis = chart->GetAxis(vtkAxis::LEFT);
+
+  // Set the axis titles
+  bottomAxis->SetTitle(xTitle);
+  leftAxis->SetTitle(yTitle);
+
+  // Increase their title font sizes
+  bottomAxis->GetTitleProperties()->SetFontSize(20);
+  leftAxis->GetTitleProperties()->SetFontSize(20);
+
+  // Increase the tick font sizes
+  bottomAxis->GetLabelProperties()->SetFontSize(20);
+  leftAxis->GetLabelProperties()->SetFontSize(20);
+
+  // Adjust the range on the x axis
+  bottomAxis->SetBehavior(vtkAxis::FIXED);
+  bottomAxis->SetRange(data.front().first, data.back().first);
+
+  // Add the data to the chart
+  vtkPlot* line = chart->AddPlot(vtkChart::LINE);
+  line->SetInputData(table, 0, 1);
+  line->SetColor(255, 0, 0, 255);
+  line->SetWidth(2.0);
+
+  // Start interactor
+  view->GetInteractor()->Initialize();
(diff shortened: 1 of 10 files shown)


psavery:
Update: I removed the `return` statement from my macro because a macro [does not handle return like a function does](https://cmake.org/cmake/help/v3.0/command/return.html).

ghutchis:
This is cool - I don't have time to start review, but it's probably worth considering whether something like this could also be done with the not-quite-finished Python plugins + GitHub downloads.

One (dream) question - is is possible to label the peaks based on the Miller planes?

psavery:
I think using it via the plugin downloader is a good idea. I still need to look at the plugin downloader and figure out how to do so (and what I would use for plotting). So I’ll look into it.

I’ll also look into labeling the peaks based on the Miller planes.

psavery:
I haven't looked at labeling peaks based upon the Miller planes, yet. But I have looked at the plugin downloader and have been trying to come up with ideas on how to go about doing this. I came up with 3 ideas, and I'd like to know your (@ghutchis) and @cryos opinions on them. What we determine to do here will probably be what I also try to do for adding yaehmop stuff.

1. Use Python and MatPlotLib. I could write Python scripts to run `genXrdPattern` and MatPlotlib to plot the results. The Python scripts and the `genXrdPattern` executable would both be downloaded via the plugin manager. I guess Avogadro2 would still have some part in it that checks for the existence of these, and if they exist, it may add the drop-down menu option. One benefit of this is that the Python scripts are cross-platform. One negative I can think of is that this would require users to have Python in their path on their computer, and MatPlotLib available via that Python. This may not always be true - especially for Windows users.

2. Build a shared object library (downloaded via the plugin manager) that the Avogadro2 plugin loader can find when it boots. I could potentially use VTK with this shared object library - assuming the shared object library can find the VTK shared object libraries (or it could be statically linked). One benefit of this is that we won't need to add any parts inside Avogadro2 itself to make this work - it would all be in the shared object library. But one negative is that it may be very difficult to maintain - we may need to always compile the shared object libraries for every operating system in the exact same build environment that the Avogadro2 executable is  built in to ensure binary and dependency compatibility.

3. Keep the parts in Avogadro2 that are presented in this PR, and only download the `genXrdPattern` executable via the plugin manager. I can make it so that the drop-down menu for generating a theoretical XRD pattern does not appear unless the executable is found (which would indicate that the user downloaded it via the plugin manager). One benefit in this method is that we can use VTK within Avogadro2 fairly easily for the plotting. A negative is that there has to be some added code within Avogadro2 itself. However, I'd like to note that the code  within Avogadro2 itself would probably not be very extensive or take long to compile.

I'm personally leaning more towards option number 3. The `genXrdPattern` executable is around 5 - 10 megabytes, so it makes sense to not ship that with Avogadro2 by default because not every user would want to use it (and if they did want it, they could download it via the plugin manager). The compiled parts within Avogadro2 that send/receive messages to the `genXrdPattern` executable are relatively small, however, so I don't think it would be too much of an issue to include them by default in the packaging of Avogadro2.

Anyways, let me know what you both think. I also had an idea that perhaps one could send a json-rpc message to Avogadro2 telling it to plot something, and Avogadro2 would generate the plot using VTK and the options set in the json-rpc. But I'm not sure if that would be a good idea or not...

ghutchis:
Sorry, I should have been more clear. I think this PR is fine - certainly for now. I want to make sure that the plugin feature has support for this kind of action (e.g., run something, return an interactive graph). So the discussion I want to have (outside of this PR, e.g., on http://discuss.avogadro.cc) is:

* What would "return data for plotting" look like? I mean, this is an easy example, you return some CSV or something like that, axes labels, maybe some particular point labels (e.g., the Miller planes). But how could a plugin return something like a band diagram or partial density of states.. what kinds of plots would be supported and how would the data be formatted?
* What needs to happen on the Avogadro side - presumably, using VTK for plotting? (e.g., some of the code in this PR)

I think Option 3 is similar, although my intent was more general - that we should be able to write plugins like this with Python scripts instead of all C++. Then the plugin downloader would grab a GitHub repository with the `genXrdPattern` binaries and Python glue.

psavery:
Okay. I [posted](https://discuss.avogadro.cc/t/plotting-strategies-for-plugins-from-the-plugin-downloader/2617) a topic about our discussion here on the Avogadro Discourse.

I will make the clang-format change to this PR right now. We can potentially modify the design of this plugin in the future to make it an optional plugin if we come up with some strategies for generating plots with the plugins.

Is there anything else I need to change in this PR?

cryos:
Looks cool, will try and take a better look at it tomorrow. See my other pull request updating VTK to more recent VTK. The chart code looks good to me, the widget will likely need updating, but I can help with that. I would rather move the download/packaging to the superbuild, but if we want it in here for now I can live with that. Agreed it would be a great external plugin, we might be able to come up with some simple JSON for plotting data - key is array name, array is values to be plotted.

cryos:
> Also, the ability to interact with the VTK plot is pretty cool! But I think it blocks the Avogadro2 event loop (Avogadro2 will gray out in the background while the VTK plot is up, and it won't respond again until the VTK plot is closed). Maybe I should spawn the VTK plot in a separate thread? Let me know if there is a better option.

Only just noticed this, you may be doing something odd with initialization/VTK. The VTK event loop is totally happy to live inside the Qt event loop, and that is how Tomviz/ParaView work. I can help sort that out, like I said, I can take a better look tomorrow. We can produce signals from VTK too, and then have them change things in Avogadro for example.

psavery:
@cryos Sounds good!

avogadro/qtplugins/plotxrd/xrdvtkplot.cpp
```diff
@@ -0,0 +1,101 @@
...
+
+  // Add the data to the chart
+  vtkPlot* line = chart->AddPlot(vtkChart::LINE);
+  line->SetInputData(table, 0, 1);
+  line->SetColor(255, 0, 0, 255);
+  line->SetWidth(2.0);
+
+  // Start interactor
+  view->GetInteractor()->Initialize();
+  view->GetInteractor()->Start();
```
cryos: This needs a few changes, but this is the heart of the issue. I can adapt this, it should use QVTKOpenGLWidget, and that integrates everything into the Qt event loop.

psavery: Oh Okay. Yeah I actually noticed earlier that the window immediately closes on OS X (and doesn't stay open). Hopefully, that will fix the issue on OS X as well.

cryos:
Don't feel like getting fancy with branches, just merging this, and the other, and will fix this up in a new branch to use the Qt VTK integration.

PR #211, opened 2018-02-04
Find Spglib under different name

heirecka:
I'm not entirely sure where the only listed name cmake searches for
originates from, but vanilla upstream only installs a libsymspg.

https://github.com/atztogo/spglib/blob/5a9ba105f4936000b8174093c10dc85ad423ebb0/CMakeLists.txt#L53
https://github.com/atztogo/spglib/blob/5a9ba105f4936000b8174093c10dc85ad423ebb0/src/Makefile.am#L1

### DIFF
--- a/cmake/FindSpglib.cmake
+++ b/cmake/FindSpglib.cmake
@@ -7,7 +7,7 @@
 #  SPGLIB_LIBRARY      - The Spglib library
 #
 find_path(SPGLIB_INCLUDE_DIR spglib/spglib.h)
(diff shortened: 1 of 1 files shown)


cryos:
I think it comes from the version built in the superbuild, I had wondered how the system version would be found, and had assumed it used the same name. We should probably change the superbuild to use the same name as upstream, and get rid of our local name in the future. This works for the short term, thanks for the patch.

PR #212, opened 2018-02-11
Vtk charts

cryos:
This switches to using the Qt support classes, I had a little trouble with the xrd executable, so below shows a quick data set I populated. I was able to interact with both the chart and the 3D molecule as they now live within the same event loop (Qt's event loop).
image

### DIFF
--- a/avogadro/qtplugins/plotxrd/xrdvtkplot.cpp
+++ b/avogadro/qtplugins/plotxrd/xrdvtkplot.cpp
@@ -19,13 +19,15 @@
 #include <vtkContextScene.h>
 #include <vtkContextView.h>
 #include <vtkFloatArray.h>
+#include <vtkGenericOpenGLRenderWindow.h>
+#include <vtkNew.h>
 #include <vtkPlot.h>
-#include <vtkRenderWindow.h>
 #include <vtkRenderWindowInteractor.h>
 #include <vtkRenderer.h>
-#include <vtkNew.h>
 #include <vtkTable.h>
 #include <vtkTextProperty.h>
+
+#include <QVTKOpenGLWidget.h>
 
 #include "xrdvtkplot.h"
 
@@ -58,7 +60,13 @@ void XrdVtkPlot::generatePlot(
   }
 
   // Set up the view
+  vtkNew<vtkGenericOpenGLRenderWindow> renderWindow;
+  QVTKOpenGLWidget* widget = new QVTKOpenGLWidget();
+  widget->SetRenderWindow(renderWindow);
+  // Hackish, but at least it won't leak
+  widget->setAttribute(Qt::WA_DeleteOnClose);
   vtkNew<vtkContextView> view;
+  view->SetRenderWindow(renderWindow);
   view->GetRenderer()->SetBackground(1.0, 1.0, 1.0);
   view->GetRenderWindow()->SetSize(600, 600);
   view->GetRenderWindow()->SetWindowName("Theoretical XRD Pattern");
@@ -92,9 +100,8 @@ void XrdVtkPlot::generatePlot(
   line->SetColor(255, 0, 0, 255);
   line->SetWidth(2.0);
 
-  // Start interactor
-  view->GetInteractor()->Initialize();
-  view->GetInteractor()->Start();
+  // Start the widget, we probably want to improve this in future.
+  widget->show();
 }
 
 } // namespace QtPlugins
--- a/avogadro/qtplugins/plotxrd/xrdvtkplot.cpp
+++ b/avogadro/qtplugins/plotxrd/xrdvtkplot.cpp
@@ -23,7 +23,7 @@
 #include <vtkRenderWindow.h>
 #include <vtkRenderWindowInteractor.h>
 #include <vtkRenderer.h>
-#include <vtkSmartPointer.h>
+#include <vtkNew.h>
 #include <vtkTable.h>
 #include <vtkTextProperty.h>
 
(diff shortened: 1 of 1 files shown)


cryos:
@psavery the chart above is obviously me hacking it up to populate it, but I think this gives us a minimum viable product using the charts. It also shows you the minor changes required to get the chart into a real Qt widget container so that it integrates properly with our event loop (it never starts its own you will notice).

I would like to merge this at some point, it would be good to figure out in future why the binary crashes.

psavery:
It is unfortunate that the binary is crashing on Arch Linux. I built it statically on Ubuntu 16.04. I am also using the executable for the program that our research group uses. I run it in a test on Travis-CI (on Ubuntu 14.04) [here](https://travis-ci.org/xtalopt/XtalOpt/jobs/332368728#L1746), and you can see that it runs successfully.

My laptop is out for maintenance until sometime next week. I will see if I can replicate the crash when I get it back. Maybe I need to change something in the way I am building the binary. But I thought it would work on most flavors of Linux if I built it statically (so that ldd says "not a dynamic executable").

cryos:
Building static binaries can be a pain for Linux, the basic command line seemed to work. I would rather focus on getting the charting working, I tested this on my Ubuntu 16.04 laptop, things look good and the binary works here. I found a small issue (Intel driver is pretty strict). I think this looks pretty good, OK with me merging it to master as it at least works as expected here.

psavery:
On OS X, I'm getting a blank window unfortunately... Not sure what the issue may be.

image

psavery:
But once we get this working, I might try to generalize it in the "vtk" directory and have it take in some kind of json input. That may help us move towards the plugins being able to generate a VTK graph.

psavery:
I tried this on my Ubuntu 16.04 desktop earlier today and it worked fine for that.

psavery:
I was looking into the OS X problem a little more today to see if the issue could be on my end. It might be.

I am working on a late 2012 Mac Mini that has macOS High Sierra on it (10.13.3). I only recently upgraded to Sierra, so I just now made sure all my packages were updated and I recompiled Avogadro2 and tried again. This time, I got a message that is a little more informative when the black window is displayed.

```
Warning: In /Users/patrick/src/openchemistry/thirdparty/VTK/Rendering/OpenGL2/vtkOpenGLRenderWindow.cxx, line 800
vtkGenericOpenGLRenderWindow (0x7f8a2d007400): VTK is designed to work with OpenGL version 3.2 but it appears it has been given a context that does not support 3.2. VTK will run in a compatibility mode designed to work with earlier versions of OpenGL but some features may not work.

ERROR: In /Users/patrick/src/openchemistry/thirdparty/VTK/Rendering/ContextOpenGL2/vtkOpenGLContextDevice2D.cxx, line 908
vtkOpenGLContextDevice2D (0x7f8a32a79430): failed after DrawPoly 1 OpenGL errors detected
  0 : (1280) Invalid enum
```

I'm going to look into this a little more and see if something is going wrong with my OpenGL.

psavery:
I took a closer look at the OpenGL stuff on my Mac, and I don't see it as a problem so far. If anyone else tries this PR out on their Mac, let me know what happens.

graphicsinfo

It's true that I have a cheap graphics card. But if I initialize the OpenGL context (and extensions) using GLEW and GLFW, and I print out the OpenGL info like so:

      const GLubyte *renderer = glGetString( GL_RENDERER );
      const GLubyte *version = glGetString( GL_VERSION );
      printf( "Renderer: %s\n", renderer );
      printf( "OpenGL version supported %s\n", version );

I get this:

    Renderer: Intel HD Graphics 4000 OpenGL Engine
    OpenGL version supported 4.1 INTEL-10.30.14

So it appears that my graphics card supports OpenGL4. I'm not sure why I'm getting that VTK error in the most recent post. If anyone has an idea, let me know.

cryos:
@psavery that GPU is unlikely to be supported well for VTK, it is a capability issue as far as I understand it. That is a nearly six year old chip, and very low end. I was talking to a few people about the Qt OpenGL widget, different OpenGL contexts, and what appear to bugs/features of Qt on macOS. 
```
QCocoaGLContext: Falling back to unshared context.
QCocoaGLContext: Falling back to unshared context.
```
Qt now does pretty much all rendering in OpenGL. I want to try and figure out what it would take to get Avogadro to render on recent Intel chips using the same default context, or a compatible one at least. The main issues as I see them are getting a default context that Avogadro's rendering and VTK's rendering can use with Qt on all three platforms.

psavery:
@cryos Okay. We can just ignore the issues I've been having on macOS, then. I may look into purchasing a newer Mac here sooner or later so I can perform better tests with it.

cryos:
So I spent some more time with a mac, a linux and a windows machine today. I hate Apple sometimes, it would appear that we are a little caught between Apple only offering certain GLSL in OpenGL 3.2 contexts, and what appears to be a bug/OS issue with Qt on Cocoa/mac.

The change will be good for Avogadro, I need to finish testing it, but essentially need to port the shaders to work with GLSL 1.5, declare their version, and a few other bits. This is a little overdue now. I will see if I can get it finished and tested on the major OS versions in the next day or so.

psavery:
Thanks for looking into this, Marcus! That sounds great!

I got my Windows laptop back, on which I have an Arch Linux virtual machine. The `genXrdPattern` executable seemed to run okay on my Arch Linux virtual machine. I might try to just compile the static executable on that Arch Linux virtual machine and test it to see if it runs properly on the other Linux OS's I have access to (CentOS and Ubuntu mainly). After playing around with it a little bit, I may be able to generate an executable that works on most Linux computers.

cryos:
So, more digging, and https://gitlab.kitware.com/vtk/vtk/merge_requests/3745#note_381484 is a branch under consideration for VTK, and it fixes the blank widget seen on macOS for me. It continues to work as expected on Linux too. I think the best course of action is to target this widget for macOS. 

Question on whether we just accept the breakage until the patch is merged into VTK, and move forward, or hold off merging this until that patch is merged...

psavery:
Well, the widget for macOS currently doesn't work on master anyways (it currently appears and disappears immediately). So I don't think it should be a problem if this PR gets merged. But it's up to you.

cryos:
I was leaning towards merging this, and then taking the fix when it gets merged in VTK. Other alternatives are more invasive, and I will keep an eye on the merge request and help where i can to get it merged. I want to get back to the 3D VTK rendering widget too, and look at generalizing VTK integration.

psavery:
That sounds good to me!

I tried compiling the static binary of `genXrdPattern` on Arch Linux with gcc 7.2 and -O2 optimization flags (no -O3). Looks like it works fine on Arch Linux and Ubuntu, but I get an illegal instruction error on CentOS.

From looking at the assembly, it looks like the illegal instruction is a `vmovdqa` instruction.

I'll keep on playing around with it. If I can get it to pass all the tests I give it, I'll probably ask you to try it out as well.

cryos:
Sounds great, merging this for now, and hoping to have good news on the updated VTK OpenGL widget soon.

Issue #226, opened 2018-03-15
Rendering: Labels


ghutchis:
Avogadro v1 offered very flexible atom and bond labels:

**Atoms**

- Atom number
- Element symbol
- Symbol and number in a group (e.g., C3 for the third carbon)
- Partial charge
- Custom data

<img width="515" alt="screen shot 2018-03-14 at 9 53 13 pm" src="https://user-images.githubusercontent.com/41128/37440232-304007a4-27d2-11e8-864d-c182778461ab.png">

ghutchis:
Much of the code is available here:
https://github.com/cryos/avogadro/blob/master/libavogadro/src/engines/labelengine.cpp

ghutchis:
The initial implementation doesn't need to be so complicated - adding atom labels with custom color would be a great start.

ghutchis:
This is addressed by https://github.com/OpenChemistry/avogadrolibs/pull/715

PR #237, opened 2018-03-17
Commit for "Rendering: Labels" (Issue #226)

foowhiz:
#226
Initial implementation for Labels engine. Shows Atom's symbol with its number in the order of atoms.
An issue faced by both Avogadro 1 & 2 is that the Hydrogen atoms are not numbered chronologically. So, if we delete selective atoms, the number of Hydrogen atoms gets messed up.

This commit numbers H atoms based on the non-H atom a particular H atom is connected to. As of now, it does not number free H atoms or H2 molecules. If this commit seems feasible, I will add free H atoms to the end of numbering of H atoms.


ghutchis:
Paging @cryos - can you take a look at this? It would be nice to have atom labels back.

avogadro/qtplugins/labels/labels.cpp
```diff
@@ -0,0 +1,188 @@
...
+    tprop.setAlign(Rendering::TextProperties::HCenter,
+                   Rendering::TextProperties::VCenter);
+    tprop.setFontFamily(Rendering::TextProperties::SansSerif);
+    tprop.setColorRgb(255, 255, 255);
+    atomLabel->setTextProperties(tprop);
+
+    //  Label Hydrogen atoms
+    const NeighborListType bonds =
+      (const_cast<Molecule&>(molecule))
+        .bonds(atom); // const_cast because bonds() accepts non const objects
```
cryos: We should avoid this pattern of comments pushing code to another line, prefer to make comments above the code they are referencing, not on the same line, and especially not if that creates multi-line code that could have been kept on less lines otherwise. We should also look at whether we need another signature to avoid the const_cast.

avogadro/qtplugins/labels/labels.cpp
```diff
@@ -0,0 +1,188 @@
...
+          "H" + QString::number(hydrogenCount, 'f', 0).toStdString());
+
+        hAtomLabel->setRenderPass(Rendering::Overlay3DPass);
+        const Vector3f a1(otherAtom.position3d().cast<float>());
+        const Vector3f& textPos(a1);
+        hAtomLabel->setAnchor(textPos);
+        geometry->addDrawable(hAtomLabel);
+
+        hAtomLabel->setTextProperties(tprop);
+        hydrogenCount++;
```
cryos: Prefer to pre-increment

avogadro/qtplugins/labels/labels.cpp
```diff
@@ -0,0 +1,188 @@
...
+    std::unordered_map<unsigned char, Index>::const_iterator lookUp;
+    lookUp = countMap.find(atomicNumber);
+
+    if (lookUp == countMap.end())
+      countMap[atomicNumber] = 1; // For first occurrence of atom
+
+    else
+      countMap[atomicNumber] += 1; // For other occurrences of atom
+
+    if (!m_showHydrogens)
```
cryos: This looks like nothing will be rendered if showHydrogens is false, which doesn't seem right.

avogadro/qtplugins/labels/labels.cpp
```diff
@@ -0,0 +1,188 @@
...
+      continue;
+
+    Rendering::TextLabel3D* atomLabel = new Rendering::TextLabel3D;
+
+    //  Label non Hydrogen atoms
+    atomLabel->setText(
+      Elements::symbol(atomicNumber) +
+      QString::number(countMap[atomicNumber], 'f', 0).toStdString());
```
cryos: I would think the ideal default (and what was in Avo 1) would be to simply display the symbol. Then we can get fancier with the symbol + its index (this is 1-based, what about 0-based?) Also other options, but they can be left for future commits, such as element name, index in the molecule, etc.

avogadro/qtplugins/labels/labels.cpp
```diff
@@ -0,0 +1,188 @@
...
+    check = new QCheckBox(tr("Show hydrogens?"));
+    check->setChecked(m_showHydrogens);
+    connect(check, SIGNAL(toggled(bool)), SLOT(showHydrogens(bool)));
+    v->addWidget(check);
+    m_setupWidget->setLayout(v);
+  }
+  return m_setupWidget;
+}
+
+void Labels::multiBonds(bool show)
```
cryos: This does not seem at all applicable, and I cannot see where it is used in the rest of the class.

avogadro/qtplugins/labels/labels.h
```diff
@@ -0,0 +1,73 @@
...
+#ifndef AVOGADRO_QTPLUGINS_LABELS_H
+#define AVOGADRO_QTPLUGINS_LABELS_H
+
+#include <avogadro/qtgui/sceneplugin.h>
+
+namespace Avogadro {
+namespace QtPlugins {
+
+/**
+ * @brief Render a molecule in the ball and stick style and display labels
```
cryos: Not ball and stick, comments should be removed or updated.

avogadro/qtplugins/labels/labels.h
```diff
@@ -0,0 +1,73 @@
...
+  Q_OBJECT
+
+public:
+  explicit Labels(QObject* parent = 0);
+  ~Labels() override;
+
+  void process(const Core::Molecule& molecule,
+               Rendering::GroupNode& node) override;
+
+  // void processEditable(const QtGui::RWMolecule &molecule,
```
cryos: Remove

avogadro/qtplugins/labels/labels.h
```diff
@@ -0,0 +1,73 @@
...
+  void multiBonds(bool show);
+  void showHydrogens(bool show);
+
+private:
+  bool m_enabled;
+
+  Rendering::GroupNode* m_group;
+
+  QWidget* m_setupWidget;
+  bool m_multiBonds;
```
cryos: What would multibond mean here?

cryos:
Example of occlusion issue, and what the rendering looks like on a small molecule:
image

foowhiz:
On it.

foowhiz:
Occlusion issue and const_cast still remain. How should I go about the occlusion issue?

ghutchis:
I'm closing this in favor of #715

ghutchis:
Thanks for the effort - this was definitely a good start.

PR #239, opened 2018-03-21
Add libmsym check for cmake

serval2412:


avogadro/core/CMakeLists.txt
```diff
@@ -8,6 +8,9 @@ if(USE_LIBSPG)
   include_directories(SYSTEM ${SPGLIB_INCLUDE_DIR})
 endif()
 
+find_package(Libmsym REQUIRED)
```
ghutchis: No, it's not required - from `avogadrolibs/CMakeLists.txt`
```
option(USE_LIBMSYM "Enable optional features using libmsym" ON)
```

ghutchis:
This is fine, but the call should only happen if USE_LIBMSYM is set.

serval2412:
Thank you ghutchis for your feedback, I changed avogadro/core/CMakeLists.txt as requested

serval2412:
Is there something I can do so the patch can be merged?
I noticed AppVeyor build failed but don't understand what's the relation with the patch.

cryos:
It may be related, it looks like the crucial line was clipped though:
`CUSTOMBUILD : CMake error : The following variables are used in this project, but they are set to NOTFOUND.`
I will try and take a look at this error, and confirm whether or not it is related.

cryos:
You should try to develop on named branches, I just pushed #252 which just cherry picks your final commit, and tested that. Looks good here - I am closing this PR, and merging the other. Thanks for your contribution.

Issue #242, opened 2018-03-27
Failed compilation with USE_LIBMSYM


berquist:
This is with the latest version of libmsym and GCC 7.3.1.

```
/media/Backup/AUR_owned/avogadrolibs-git/src/avogadrolibs/avogadro/qtplugins/symmetry/symmetry.cpp: In member function ‘void Avogadro::QtPlugins::Symmetry::detectSymmetry()’:
/media/Backup/AUR_owned/avogadrolibs-git/src/avogadrolibs/avogadro/qtplugins/symmetry/symmetry.cpp:263:56: error: invalid conversion from ‘msym::msym_symmetry_operation_t** {aka msym::_msym_symmetry_operation**}’ to ‘const msym_symmetry_operation_t** {aka const msym::_msym_symmetry_operation**}’ [-fpermissive]
       (ret = msymGetSymmetryOperations(m_ctx, &msopsl, &msops))) {
                                                        ^~~~~~
In file included from /media/Backup/AUR_owned/avogadrolibs-git/src/avogadrolibs/avogadro/qtplugins/symmetry/operationstablemodel.h:22:0,
                 from /media/Backup/AUR_owned/avogadrolibs-git/src/avogadrolibs/avogadro/qtplugins/symmetry/symmetrywidget.h:31,
                 from /media/Backup/AUR_owned/avogadrolibs-git/src/avogadrolibs/avogadro/qtplugins/symmetry/symmetry.h:22,
                 from /media/Backup/AUR_owned/avogadrolibs-git/src/avogadrolibs/avogadro/qtplugins/symmetry/symmetry.cpp:17:
/usr/include/libmsym/msym.h:181:30: note:   initializing argument 3 of ‘msym::msym_error_t msym::msymGetSymmetryOperations(msym::msym_context, int*, const msym_symmetry_operation_t**)’
     msym_error_t MSYM_EXPORT msymGetSymmetryOperations(msym_context ctx, int *sopsl, const msym_symmetry_operation_t **sops);
                              ^~~~~~~~~~~~~~~~~~~~~~~~~
/media/Backup/AUR_owned/avogadrolibs-git/src/avogadrolibs/avogadro/qtplugins/symmetry/symmetry.cpp:294:63: error: invalid conversion from ‘msym::msym_subgroup_t** {aka msym::_msym_subgroup**}’ to ‘const msym_subgroup_t** {aka const msym::_msym_subgroup**}’ [-fpermissive]
[...]

PR #247, opened 2018-04-01
Clean up symmetry code to use const-correct version of libmsym.

ghutchis:
Should solve #242 exposed by building "native" libmsym.


ghutchis:
I have a few related changes that I'll submit soon, including display of symmetry-equivalent atoms.

PR #248, opened 2018-04-01
Bump version number to 1.91 for next release.

ghutchis:
(Actually 1.90 is over a year old)

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -20,7 +20,7 @@ include(DetermineVersion)
 
 # Set up our version.
 set(AvogadroLibs_VERSION_MAJOR   "1")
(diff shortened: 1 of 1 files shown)


cryos:
I usually do this right before release, but now is fine. We get the version from the last tag, and the output of git-describe. It isn't used as much as it used to be.

ghutchis:
I'm drafting a message basically arguing that we're long past due for a release and that we should set up some timelines for getting 2.0 out the door.

cryos:
Sounds good, I am going to get to pushing an updated release out, and am in full agreement. The version number in CMake only really gets used in non-git release tarballs now, just for reference.

PR #251, opened 2018-04-01
Set up translation templates with adapted scripts from Avogadro v1

ghutchis:
Run in main directory `sh scripts/extract-messages.sh`
This should be set up with Ubuntu Launchpad to automatically
update - both uploading the template and the translations to Git

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -41,6 +41,7 @@ include_directories(BEFORE ${CMAKE_CURRENT_SOURCE_DIR}
   ${CMAKE_CURRENT_BINARY_DIR})
 
 option(ENABLE_TESTING "Enable testing and building the tests." OFF)
+option(ENABLE_TRANSLATIONS "Enable building translations with Qt5 Linguist" OFF)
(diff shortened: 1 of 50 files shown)


cryos:
Sounds good, my main concern is whether these would be better in a submodule, but I am not sure it is worth complicating things too much. I think these are a reasonable exception as they are actually text files, and having diffs would likely be useful for translators, and they are quite tightly coupled to the strings in the code. I haven't had time to test these, but am fine with this being merged if you think it would be useful. Thanks for working on the translation infrastructure.

ghutchis:
My feeling is that the PO files (as text) should be in the repository - this allows Launchpad integration so that they can be updated automatically as in Avogadro 1.x.

My plan is to set up a separate repository for the Qt-generated QM files that works with the download manager and can be integrated into the superbuild (e.g., when building packages, you'll get the latest QM files and you can download updates over time).

cryos:
> My feeling is that the PO files (as text) should be in the repository - this allows Launchpad integration so that they can be updated automatically as in Avogadro 1.x.

It sounds like we are in agreement Geoff.

PR #259, opened 2018-04-06
Fix name of CMake libmsym detection file

berquist:
This fixes the `make install` step.

Another option is to change the name of the CMake file. What is the CMake convention when the library name is all lowercase?


berquist:
Hm, beat to the punch.

berquist:
Hm, beat to the punch I see.

cryos:
Thanks anyway, sorry I missed this when I was fixing up the find logic for libmsym.

PR #263, opened 2018-04-10
Fix build with Qt-5.11

a17r:


avogadro/qtplugins/gamessinput/gamessinput.cpp
```diff
@@ -24,9 +24,9 @@
 
 #include <molequeue/client/jobobject.h>
 
-#include <QtCore/QDebug>
-
-#include <QtWidgets/QMessageBox>
+#include <QAction>
+#include <QDebug>
```
cryos: Does this require that we remove the Qt module, or just the addition of the QAction include?

a17r: No, the fix is just adding the missing QAction header. But with Qt5 the recommendation is to not use the Qt module prefix anymore, so I drop wherever I see them.

cryos: Fair enough, we use it elsewhere so I would prefer not to drop  it, I can append to your commit and merge that though.

a17r: I can force-push just the QAction change in a second if you want me to

cryos: That would be awesome if you don't mind

cryos:
Merged. Thanks.

PR #265, opened 2018-04-17
Fix build with Qt 5.11_beta3 (dropping qt5_use_modules)

a17r:


tests/qtopengl/CMakeLists.txt
```diff
@@ -44,13 +44,12 @@ create_test_sourcelist(testDriver qtopengltests.cpp ${testSrcs})
...
 
-qt5_use_modules(AvogadroQtOpenGLTests OpenGL)
-
 target_link_libraries(AvogadroQtOpenGLTests
   AvogadroQtOpenGL
   vtkImagingCore
   vtkIOImage
-  vtkRenderingQt)
+  vtkRenderingQt
+  Qt5::OpenGL)
```
cryos: Are you sure this is required? I factored out the need for the Qt5 OpenGL module as it is deprecated and moved to the new QOpenGL* classes.

a17r: I can try without, this is just the result of merging the qt5_use_modules in here.

PR #266, opened 2018-04-23
Added a more generic Vtk plot function

psavery:
Moved the Vtk plot function from 'avogadro/qtplugins/plotxrd'
to 'avogadro/vtk' and made it a little more generic.

The function is now able to plot more than one line at a time
(although some improvements can still be made such as the addition
of a legend).

This has been done in preparation for generating band structure
plots, which require many lines.


psavery:
Hi Marcus! Feel free to let me know if I should change any of my design decisions here.

A few decisions to note:

1. I added the vtk `include_directories` statement [here](https://github.com/OpenChemistry/avogadrolibs/pull/266/files#diff-3c6a377c1f136182c350ff82841255dcR55) in the CMakeLists.txt file because it is now needed by some of the QtPlugins (in particular, `plotxrd`). I couldn't move the vtk `add_subdirectory` located [here](https://github.com/OpenChemistry/avogadrolibs/pull/266/files#diff-3c6a377c1f136182c350ff82841255dcR61) to any higher because it requires some of the Qt stuff before it. I hope it isn't a problem that the vtk `include_directories` and vtk `add_subdirectory` statements are not adjacent.

2. I removed the components from VTK for `find_package` [here](https://github.com/OpenChemistry/avogadrolibs/pull/266/files#diff-ea60a17665b0a47869fd996219598d72L15) just because it is easier to type, and I included the VTK_USE_FILE. Let me know if this is an issue, and I can change it back and add the specific components for the charts. Likewise, this change carried over to the `target_link_libraries` [here](https://github.com/OpenChemistry/avogadrolibs/pull/266/files#diff-ea60a17665b0a47869fd996219598d72L35).

I tested plotting multiple lines, and it seems to work. You can see a result pasted below (I added a dummy green line for this image just to test it out - it is not part of this PR).
avogadro2-vtk-plot

It seems that the Xrd plotting works fine with this change.

ghutchis:
FYI - there are some great band plots in sumo:
http://sumo.readthedocs.io/en/latest/gallery.html

avogadro/vtk/CMakeLists.txt
```diff
@@ -11,26 +11,24 @@ endif()
 
 find_package(Qt5 COMPONENTS Widgets REQUIRED)
 
-find_package(VTK
-  COMPONENTS
-    vtkRenderingOpenGL2 vtkGUISupportQt vtkDomainsChemistry
-    vtkRenderingVolumeOpenGL2 vtkViewsCore vtkRenderingFreeType
-  REQUIRED)
+find_package(VTK REQUIRED)
```
cryos: We should stick to requesting the relevant modules, not just all of VTK - this will lead to overlinking, and making it less clear when modules we need are not built.

avogadro/vtk/CMakeLists.txt
```diff
@@ -11,26 +11,24 @@ endif()
...
+include(${VTK_USE_FILE})
```
cryos: This does stuff I would like to avoid too

psavery:
I fixed it by requiring cmake to look for the specific VTK modules (the new ones are: `vtkChartsCore`, `vtkViewsContext2D`, and `vtkRenderingContextOpenGL2`). And I removed the VTK_USE_FILE usage.

Let me know if any other changes are needed.

avogadro/CMakeLists.txt
```diff
@@ -50,12 +50,17 @@ if(USE_QT)
     add_subdirectory(molequeue)
     include_directories(${CMAKE_CURRENT_BINARY_DIR}/molequeue)
   endif()
+
+  if(USE_VTK)
+    include_directories(${CMAKE_CURRENT_BINARY_DIR}/vtk)
```
cryos: We should move the add_subdirectory(vtk) up to this block too, and get rid of the block below. The VTK we are building requires Qt too, and so that should be represented in the logic. Other than that looks good.

psavery:
Just pushed an update. Hopefully, it will be good!

PR #267, opened 2018-04-24
In tests/qtopengl, link to Qt5::OpenGL

psavery:
This fixes a linking error that occurs if you use the
cmake definitions '-DUSE_VTK=ON' and '-DENABLE_TESTING=ON'


cryos:
@psavery Are you sure it needs OpenGL? I can try building it in a few, but I did quite a bit of work to remove the QGL deprecated classes, and so I am surprised it doesn't just need widgets adding - we should only be using QOpenGL classes that are part of Gui/Widgets.

cryos:
Closing in favor of #271

PR #268, opened 2018-04-24
Print the clang-format config info on failure

psavery:
Some versions of clang-format use different
settings (even with the same "BasedOnStyle" flag).

Printing the clang-format configuration upon failure
can speed up the process of fixing clang-format issues.

### DIFF
--- a/scripts/travis/run_clang_format_diff.sh
+++ b/scripts/travis/run_clang_format_diff.sh
@@ -4,5 +4,6 @@ if [ -z "$DIFF" ]; then
   exit 0
 else
   printf "ERROR: clang-format-diff detected formatting issues. Please run clang-format on your branch.\nThe following formatting changes are suggested:\n\n%s" "$DIFF"
+  printf "\n\nThe following is the clang-format configuration:\n%s\n" "`clang-format-3.8 -dump-config`"
   exit 1
 fi


psavery:
Note some of the differences in the clang-format-3.8 configuration and the clang-format-6.0 configuration (both using the .clang-format file [here](https://github.com/OpenChemistry/avogadrolibs/blob/master/.clang-format)).

```
$ diff clang-format-3.8-config.out clang-format-6.0-config.out 
7c7
< AlignEscapedNewlinesLeft: false
---
> AlignEscapedNewlines: Right
20,22c20,22
< BinPackArguments: true
< BinPackParameters: true
< BraceWrapping:
---
> BinPackArguments: false
> BinPackParameters: false
> BraceWrapping:   
30a31
>   AfterExternBlock: true
33a35,37
>   SplitEmptyFunction: true
>   SplitEmptyRecord: false
>   SplitEmptyNamespace: true
35a40
> BreakBeforeInheritanceComma: true
37a43,45
> BreakConstructorInitializers: BeforeComma
> BreakAfterJavaFieldAnnotations: false
> BreakStringLiterals: true
39a48
> CompactNamespaces: false
47,48c56,62
< ForEachMacros:   [ foreach, Q_FOREACH, BOOST_FOREACH ]
< IncludeCategories:
---
> FixNamespaceComments: false
> ForEachMacros:   
>   - foreach
>   - Q_FOREACH
>   - BOOST_FOREACH
> IncludeBlocks:   Preserve
> IncludeCategories: 
51c65
<   - Regex:           '^(<|"(gtest|isl|json)/)'
---
>   - Regex:           '^(<|"(gtest|gmock|isl|json)/)'
54a69
> IncludeIsMainRegex: '(Test)?$'
55a71
> IndentPPDirectives: None
57a74,75
> JavaScriptQuotes: Leave
> JavaScriptWrapImports: true
65a84
> PenaltyBreakAssignment: 2
72a92,95
> RawStringFormats: 
>   - Delimiter:       pb
>     Language:        TextProto
>     BasedOnStyle:    google
74a98
> SortUsingDeclarations: true
75a100
> SpaceAfterTemplateKeyword: false
88a114
> 
```

cryos:
They never guaranteed the strict compatibility between versions, hence printing the before and after diff if there are issues so you can apply them locally. I would rather make our .clang-format more explicit and/or upgrade the clang-format used on Travis than print this out. It obscures the diff by adding more output in the error, and doesn't really tell you what to do with it.

I use Arch, and can test with whatever version I have locally, but as I see it the better options are to simply paste this output once into the wiki, or a page documenting the formatting (which we have), or to update .clang-format and/or the travis CI's clang-format so running it locally produces identical output.

psavery:
Hi Marcus!

That makes sense! I saw that we can upgrade clang-format to 6.0 if we use `trusty` instead of `precise` for the distro on travis-ci.

One of the big changes I was facing was the changes in the defaults for `BinPackArguments` and `BinPackParameters`. But I think there have also been a few other changes as well including the way it arranges constructor initialization lists (via the new option `BreakConstructorInitializers`).

I submitted PR #269 to succeed this one. Let me know what you think of it.

PR #271, opened 2018-05-06
Port the GL initialization code to the new API

cryos:
The QGL* classes won't help much now that the widgets use the new QOpenGL*
based classes, there are still some issues with these tests that need to
be looked at but they build/link successfully.


cryos:
This supersedes #267, converting over to the QOpenGL* classes.

PR #273, opened 2018-05-26
GROMACS trr trajectory support

badarsh2:
This PR aims at adding support for reading GROMACS trr trajectory files.

### DIFF
--- a/avogadro/io/trrformat.cpp
+++ b/avogadro/io/trrformat.cpp
@@ -1,0 +1,445 @@
+/******************************************************************************
+
+  This source file is part of the Avogadro project.
+
+  Copyright 2018 Kitware, Inc.
+
+  This source code is released under the New BSD License, (the "License").
+
+  Unless required by applicable law or agreed to in writing, software
+  distributed under the License is distributed on an "AS IS" BASIS,
+  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+  See the License for the specific language governing permissions and
+  limitations under the License.
+
+******************************************************************************/
+
+#include "trrformat.h"
+#include "struct.h"
+
+#include <avogadro/core/elements.h>
+#include <avogadro/core/molecule.h>
+#include <avogadro/core/unitcell.h>
+#include <avogadro/core/utilities.h>
+#include <avogadro/core/vector.h>
+
+#include <iomanip>
+#include <istream>
+#include <ostream>
+#include <sstream>
+#include <string>
+
+using std::endl;
+using std::getline;
+using std::map;
+using std::pair;
+using std::string;
+using std::to_string;
+using std::vector;
+
+namespace Avogadro {
+namespace Io {
+
+using Core::Array;
+using Core::Atom;
+using Core::Elements;
+using Core::lexicalCast;
+using Core::Molecule;
+using Core::split;
+using Core::trimmed;
+using Core::UnitCell;
+
+#ifndef _WIN32
+using std::isalpha;
+#endif
+
+#define GROMACS_MAGIC 1993
+#define DIM 3
+#define NM_TO_ANGSTROM 10.0
+string TRRVERSION = "GMX_trn_file";
+string HEADITEMS[] = { "ir_size",   "e_size",   "box_size", "vir_size",
+                       "pres_size", "top_size", "sym_size", "x_size",
+                       "v_size",    "f_size",   "natoms",   "step",
+                       "nre",       "time",     "lambda" };
+
+int swapInteger(int inp)
+{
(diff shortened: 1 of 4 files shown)


badarsh2:
The trr file has the trajectory information in portable binary format. For unpacking the binary data, the trrformat class relies on two other headers struct.h and struct_endian.h which I've sourced from [here](https://github.com/svperbeast/struct). As of now, I've placed them in the avogadro/io folder, but these can be relocated to a more sensible location. I would like your suggestion for the same.

cryos:
This is useful as it shows a compilation issue on Windows in the third party code, which would also need to be addressed before we can merge this topic. I wonder if the library has been looking at the Windows MSVC toolchain at all, you can see the error towards the end of the appveyor build.

cryos:
It may actually just be types in the code in trrformat taking a closer look, line 282 where the != may need to have the type more correctly cast to match one of the overloads. There is a lot of very C calls being made too, we may want to look at if they are the best option, but as they are binary files it may be preferred. The Windows support for C99 is not there, but on pretty much every other platform.

avogadro/io/trrformat.cpp
```diff
@@ -0,0 +1,446 @@
...
+#define GROMACS_MAGIC 1993
+#define DIM 3
+#define NM_TO_ANGSTROM 10.0
+string TRR_VERSION = "GMX_trn_file";
+string HEAD_ITEMS[] = { "ir_size",   "e_size",   "box_size", "vir_size",
+                        "pres_size", "top_size", "sym_size", "x_size",
+                        "v_size",    "f_size",   "natoms",   "step",
+                        "nre",       "time",     "lambda" };
+
+int swap_integer(int inp)
```
cryos: Should stick to the same naming convention, swapInteger, rather than STL style swap_integer

avogadro/io/trrformat.cpp
```diff
@@ -0,0 +1,446 @@
...
+{
+  return (((inp << 24) & 0xff000000) | ((inp << 8) & 0x00ff0000) |
+          ((inp >> 8) & 0x0000ff00) | ((inp >> 24) & 0x000000ff));
+}
+
+char swap_endian(char endian)
```
cryos: As above

avogadro/io/trrformat.cpp
```diff
@@ -0,0 +1,446 @@
...
+{
+  if (endian == '>')
+    return '<';
+  else
+    return '>';
+}
+
+/* Checks whether the data stored in the binary file is of float or double type
+ */
+int is_double(map<string, int>& header)
```
cryos: As above

avogadro/io/trrformat.cpp
```diff
@@ -0,0 +1,446 @@
...
+  bool double_status;
+  char endian = '>', buff[BUFSIZ], fmt[BUFSIZ], raw[1000];
+  int magic, natoms, slen0, slen1, headval[13];
+  string subs, key_check[] = { "box_size", "vir_size", "pres_size" },
+               key_check_2[] = { "x_size", "v_size", "f_size" };
+  map<string, int> header;
+
+  // Determining size of file
+  inStream.seekg(0, inStream.end);
+  int file_len = inStream.tellg();
```
cryos: Try to stick with camelCase naming convention

avogadro/io/trrformat.cpp
```diff
@@ -0,0 +1,446 @@
...
+      }
+      mol.setCustomElementMap(elementMap);
+    }
+  }
+  mol.setCoordinate3d(mol.atomPositions3d(), 0);
+
+  // Do we have an animation?
+  // EOF check
+  int coordSet = 1;
+  while (inStream.tellg() != file_len) {
```
cryos: This seems to be where the compiler error on Windows is coming from - from memory I didn't think file lengths were expressed in int, but it looks like an ambiguous overload.

cryos:
We need to edit away the addition/removal of struct from the branch. You can squash all the commits together for the simplest approach. We are not counting commits, but useful change, this churn just adds to the repository history size.

avogadro/io/trrformat.cpp
```diff
@@ -277,8 +277,7 @@ bool TrrFormat::read(std::istream& inStream, Core::Molecule& mol)
   // Do we have an animation?
   // EOF check
   int coordSet = 1;
-  int currPos = inStream.tellg();
-  while (currPos != fileLen) {
+  while ((int)inStream.tellg() != fileLen) {
```
cryos: Prefer C++ static_cast over C type casting

badarsh2:
I've made the change and amended the commit. The travis build seems to be failing, and it looks like there's an issue with spglib download in the gcc-6.3 build job.

PR #277, opened 2018-06-08
Addition of timestep field in molecule class

badarsh2:
This PR adds facility to store timestep values from molecular dynamics trajectories in the molecule class.

### DIFF
--- a/avogadro/core/molecule.h
+++ b/avogadro/core/molecule.h
@@ -474,9 +474,9 @@ public:
   void clearCubes();
 
   /**
-* @brief Get the cubes vector set (if present) for the molecule.
-* @return The cube vector for the molecule
-*/
+   * @brief Get the cubes vector set (if present) for the molecule.
+   * @return The cube vector for the molecule
+   */
   std::vector<Cube*> cubes() { return m_cubes; }
   const std::vector<Cube*> cubes() const { return m_cubes; }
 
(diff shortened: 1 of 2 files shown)


avogadro/core/molecule.h
```diff
@@ -533,6 +533,12 @@ class AVOGADROCORE_EXPORT Molecule
   int coordinate3d() const;
   bool setCoordinate3d(const Array<Vector3>& coords, int index);
 
+  /**
+   * Timestep property is used when molecular dynamics trajectories are read
+   */
+  bool setTimeStep(double timestep, int index);
+  bool getTimeStep(double& timestep, int index);
```
cryos: We follow Qt naming convention, setTimeStep, but timeStep for the getter. I would also prefer to follow Qt's convention on the getter - return a double as one would expect, but have an optional bool passed in as the final argument, by reference, that can indicate if the time step was valid.

ghutchis:
I'd really like to stop adding fields to the Molecule class when they won't always be used. Can we set up some level of generic key/value?

In this case, I'm thinking it would be good to have a data structure that reflects the multiple coordinate sets. Beyond time steps, there are multiple properties that we want to store for each coordinate:
* energy
* pressure
* radial distribution function
* etc.

So I'd rather push for time steps to be part of a separate class/struct called CoordSetProperties or something similar.

cryos:
I don't think it is too bad, as when not used it is just an array that never gets space allocated. We can certainly explore other approaches, the coordinate sets are already part of molecule, and just not used when there is only one geometry. I am open to creating a fuller object. I am doing some work on another project where we will load in a set of coordinates, updated electronic structure, etc per frame. It may ultimately be better to have some support for metadata about the sequence, but have the molecule simply reflect the current frame, loading previous/next frames on demand. That is a bigger set of changes, and we may not want to attach too much important to this one addition.

PR #280, opened 2018-06-13
Completing coordinate3D() function in molecule class

badarsh2:
I've completed the function which has been incomplete for quite some time, so that it can be used for conducting checks in each element of the coordinate sets in the tests for trajectory readers.


cryos:
This will need rebasing on master.

PR #281, opened 2018-06-14
(WIP) Implementation of custom element assignment from file

badarsh2:
This is one of the ideas that I had proposed earlier to account for the non-availability of element information in many of the trajectory formats. Particularly in situations where the number of atoms in the system is huge, it becomes impractical to use the CustomElementDialog tool. Hence, in addition to the custom element mapping tool in the build menu, I've included an additional option to import the element information from file. This creates a temporary molecule class instance for the file imported, and the element information is copied from this molecule to the molecule in question. For example, if a GROMACS trr trajectory has been read, one can use this option to import a .gro file (just like normal file import, which is done using Avogadro's readers or OpenBabel if available) and the details of the elements will be reflected in the trajectory molecule. (This does not explicitly create a CustomElementMap, I have just used the setAtomicNumber() function to achieve the aforementioned).


badarsh2:
image

avogadro/qtplugins/customelements/customelements.cpp
```diff
@@ -46,12 +59,12 @@ QString CustomElements::description() const
...
 QList<QAction*> CustomElements::actions() const
 {
-  return QList<QAction*>() << m_reassignAction;
+  return QList<QAction*>() << m_reassignUsingTool << m_reassignFromFile;
 }
 
 QStringList CustomElements::menuPath(QAction*) const
 {
-  return QStringList() << tr("&Build");
+  return QStringList() << tr("&Build") << tr("Assign &Custom Elements...");
```
ghutchis: I'm unsure on naming here. Generally I hear "topology" more - I'm not sure if one is more prevalent. What do you think?

avogadro/qtplugins/customelements/customelements.cpp
```diff
@@ -86,9 +99,123 @@ void CustomElements::reassign()
...
+
+  // Setup a progress dialog in case file loading is slow
+  m_progressDialog = new QProgressDialog(qobject_cast<QWidget*>(parent()));
+  m_progressDialog->setRange(0, 0);
+  m_progressDialog->setValue(0);
+  m_progressDialog->setMinimumDuration(750);
+  m_progressDialog->setWindowTitle(tr("Reading File"));
+  m_progressDialog->setLabelText(
+    tr("Opening file '%1'\nwith '%2'").arg(fileName).arg(ident));
+  /// @todo Add API to abort file ops
```
ghutchis: Wouldn't you just connect the cancel signal with m_fileReadThread => quit() instead?

avogadro/qtplugins/customelements/customelements.cpp
```diff
@@ -86,9 +99,123 @@ void CustomElements::reassign()
...
+void CustomElements::setMapFromMolecule(QtGui::Molecule* mol)
+{
+  if (mol->atomCount() != m_molecule->atomCount()) {
+    QMessageBox::critical(
+      qobject_cast<QWidget*>(parent()), tr("Error"),
+      tr("Atom count mismatch.\nExpected %1 atoms, found %2.")
+        .arg(m_molecule->atomCount())
+        .arg(mol->atomCount()));
+  } else {
+    size_t n = m_molecule->atomCount();
```
ghutchis: If you get a file, do you want to map bonds as well? For the molecular-based trajectories (even XYZ) this seems like a good idea.

badarsh2:
Made the above suggested changes.

badarsh2:
I moved the background file format script and headers to a new library folder and tried using these headers in the app, but mainwindow.cpp doesn't seem to recognise the header path, so I guess I'm missing out on something. Anything I should ensure while moving the files to a separate plugin and using them in the app?

cryos:
They should be moved to qtgui, and would need the export macros adding as other classes do so that the symbols are exported. They would also need putting into the same namespace as the other classes, they don't need to be in an entirely new library.

badarsh2:
@cryos I've made the suggested changes, and the plugin now uses the relocated backgroundfileformat class.

ghutchis:
@cryos - can you re-review this so we can merge if it's ready?

ghutchis:
Ping @cryos

PR #287, opened 2018-06-26
(Experimental) Support for AMBER mdcrd trajectories

badarsh2:
This PR aims at providing experimental support in Avogadro for reading AMBER mdcrd trajectory files.
The reason for mentioning it as experimental is because there is no tractable solution for obtaining the number of atoms from the trajectory directly. It has to be inputted by the user, for which I've tried using a QInputDialog, which poses a problem because there is no directly inheritable parent widget, so I'm calling QInputDialog with a null parent, which is dangerous. Running it in a separate thread is highly advised against, and I myself have experienced a couple of segfault crashes while running the application.
The cmake tree also seemed to be a bit complicated, that trying to create a plugin or a standalone gui class seemed impossible due to cyclic dependencies.


cryos:
I am not sure why you think there are cyclic dependencies, you should outline where you think these are. A plugin can certainly depend upon the AvogadroCore, AvogadroIo libraries, extend the plugin interface and use Qt code. If you provide an example of this complication I would be happy to take a look, but will not add Qt dependencies to io.

ghutchis:
I'd also suggest that it'd make sense that the `io` class takes a parameter for the number of atoms - or ideally the topology Molecule.

Then the qtplugin can use a QInputDialog or other mechanism to feed the number of atoms to the format.

This ensures proper separation of modularity without forcing the class into another directory (i.e., it makes logical sense to exist in `io`).

badarsh2:
Pardon me on the cyclic dependency thing I mentioned earlier. I was trying to add `AvogadroQtGui` to the dependency list in `io` so that I can directly call a custom dialog class, but I learnt later that's not how it's meant to work :/

I've come up with a different approach in the latest [commit](https://github.com/OpenChemistry/avogadrolibs/pull/287/commits/34a6af09742f1fa048d3a79a36032171e8fe8689). Since the mdcrd datatype is merely a huge sequence of floats, I'm reading them an storing them in the molecule class as a coordinate set for now. Then I invoke the custom gui dialog [here](https://github.com/badarsh2/avogadroapp/commit/16b2031f06858c02f4a687965613aa0d61a7fa04) which gets the numAtoms from the user. Appropriately, the atoms are then added, and emitChanged() is called to update the scene. The scene isn't getting appropriately centered though after the atoms are assigned (probably I'm missing out on emitting or calling some other function/signal). 

This approach is still a bit rickety and feels kind of unauthentic to me (due to the lack of precision in the ASCII mdcrd format, which is also one of the reasons why people in the AMBER community prefer using the netcdf trajectory format, but netcdf libraries aren't friendly to deal with).

PR #288, opened 2018-06-29
RMSD VTK chart plugin for multimolecules

badarsh2:
This PR adds support for displaying the Root Mean Square Deviation curve for multimolecules. This will be useful for analysing MD trajectories.
image

### DIFF
--- a/avogadro/qtplugins/CMakeLists.txt
+++ b/avogadro/qtplugins/CMakeLists.txt
@@ -121,6 +121,7 @@ add_subdirectory(vrml)
 add_subdirectory(workflows)
 
 if(USE_VTK)
+  add_subdirectory(plotrmsd)
   add_subdirectory(plotxrd)
 endif()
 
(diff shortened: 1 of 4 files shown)


ghutchis:
This is great - also useful for conformer generation, where people want to compare RMSD. Thanks!

badarsh2:
Great! I was just wondering about where to position the widget in the menu. I've positioned it in the crystal menu for now, though technically RMSD is applicable to any general system and not just crystalline systems. If you have suggestions for placing it in a more appropriate menu, please let me know.

ghutchis:
I think there needs to be a "Properties" submenu of View, much like Avogadro v1.

Then "Molecular Properties" and "Symmetry Properties" and RMSD can all live there.

I haven't had a chance to resurrect the Atom, Bond, Angle, and Torsion tables, but that's on the table for the next few weeks too. (All in the View ⇒ Properties path)

badarsh2:
Shall I add an issue for relocation of the RMSD menu option for now, so that it can be addressed later as well?

ghutchis:
Sure. +1

PR #296, opened 2018-07-05
Pair Distribution Function VTK chart 

badarsh2:
This PR adds feature for plotting the Pair (Radial) distribution function. 

**How it works:**
As of now, only molecules having a box (unit cell) are allowed this feature. The maximum radius (say `maxRadius`) till which the PDF is to be computed and the step value are obtained as user input. A duplicate crystal is created by expanding the original crystal in such a way that spheres of radius `maxRadius` centered at each atom of the central crystal are completely filled without voids. Since the original crystal coordinates are still pivoted at the actual origin, they're displaced to the centre of the expanded crystal. Pairs of atoms are iteratively chosen with the centre atom from the centered crystal and the dangling atom from the expanded crystal, and the distances are calculated. Bins with spacing of `step` are created and the pairwise distances are binned. Subsequently, the PDF is calculated by taking average of the bin counts and dividing by the corresponding elemental volume.

**Sample Plot for crystalline alpha-Ti:**
image

**Sample Plot for Si melt:**
image

The distinction between crystalline and amorphous / liquid phases is clearly derivable from the nature of the PDF plots.

### DIFF
--- a/avogadro/qtplugins/plotpdf/pdfoptionsdialog.h
+++ b/avogadro/qtplugins/plotpdf/pdfoptionsdialog.h
@@ -1,0 +1,52 @@
+/******************************************************************************
+
+  This source file is part of the Avogadro project.
+
+  Copyright 2018 Kitware, Inc.
+
+  This source code is released under the New BSD License, (the "License").
+
+  Unless required by applicable law or agreed to in writing, software
+  distributed under the License is distributed on an "AS IS" BASIS,
+  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+  See the License for the specific language governing permissions and
+  limitations under the License.
+
+******************************************************************************/
+
+#ifndef AVOGADRO_QTPLUGINS_PDFOPTIONSDIALOG_H
+#define AVOGADRO_QTPLUGINS_PDFOPTIONSDIALOG_H
+
+#include <QDialog>
+
+namespace Avogadro {
+namespace QtPlugins {
+
+namespace Ui {
+class PdfOptionsDialog;
+}
+
+/**
+ * @brief Dialog to set options for PDF curve plotting.
+ */
+class PdfOptionsDialog : public QDialog
+{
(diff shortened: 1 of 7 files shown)


avogadro/qtplugins/plotpdf/plotpdf.h
```diff
@@ -0,0 +1,86 @@
...
+  limitations under the License.
+
+*******************************************************************************/
+
+#ifndef AVOGADRO_QTPLUGINS_PLOTPDF_H
+#define AVOGADRO_QTPLUGINS_PLOTPDF_H
+
+#include <avogadro/qtgui/extensionplugin.h>
+
+#include <memory>
```
cryos: Why is this needed in the header?

avogadro/qtplugins/plotpdf/pdfoptionsdialog.h
```diff
@@ -0,0 +1,54 @@
...
+  ~PdfOptionsDialog();
+
+  double maxRadius() const;
+  double step() const;
+
+protected slots:
+  void accept();
+
+private:
+  std::unique_ptr<Ui::PdfOptionsDialog> m_ui;
```
cryos: Should prefer the Qt pointers over STL in Qt classes - QScopedPointer

avogadro/qtplugins/plotpdf/plotpdf.h
```diff
@@ -0,0 +1,86 @@
...
+  // err will be set to an error string if the function fails.
+  // radius is in Angstroms.
+  static bool generatePdfPattern(QtGui::Molecule& mol, PdfData& results,
+                                 QString& err, double maxRadius = 10.0,
+                                 double step = 0.1);
+
+  QList<QAction*> m_actions;
+  QtGui::Molecule* m_molecule;
+
+  std::unique_ptr<PdfOptionsDialog> m_pdfOptionsDialog;
```
cryos: I think QScopedPointer should be used with Qt classes

PR #297, opened 2018-07-05
WIP: Molecular surfaces

ghutchis:
This is work in progress to restructure the surfaces code to be more general. Eventual goals include:
- Support for molecular surfaces like VdW, Solvent Accessible, and Solvent Excluded
  - This will use code ported from EDTSurf https://zhanglab.ccmb.med.umich.edu/EDTSurf/
- Improved separation of QM surface generation (e.g., to enable alpha/beta orbitals)
- Allowing surfaces to be colored by any other cube (e.g., color VdW by HOMO or LUMO)

screen shot 2018-07-05 at 1 40 20 pm

I'm trying to figure out how to break into manageable pull requests. I might merge soon, then work on step 1 with Sean Donahue, who is cleaning up EDTSurf and porting to Eigen.


cryos:
I am glad you pushed this, I have been working on the basis set code a little, and looking to generalize it/add support for alpha/beta. I have it all loaded in my head right now, and would like to get this tested soon. I can test this, and would like to get to a point where it can be merged so that we can get alpha/beta exposed in the UI.

ghutchis:
I remembered you mentioned work on the basis set side, so we'll avoid work there for now.

We're going to focus on efficient molecular surfaces and mark places in the code for alpha/beta or unrestricted issues.

At the moment, there doesn't seem to be any existing code for computing color-mapped properties, correct?

cryos:
There is no code for coloring meshes, and I think the rendering code will need updates. I am working on alpha/beta orbitals, and also being able to move through a set of related calculations with an animation type GUI (i.e. not change camera, update geometry, electronic structure). I can likely work in a custom GUI, and then look at how/if that might be integrated.

The alpha/beta stuff is pretty simple, although another side I want to improve is reusing cubes if we just want an updated isovalue. Now I know you are moving all of this code around it would be good to merge some of it, but it isn't the end of the world if it lives in a separate plugin for a little while. I should also be getting back to VTK volume rendering very soon, the new QVTKOpenGLWidget code was merged recently.

Issue #298, opened 2018-07-05
Continue modernizing code


ghutchis:
There are some additional static-analysis cleanups that could be done to minimize warnings and improve the code. Some of these can be run easily from Qt Creator.

- [x] Cppcheck: https://github.com/danmar/cppcheck
- [x] Clang static analyzers / clang-tidy: http://clang.llvm.org/extra/clang-tidy/
- [x] Clazy (clang KDE/Qt checks): https://github.com/KDE/clazy
- [x] Clang sanitizers: https://clang.llvm.org/docs/AddressSanitizer.html
- [ ] Include-what-you-use: https://github.com/include-what-you-use/include-what-you-use

For example, many classes do not use the `override` keyword in C++11:
https://en.cppreference.com/w/cpp/language/override

The older Krazy2 project might also have some useful checkers:
- [ ] https://github.com/Krazy-collection/krazy

ghutchis:
For right now, I'm using the following clazy checks:
```
export CLAZY_CHECKS="level0,level1,reserve-candidates,container-inside-loop,tr-non-literal,unneeded-cast,isempty-vs-count,static-pmf,qstring-allocations,old-style-connect,base-class-event,ctor-missing-parent-argument,missing-qobject-macro"
```

ghutchis:
There are a *lot* of minor bits turned up with `clazy` many of which will generally improve the code.

For example, using Qt5, I'd suggest we move to the new signal/slot connection using pointers:
https://github.com/KDE/clazy/blob/master/docs/checks/README-old-style-connect.md
https://woboq.com/blog/new-signals-slots-syntax-in-qt5.html

ghutchis:
For suggestions on running `clang-tidy` there's a KDAB article:
https://www.kdab.com/clang-tidy-part-1-modernize-source-code-using-c11c14/

ghutchis:
(Whoops, didn't mean to close this.)

ghutchis:
Okay, I'm going to close this one - a lot of cleanups were fixed by #1249

PR #299, opened 2018-07-05
Use clazy fixit for string literals to reduce memory allocs

ghutchis:
https://github.com/KDE/clazy/blob/master/docs/checks/README-qstring-allocations.md

Part of #298

Change-Id: I412195b0d9fc60d26f2116e78b0bbc310fcd197f


ghutchis:
The changes look pretty innocuous to me - certainly makes sense to minimize string temporaries.

ghutchis:
Yeah, I'm not thrilled either - I'm surprised Qt doesn't have a macro to hide some of that.

PR #300, opened 2018-07-05
Use clang-tidy modernize-use-override

ghutchis:
Minimizes warnings from compilation with C++11 compilers.

Change-Id: If051516da3a1760b70c79e21be795ea5f7f3328b


ghutchis:
This will probably require clang-format pass too..

PR #306, opened 2018-07-13
WIP: Biological Residue Class

badarsh2:
Keeps track of residue name, residue id, chain id, atom names and physically stores the residue atoms: might be useful later to implement something like selecting atoms in a residue.
Bond assignment is done by using preset data from `residuedata` header. As of now, ALA and LEU have been added.

Signed-off-by: Adarsh Balasubramanian <badarsh2@gmail.com>

Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

### DIFF
--- a/avogadro/core/residue.cpp
+++ b/avogadro/core/residue.cpp
@@ -1,0 +1,129 @@
+/******************************************************************************
+
+  This source file is part of the Avogadro project.
+
+  Copyright 2018 Kitware, Inc.
+
+  This source code is released under the New BSD License, (the "License").
+
+  Unless required by applicable law or agreed to in writing, software
+  distributed under the License is distributed on an "AS IS" BASIS,
+  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+  See the License for the specific language governing permissions and
+  limitations under the License.
+
+******************************************************************************/
+
+#include "residue.h"
+#include "molecule.h"
+#include "residuedata.h"
+
+#include <algorithm>
+#include <cassert>
+
+namespace Avogadro {
+namespace Core {
+
+Residue::Residue() {}
+
+Residue::Residue(std::string& name)
+  : m_residueName(name)
+{}
+
+Residue::Residue(std::string& name, Index& number)
+  : m_residueName(name)
+  , m_residueId(number)
+{}
+
+Residue::Residue(std::string& name, Index& number, char& id)
+  : m_residueName(name)
+  , m_residueId(number)
+  , m_chainId(id)
+{}
+
+Residue::Residue(const Residue& other)
+  : m_residueName(other.m_residueName)
+  , m_residueId(other.m_residueId)
+  , m_atomNameMap(other.m_atomNameMap)
+{}
+
+Residue& Residue::operator=(Residue other)
+{
+  using std::swap;
+  swap(*this, other);
+  return *this;
+}
+
+Residue::~Residue() {}
+
+std::string Residue::residueName()
+{
+  return m_residueName;
+}
+
+void Residue::setResidueName(std::string& name)
+{
+  m_residueName = name;
+}
+
+Index Residue::residueId()
+{
+  return m_residueId;
+}
+
+void Residue::setResidueId(Index& number)
+{
+  m_residueId = number;
+}
+
+char Residue::chainId()
+{
+  return m_chainId;
+}
+
+void Residue::setChainId(char& id)
+{
+  m_chainId = id;
+}
+
+void Residue::addResidueAtom(std::string& name, Atom& atom)
+{
+  m_atomNameMap.insert(std::pair<std::string, Atom>(name, atom));
+}
+
+std::vector<Atom> Residue::residueAtoms()
+{
+  std::vector<Atom> res;
+  for (AtomNameMap::iterator it = m_atomNameMap.begin();
+       it != m_atomNameMap.end(); ++it) {
+    res.push_back(it->second);
+  }
+  return res;
+}
+
+void Residue::resolveResidueBonds(Molecule& mol)
+{
+  std::vector<std::pair<std::string, std::string>> bondSeq;
+  if (residueDict.find(m_residueName) != residueDict.end()) {
+    size_t i = 0;
+    bondSeq = residueDict[m_residueName].residueSingleBonds();
+    for (i = 0; i < bondSeq.size(); ++i) {
+      if (m_atomNameMap.find(bondSeq[i].first) != m_atomNameMap.end() &&
+          m_atomNameMap.find(bondSeq[i].second) != m_atomNameMap.end()) {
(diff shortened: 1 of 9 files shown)


avogadro/core/residuedata.h
```diff
@@ -0,0 +1,110 @@
...
+    { "CA", "C" },
+    { "N", "H" },
+    { "N", "HN" },
+    { "HA", "CA" },
+    { "HB1", "CB" },
+    { "HB2", "CB" },
+    { "HB3", "CB" } 
+  },
+  // Double bonds
+  { { "C", "O" } });
```
ghutchis: I'm a bit torn - at least for proteins, this will always be a double bond, so it could be resolved as a special case in code. It's a minor point, though.

avogadro/core/residuedata.h
```diff
@@ -0,0 +1,110 @@
...
+    { "HD12", "CD1" },
+    { "HD13", "CD1" },
+    { "HD21", "CD2" },
+    { "HD22", "CD2" },
+    { "HD23", "CD2" },
+    { "HG", "CG" },
+  },
+  // Double bonds
+  { { "C", "O" } });
+
```
ghutchis: What I'd suggest is to write a short Python script or keyboard macro that processes the resdata.txt from Open Babel into this form and 'just do it'.

ghutchis: It's also probably worth saving the script, since I think there's a larger ligand database now that could eventually be processed. (OB includes HOH, HEM, etc.)

avogadro/core/residue.h
```diff
@@ -0,0 +1,87 @@
...
+  Index residueId();
+  void setResidueId(Index& number);
+
+  char chainId();
+  void setChainId(char& id);
+
+  /** Adds an atom to the residue class */
+  void addResidueAtom(std::string& name, Atom& atom);
+
+  /** Returns a vector of atoms in the */
```
ghutchis: Should be ``@return a vector of atoms in ``` ***???***
(last part of comment is missing)

avogadro/core/residue.cpp
```diff
@@ -0,0 +1,129 @@
...
+{
+  std::vector<Atom> res;
+  for (AtomNameMap::iterator it = m_atomNameMap.begin();
+       it != m_atomNameMap.end(); ++it) {
+    res.push_back(it->second);
+  }
+  return res;
+}
+
+void Residue::resolveResidueBonds(Molecule& mol)
```
ghutchis: I'm not sure this best lives here. I mean, the code is okay, but if it's only in the Residue class, then to process a PDB file, you need to call this each time you finish a residue? Wouldn't you rather have a Molecule call or a separate ResidueBondPerception class that handles the whole molecule at once?

badarsh2: Agreed. So, instead of calling the method upon every new residue, how about modifying the Molecule class to store an array of residues and invoke a function such that once the reading is done, call the above method for each and every residue instance from the residue array?

avogadro/core/residue.h
badarsh2: Apologies :( Somehow it went unnoticed.

avogadro/core/residuedata.h
badarsh2: Sure. Can I save a local copy of resData (or if it is possible to retrieve resData.txt from the OpenBabel prefix, how do I go about) and write a parser to fetch data from the file?

ghutchis: No, I think it's better to have code like this - I'm just saying that you can write a script that will automate the syntax you've used here. (Similar to `utilities/bodr/bodrparse.cxx`)

avogadro/core/residue.cpp
ghutchis: Yes, exactly. Clearly the Molecule class needs to track the array of residues (if present).

badarsh2:
I've made some changes: I've now added a facility to store residues in the `Molecule` class, and also added utility to generate residue data header from resdata-like text file. I've also added a bond perception in the molecule class, so the bond resolving method does not have to be called upon every new residue. I thought I'll leave the default residueId / chainID assignment part for now. 

There are a couple other things I just noticed: In some residues, there are different position indicators, indicating alternate position possibilities. Though it is something to be dealt with in the PDB reader, I would like to know how the residue creation can be handled in such cases, because I believe that with the current implementation, only the 'A' or the firstly occurring position indicated atoms gets the bonds assigned.

cryos:
I am not convinced that having an array of residue objects is going to be the most scalable approach, but I don't think it will hurt in the short term to push in this direction, get things working, and see how it turns out. I still don't see a test exercising the new functionality, and I think that will help highlight how the API is intended to work. Thanks for working on this, I don't think PDB is central to the MD work, but it is a welcome addition.

badarsh2:
Though PDB may not be central to the MD work, I've encountered it being used as a common starting format before the molecule is modified and converted to other formats for MD simulations. Hence I thought I can help completing the long pending PDB reader classes.

If you would like me to take this forward now, it would be great if you can direct me on the same: what are the changes to be made in the current implementation. Or otherwise I can wait and complete it later as well.

ghutchis:
Sorry - I thought this was pending further review from Marcus. I think the current version is fine for now and can be revised if needed.

I'm not sure what you mean by this:

> There are a couple other things I just noticed: In some residues, there are different position indicators, indicating alternate position possibilities.

I admit, I'm not an expert on PDB parsing, but that's probably something to bring up in handling PDB files.

(Yes, handling PDB would be a big help - among other things, it would make initial peptide and RNA/DNA builders much easier.)

badarsh2:
Taking the case of [this](https://www.rcsb.org/structure/1yrf) particular protein, and the following atoms in particular,
```
ATOM     55  OE1AGLU A  45       3.359  10.202   1.014  0.32 26.48           O  
ATOM     56  OE1BGLU A  45       3.806   9.109   5.768  0.68 34.43           O  
ATOM     57  OE2AGLU A  45       3.672  11.747   2.509  0.32 37.15           O  
ATOM     58  OE2BGLU A  45       1.631   9.146   6.155  0.68 25.44           O 
```
the 17th index represents the position indicator (1,1,2 and 2 here respectively). Here, atom 55 and 57 represent the same atom but on account of different position possibilities for the atom, they have been given two different atom IDs denoting the two position possibilities. Likewise is the case of 56 and 58 as well. Quoting from [this](https://www.biostars.org/p/105413/) link:

> When interpreting the x-ray diffraction and its correspondent electron density map, sometimes the position the a particular atoms is not crystal clear (pun intended).

I haven't addressed it in the residue class, I guess it's better to handle such cases in the PDB reader.

ghutchis:
I'm going to merge this and we can refine from here. I'm not sure how to handle cases where there are two atoms with different IDs - we can discuss that in work for the PDB reader.

PR #307, opened 2018-07-14
Proposed shorter source file preamble for project

cryos:
After some research into best practices, copyright law, and tradition
versus requirements I am proposing this much shorter source file
preamble. The LICENSE file is the only full license, and should be
kept with source files moved into other projects, git keeps far better
track of who owns the copyright of any given source code, the developer
certificate of origin offers assurance that those developers have the
right to contribute code under the license.

The lines help where files are copied, and it is possible to see what
project they came from, and the unambiguous license the code is released
under. When snippets of code are copied this change won't make things any
better or worse.

Signed-off-by: Marcus D. Hanwell <marcus.hanwell@kitware.com>

### DIFF
--- a/avogadro/core/array.h
+++ b/avogadro/core/array.h
@@ -1,17 +1,6 @@
 /******************************************************************************
-
   This source file is part of the Avogadro project.
-
-  Copyright 2011-2012 Kitware, Inc.
-
-  This source code is released under the New BSD License, (the "License").
-
-  Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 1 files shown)


ghutchis:
Fine with me. We might need to pick a particular day to revise the headers at once.

cryos:
Yes, I just wanted to check I hadn't missed anything. It likely won't cause as many conflicts as you might thing because we rarely touch the top lines of source files.

PR #311, opened 2018-07-15
Migrate Chemical JSON to nlohmann json

cryos:
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.


psavery:
Strangely enough, it looks like something in the Travis-CI build failed for gcc 4.8, but it still comes up as a pass. See [here](https://travis-ci.org/OpenChemistry/avogadrolibs/jobs/404250632#L4147) and [here](https://travis-ci.org/OpenChemistry/avogadrolibs/jobs/404250632#L4260). I can look into it.

Also, the build failure is because nlohmann json does not support gcc 4.8. gcc 4.9 is the earliest version of gcc it supports (see [supported compilers](https://github.com/nlohmann/json#supported-compilers)). Should we bump the earliest compiler we support to gcc 4.9?

cryos:
That is odd - I saw all the CI passed, and assumed we were good. I think GCC 4.9 is more than reasonable, once we start using pybind11 it has a similar minimum.

psavery:
It was reported as a success because `set -e` wasn't in the Travis-CI `build.sh` script.

Because `set -e` wasn't present, the script continued even with an error. And the last command in the build script was a success, so Travis-CI reported it as a success. PR #312 should fix that.

Right now, the only test that runs in Travis-CI is `AvogadroTests`. We can also look into adding other tests as well such as `AvogadroIOTests`.

psavery:
Travis-CI minimum gcc version is bumped to 4.9 in PR #314.
AvogadroIOTests are added in PR #313 (although a few of them are failing).

psavery:
This looks good to me for the most part. Everything worked fine in my basic testing in the GUI. It also passes all of the tests. The basis sets are written/read, which is nice.

One issue I noticed: there is a problem with repeated reading/writing of `moCoefficients` in `orbitals`. For instance, if I open 'd-only.fchk' from the data, write it out to 'd-only.cjson', `moCoefficients` looks fine. But then if I open 'd-only.cjson' and write it again to 'd-only2.cjson', I noticed that it says `"moCoefficients": null` in the cjson. You also get `No orbital cofficients found!` in the debug output.

A few things we can do in the future (possibly now, but not necessarily now):
- Add more tests
- Explore the possibility of reading in cube data (right now, it just writes it out)
- Eliminate all usage of cppjson so we can eliminate cppjson entirely. Current, the qtplugins `plugindownloader` and `importpqr` use cppjson. Also the quantumio `nwchemjson` uses cppjson.

cryos:
I will fix up the coefficients, I think migrating plugins away should be done in a separate PR, along with the other reader. I am eager to drop jsoncpp, but would rather not force too much into single PRs, there is already quite a bit of time sunk into this, and it has a number of improvements already.

cryos:
I think this is ready, it has a number of consistency improvements over the previous version. I want to make some follow up PRs to convert over plugins, and the other reader. Any objections to getting this merged? After having worked with it I prefer the API over all the other C++ JSON libraries I have worked with.

ghutchis:
No, I'm fine to merge. As I said, I'm inclined to push over time to use this for JSON parsing throughout. (I'd even suggest we replace some of the QJsonXXX use too.)

cryos:
I am with you @ghutchis, although I think we should be happy to use whichever of the two makes most sense, and work to get rid of jsoncpp as soon as we can.

Issue #149, opened 2016-12-28
Default input Generators, etc. should be pulled from Git repositories


ghutchis:
Either using CMake or Git, the following directories should be pulled from separate repositories:

* avogadro/qtplugins/quantuminput/inputGenerators -> openchemistry/avogenerators
* avogadro/qtplugins/scriptfileformats/formatScripts
* avogadro/qtplugins/workflows/scripts

IMHO, This will make these features more prominent, ensure that builds will use the latest versions, and make the installer/updater feature to work better

PR #303, opened 2018-07-10
Migrate to a unified Python script loading code.

ghutchis:
Eliminates duplicate code between different script types
but also gives one place to handle conflict resolution, etc.

PR #315, opened 2018-07-16
Search for the generators in a sibling repository

cryos:
Remove the generators from the repository, and search for them in a
sibling repository, this will be added to the superbuild so that it
is present when packaging.

Signed-off-by: Marcus D. Hanwell <marcus.hanwell@kitware.com>

Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.


cryos:
This fixes #149 when coupled with the superbuild update.

ghutchis:
This is okay, but I'd much rather match everything with *.py and then use LIST(REMOVE_ITEM ..) to remove the test script. 

Consider that I just added the orca.py script and I'm sure we'll gain more over time.

ghutchis:
I'm also troubled by the use of the "unique hidden file" - why is this needed?

cryos:
It doesn't seem that troubling to me, it is a really simple double check that the repository is what we think it is. You could also have a directory with that name that happens to be something else, and I wanted to be reasonably certain as the repository isn't necessarily in a clean superbuild.

I added the orca file, I think the risk is minimal, I want to ensure the build is consistent. Globs can get out of date, but I could take a look at adding a CMake file listing all plugins in the repository. I was hoping to keep it simple, but agree it would  be preferable to keep a listing of files to be installed in the repository. I can likely make that the uniquely named file too, and have it serve the dual purpose.

This is just aimed at packaging releases, and the generator SHA needs pushing forward before releases.

ghutchis:
You can use the plugin.json file:
```
grep url plugin.json  | cut -d '"' -f 4 # gives https://github.com/openchemistry/avogenerators
```

If you want the list of commands:
```
grep 'command' plugin.json | cut -d '"' -f 4
```

ghutchis:
The point of the plugin.json was to give a short manifest useful for the plugin loader (#303) the download feature, and eventually a plugin directory. We picked JSON because basically everything else was based on JSON.

It offers a list of commands, since plugin authors might want to have other files in the directory that aren't commands. (Personally, I'd suggest copying the whole `avogenerators` directory on install, since i18n efforts will introduce translation files.)

cryos:
The grep and cut commands are not available in all build environments, and so that would be a terrible idea. JSON is great, but CMake doesn't currently ingest it, I don't want to move to something where we are just globbing blindly as that is unreliable and would be a setback. You asked for recommendations, this is my recommendation, I see your point about some duplication, but think that it is worth it to have a reliable build environment while still supporting some separation.

cryos:
I see the desire for *one true way*, but don't think adding a single file that contains a listing is files to install is unreasonable. We have two distinct goals, to install and distribute the generators with the binary bundles, using specific SHAs specified in the superbuild, and to dynamically download updates between releases/outside of releases that can override these files should users wish.

Using hackish solutions to avoid adding a file that can specify to the CMake build system unambiguously what should be installed is unreasonable IMHO. Maybe we should just copy the avogenerator code periodically back into this source tree right before releases, and leave things as they are. That would make it obvious when new files are added, and give an opportunity to update the build system here instead.

I can close this PR and plan on that if preferred.

ghutchis:
The .cmake file is not a big deal. Obviously, I'd rather ingest the JSON, but it's fine. I was writing my comment about the same time you were making that change - so the wires crossed. The main point should be that `avogadrolibs` shouldn't keep a fixed list of scripts - it should get the list from `avogenerators`

I stand by my point that creating a unique hidden file as a way to identify a repository isn't a good solution. Are there really going to be many directories labeled as 'avogenerators' that happen to be in that particular path? Seems unlikely IMHO.

cryos:
The hidden file is a simple way to be much more certain that it isn't an empty directory where the --recursive/git submodule update --init hasn't been used, or a bad tree checked out, or other possibilities that can be more subtle to debug and mitigated by a simple check.

I think it is much cleaner keeping the listing in the other repo, and that conveniently gives a uniquely named file that has the information requested,  with a clean path to bail if not present. I prefer to err on easy to debug/increased certainty after working on complex builds for many years, and the cost is pretty minimal. I want to keep things as robust as reasonable.

ghutchis:
The appveyor build has been really flaky. I'm okay to merge if you're okay with ignoring that build failure.

(Not sure if there's another way to get Windows build tests.)

cryos:
It is obviously not related to this branch, merging, then I will prepare an updated superbuild commit that uses the new avogadrolibs and avogadrogenerators SHAs. Thanks for talking this through, I think this is a reasonable compromise that will be pretty robust.

PR #318, opened 2018-07-18
Alpha beta

cryos:
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

### DIFF
--- a/avogadro/io/cjsonformat.cpp
+++ b/avogadro/io/cjsonformat.cpp
@@ -250,22 +250,21 @@ bool CjsonFormat::read(std::istream& file, Molecule& molecule)
     if (orbitals.is_object() && basis->isValid()) {
       basis->setElectronCount(orbitals["electronCount"]);
       json moCoefficients = orbitals["moCoefficients"];
-      json moCoefficientsA = orbitals["alpha"];
-      json moCoefficientsB = orbitals["beta"];
+      json moCoefficientsA = orbitals["alphaCoefficients"];
+      json moCoefficientsB = orbitals["betaCoefficients"];
       if (isNumericArray(moCoefficients)) {
         std::vector<double> coeffs;
         for (unsigned int i = 0; i < moCoefficients.size(); ++i)
           coeffs.push_back(static_cast<double>(moCoefficients[i]));
         basis->setMolecularOrbitals(coeffs);
       } else if (isNumericArray(moCoefficientsA) &&
-                 isNumericArray(moCoefficientsB) &&
-                 moCoefficientsA.size() == moCoefficientsB.size()) {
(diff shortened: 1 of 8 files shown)


ghutchis:
I'll build and review a bit later, but this will be a definite advance over Avo1.x.

avogadro/io/cjsonformat.cpp
```diff
@@ -387,13 +398,23 @@ bool CjsonFormat::write(std::ostream& file, const Molecule& molecule)
...
         moCoefficients.push_back(moMatrix(i, j));
-      }
+
+    if (betaMatrix.cols() > 0 && betaMatrix.rows() > 0) {
+      json moBeta;
+      for (int j = 0; j < moMatrix.cols(); ++j)
+        for (int i = 0; i < moMatrix.rows(); ++i)
+          moBeta.push_back(moMatrix(i, j));
+
+      root["orbitals"]["alpha"] = moCoefficients;
```
ghutchis: I think it should be clearer that these are coefficients. The key could also imply energies/eigenvalues.

How about "alphaCoeff" or "alphaCoefficients"

avogadro/io/cjsonformat.cpp
```diff
@@ -257,6 +257,17 @@ bool CjsonFormat::read(std::istream& file, Molecule& molecule)
         for (unsigned int i = 0; i < moCoefficients.size(); ++i)
           coeffs.push_back(static_cast<double>(moCoefficients[i]));
         basis->setMolecularOrbitals(coeffs);
+      } else if (isNumericArray(moCoefficientsA) &&
+                 isNumericArray(moCoefficientsB) &&
+                 moCoefficientsA.size() == moCoefficientsB.size()) {
```
ghutchis: I have a nagging feeling that some programs may not keep this last equivalence (e.g., # of alpha and # of beta orbitals may be different)

ghutchis:
My main concern is the choice of json key for alpha/beta MO coefficients. I'd love to see eigenvalue/energies and occupancies come through, but those can be separate patches.

avogadro/io/cjsonformat.cpp
cryos: Sounds good, this is pretty minor, I can update it.

avogadro/io/cjsonformat.cpp
cryos: I wondered, I can fix the reader to be a little more relaxed/do two loops...

cryos:
I think this addresses your concerns, absolutely agree about feeding more data through in follow up patches. I really want to get to a point where this format fully round trips, and can help diagnose if other formats are being parsed correctly.

PR #319, opened 2018-07-19
WIP: Forcefield framework

ghutchis:
This is non-functional work in progress. It will use the cppoptlib header-only library for optimization.

`cppopt` requires value(Eigen::VectorXd) methods for the function call.

We can cast fairly easily from positions3d()
```
  int n = m_molecule->atomCount();
  Core::Array<Vector3> pos = m_molecule->atomPositions3d();
  double *p = pos[0].data();
  Eigen::Map<Eigen::VectorXd> map(p, 3*n);
  Eigen::VectorXd positions = map;
  Real energy = lj.value(positions);
```

The problem comes from translating the VectorXd *back* to a set of atom positions:
```
Real EnergyCalculator::value(const TVector& x)
{
  const double *d = x.data();
  m_positions.data() = d;
  return calculateEnergy(m_positions);
}
```

Ideas welcome.


cryos:
The documentation for Eigen::Map isn't great, but it looks like it copies the data. So it is a short way to copy the data, but a copy is taking place there. We can just do something like:
```
auto posPtr = &m_positions[0];
const double *d = x.data();
for (size_t i = 0; i < m_positions.size() * 3; ++i)
  *(posPtr++) = *(d++);
```
I didn't test it, I think that would work, and we could always add some utility code to help if this needs to be done a lot. It looks like the code is already making copies, so making one more isn't going to hurt, we could always look at improving things in the future if the memory copies look like a bottleneck. I didn't test the above, but something like that should work I think.

ghutchis:
Eigen::Map doesn't copy data by default. I'm copying because I want to make sure the data is aligned for vectorized math:
https://stackoverflow.com/a/40859170/131896

I'm using cppoptlib because it saves me a ton of code (e.g., implementing line searches, conjugate gradients, BFGS, etc.) and I figured it would be fairly easy to handle with some casting. As mentioned above, I can get away without a copy.

Yes, I *can* copy, but force fields take 100s of steps, so that's a lot of copying...

cryos:
I think you need to either take the double *, and update that directly, or pay the price for the copies. I don't  see a reasonable way to take a pointer to a double * array, and assume ownership of it. There may be something we can work out once this is working, but I am not sure it is worth it unless we see that the copies are a bottleneck.

The ability to cast to double * lets you pass the array into other things that will operate on raw arrays, but they have to operate directly on that memory. You need to get cppoptlib to operate directly on the double * array passed in to avoid the copies, in which case there is no need to ever pass this back into the the Array<Vector3>.

PR #320, opened 2018-07-21
WIP: LAMMPS Input generator

badarsh2:
image

Most of the raw content for this has been ported from Avogadro 1. A new modification I've brought here is the direct previewing of the molecule data file; in Avogadro 1, only the main file could be created using the input generator class and the data file had to be saved manually using the export option. Here I've made it such that the FileFormat class can be used to write out the molecule in a supported format and the output can be previewed alongside the main config file. In addition to that, I've made some other minor changes, for instance, the water potential option has been removed for now, as it requires some molecule-side information from the parent generator class. 

For now, the LAMMPS lmpdat data file is generated using OpenBabel writers; native lmpdat writer is still WIP. 

To check whether the generator does the job correctly, I tried generating inputs for bulk Silicon high temperature simulation. I had to manually add an extra line in the config file to read the Stillinger-Weber potentials for Silicon. Upon doing that and running the simulation using locally installed LAMMPS, it got completed without errors.

------------------------------------------------------------------------------------------------------------------
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.


cryos:
This patch will need some work, we have moved the input generator Python code to a separate repository so that it can be updated between releases. I will update our documentation too, but it is important to only run clang-format on source that was changed, and source that is C++, i.e. CMakeLists.txt won't work. We need to use the git integration to only run it on the diff too, and I might need to update the .clang-format a little as it is really tough seeing what you actually changed versus unrelated changes made.

ghutchis:
IMHO, for a complex input generator like this, the best approach will be to implement straight-up in Qt like APBS or GAMESS. It'll be easier to lay out the UI in Qt Designer, rather than hacking the input generator code for a lot of special cases.

cryos:
I was wondering if that might be better too, and it wouldn't hurt to have a few more examples of specialized C++ input generators. You get events, more freedom, and we could always look at generalizing MD input in the future, but the current input generator code never considered MD.

cryos:
No worries, it motivated me to improve our documentation a little, and point out a better way to do clang-format. I think a C++ generator would likely be the best solution at this stage.

PR #339, opened 2018-07-29
Added yaehmop with band structure calc

psavery:
This is still a work in progress, but this commit adds
yaehmop with band structure calculations. It runs, but
several features (especially in the GUI) still need to
be coded.

Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.


psavery:
As you can see from the images below, I've managed to get the basic functionality of yaehmop and the yaehmop extension working in Avogadro2. I can generate a band structure similar to the one in Avogadro1.

Avogadro1 yaehmop band structure:
avogadro1-band-structure

Avogadro2 yaehmop band structure:
avogadro2-band-structure

The current Avogadro2 band structure dialog (it is very similar to the Avogadro1 band structure dialog):
image

One really nice feature of the VTK plot in Avogadro2 is that it will tell you the x and y values of the curves if you put your mouse over a spot on the curve.

There are still several things that need to be done before this PR is complete, though.

- Marcus, do you want me to move the yaehmop download to the OpenChemistry repo? If that's the case, should yaehmop download when the user first runs cmake (assuming the correct cmake options are set)?
- I need to make it so that spglib can put unit cells in their standard orientation
- I need to add automatic k-point generation based on the detected space group
- I need to make it so the vtk plot can use custom tick markers so we can use the special k points on the x axis
- Several gui elements in the band structure dialog don't actually do anything right now (such as setting the fermi level). I need to make all the gui elements work.


Probably more stuff needs to be done, too...

Let me know if you have any thoughts or concerns about this, though.

cryos:
Patrick, yes yaehmop should be made into an external project in the superbuild, avogadrolibs should depend on it, and the download should happen in the build step of that. At some point it would be good to make the option of download the binary or build it, but that can certainly be pushed off until later. We can talk about this tomorrow if that helps. The custom axis markers for the chart should be pretty easy, you can just pass the value numerically and what the label should be.

ghutchis:
Outside of this pull request, I should have some time before the UGM to code up parsing MO coefficients, etc. It's behind the force field work in priority, but I'd like to enable yaehmop for molecules.

psavery:
Unfortunately, it still is WIP. Several buttons in the pop-up dialog do not work. I should probably at least get them working (or remove them) before merging.

We should also merge the PR mentioned in my previous comment before this one.

psavery:
This PR is almost complete. I just need to add the ability to plot and adjust the fermi level to the band structures.

It now also depends on PR #379.

I can squash some of these commits when complete if needed.

avogadro/core/molecule.h
```diff
@@ -583,6 +592,9 @@ class AVOGADROCORE_EXPORT Molecule
   UnitCell* m_unitCell;
   Array<Residue> m_residues;
 
+  // This will be stored from the last space group operation
+  unsigned short m_hallNumber = 0;
```
psavery: @cryos 
What do you think about initializing class members at their declaration like this?

Personally, I think it is a nice addition in C++11, since it allows for a default value for all constructors that can be overridden (if desired) by a constructor's member initializer list.

psavery:
By the way, this PR is pretty much complete and ready for review. The only thing I am waiting for is PR #379, since that is a dependency for this PR.

ghutchis:
I'm going to merge this. I'd love a better way to check the whole PATH, but that's a separate issue...

PR #353, opened 2018-08-11
Surface animation

cryos:


cryos:
I will fix things up in post review, trying to get some things finished up, and don't want to create a huge topic branch, or interdependent branches. There are a lot of nice improvements in here, and we can certainly discuss some of the finer points of animating surfaces which is pretty much hidden unless you have a multi-step CJSON file at this stage. I will add write functionality soon too.

PR #358, opened 2018-08-13
ImportPQR: move to nlohmann json

psavery:
Part of the transfer to use nlohmann json instead of cpp json, this
eliminates the cpp json usage in the import pqr plugin to use nlohmann
json instead.

I also fixed a few memory leaks that were present.

Signed-off-by: Patrick Avery <psavery@buffalo.edu>

Developer Certificate of Origin
Version 1.1

Copyright (C) 2004, 2006 The Linux Foundation and its contributors.
1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

Everyone is permitted to copy and distribute verbatim copies of this
license document, but changing it is not allowed.

Developer's Certificate of Origin 1.1

By making a contribution to this project, I certify that:

(a) The contribution was created in whole or in part by me and I
    have the right to submit it under the open source license
    indicated in the file; or

(b) The contribution is based upon previous work that, to the best
    of my knowledge, is covered under an appropriate open source
    license and I have the right under that license to submit that
    work with modifications, whether created in whole or in part
    by me, under the same open source license (unless I am
    permitted to submit under a different license), as indicated
    in the file; or

(c) The contribution was provided directly to me by some other
    person who certified (a), (b) or (c) and I have not modified
    it.

(d) I understand and agree that this project and the contribution
    are public and that a record of the contribution (including all
    personal information I submit with it, including my sign-off) is
    maintained indefinitely and may be redistributed consistent with
    this project or the open source license(s) involved.

PR #355, opened 2018-08-13
Plugin Downloader: move to nlohmann json

psavery:
Part of the transfer to use nlohmann json instead of cpp json, this
eliminates the cpp json usage in the plugin downloader to use nlohmann
json instead.

I also fixed a few memory leaks that were present.

Signed-off-by: Patrick Avery <psavery@buffalo.edu>

Developer Certificate of Origin
Version 1.1

Copyright (C) 2004, 2006 The Linux Foundation and its contributors.
1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

Everyone is permitted to copy and distribute verbatim copies of this
license document, but changing it is not allowed.

Developer's Certificate of Origin 1.1

By making a contribution to this project, I certify that:

(a) The contribution was created in whole or in part by me and I
    have the right to submit it under the open source license
    indicated in the file; or

(b) The contribution is based upon previous work that, to the best
    of my knowledge, is covered under an appropriate open source
    license and I have the right under that license to submit that
    work with modifications, whether created in whole or in part
    by me, under the same open source license (unless I am
    permitted to submit under a different license), as indicated
    in the file; or

(c) The contribution was provided directly to me by some other
    person who certified (a), (b) or (c) and I have not modified
    it.

(d) I understand and agree that this project and the contribution
    are public and that a record of the contribution (including all
    personal information I submit with it, including my sign-off) is
    maintained indefinitely and may be redistributed consistent with
    this project or the open source license(s) involved.

PR #359, opened 2018-08-13
NWChemJson: Use nlohmann json

psavery:
Part of the transfer to use nlohmann json instead of cpp json, this
eliminates the cpp json usage in quantumio to use nlohmann
json instead.

Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.


psavery:
This compiles, but it is untested. I'm not convinced that it works, nor that it won't crash upon usage.

I still need to look over the changes some more.

psavery:
Hi Marcus @cryos, 

This is now ready for review.

I tested it using [this file](https://github.com/wadejong/NWChemOutputToJson/blob/master/prop_h2o.out.json) and compared running the jsoncpp version of this file to the nlohmann json version.

I was able to do "Quantum"->"Vibrational modes..." for both, and they matched.

I was able to do "Extensions"->"Create Surfaces...", for the HOMO and LUMO molecular orbitals, and they both matched.

Then, I wrote out a .cjson file for both. Both files were identical and contained things such as the basisSet, orbitals, and vibrations.

Let me know if there is anything else I should test or change.

Thanks,
Patrick

psavery:
If we get this one, PR #355, and PR #358 merged, I think we will have effectively removed the jsoncpp dependencies from Avogadro2.

cryos:
I want hold off on this one until I can verify things work, I would rather leave jsoncpp a little longer if necessary. I will dig out the files I have, or see if we can generate a few new ones.

psavery:
Okay, that's fine. The PR can stay here as long as needed till it is ready to merge.

avogadro/quantumio/nwchemjson.cpp
```diff
@@ -47,112 +45,116 @@ using Core::CrystalTools;
...
 
 bool NWChemJson::read(std::istream& file, Molecule& molecule)
 {
-  Value root;
-  Reader reader;
-  bool ok = reader.parse(file, root);
-  if (!ok) {
-    appendError("Error parsing JSON: " + reader.getFormatedErrorMessages());
+  json root;
+  try {
```
cryos: I am guessing you are using this as you want to get at the parse error object? If you want to do this, then I think we need to do the same for the Chemical JSON format too, probably preferable but I would like us to be consistent.

psavery: Well, I originally wanted to do it because nlohmann json always throws an exception when there is a parse error, and I didn't want Avogadro2 to crash while parsing an invalid json file.

Since we are catching the exception anyways, I thought it'd be useful to print out the error message.

PR #369, opened 2018-08-27
Changes for unit cell systems - always save real-space & unit axes

ghutchis:

Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

### DIFF
--- a/avogadro/io/cjsonformat.cpp
+++ b/avogadro/io/cjsonformat.cpp
@@ -192,20 +192,33 @@ bool CjsonFormat::read(std::istream& file, Molecule& molecule)
   json unitCell = jsonRoot["unit cell"];
   if (!unitCell.is_object())
     unitCell = jsonRoot["unitCell"];
-  if (unitCell.is_object() && unitCell["a"].is_number() &&
-      unitCell["b"].is_number() && unitCell["c"].is_number() &&
-      unitCell["alpha"].is_number() && unitCell["beta"].is_number() &&
-      unitCell["gamma"].is_number()) {
-    Real a = static_cast<Real>(unitCell["a"]);
-    Real b = static_cast<Real>(unitCell["b"]);
-    Real c = static_cast<Real>(unitCell["c"]);
-    Real alpha = static_cast<Real>(unitCell["alpha"]) * DEG_TO_RAD;
-    Real beta = static_cast<Real>(unitCell["beta"]) * DEG_TO_RAD;
-    Real gamma = static_cast<Real>(unitCell["gamma"]) * DEG_TO_RAD;
-    Core::UnitCell* unitCellObject =
-      new Core::UnitCell(a, b, c, alpha, beta, gamma);
-    molecule.setUnitCell(unitCellObject);
+
+  if (unitCell.is_object()) {
+    Core::UnitCell* unitCellObject = nullptr;
+
+    // read in cell vectors in preference to a, b, c parameters
+    json cellVectors = unitCell["cellVectors"];
+    if (cellVectors.is_array() && cellVectors.size() == 9 &&
+        isNumericArray(cellVectors)) {
+      Vector3 aVector(cellVectors[0], cellVectors[1], cellVectors[2]);
+      Vector3 bVector(cellVectors[3], cellVectors[4], cellVectors[5]);
+      Vector3 cVector(cellVectors[6], cellVectors[7], cellVectors[8]);
+      unitCellObject = new Core::UnitCell(aVector, bVector, cVector);
+    } else if (unitCell["a"].is_number() && unitCell["b"].is_number() &&
+               unitCell["c"].is_number() && unitCell["alpha"].is_number() &&
+               unitCell["beta"].is_number() && unitCell["gamma"].is_number()) {
+      Real a = static_cast<Real>(unitCell["a"]);
+      Real b = static_cast<Real>(unitCell["b"]);
+      Real c = static_cast<Real>(unitCell["c"]);
+      Real alpha = static_cast<Real>(unitCell["alpha"]) * DEG_TO_RAD;
+      Real beta = static_cast<Real>(unitCell["beta"]) * DEG_TO_RAD;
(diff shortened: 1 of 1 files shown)


ghutchis:
This will be useful for solid-state codes. Most use real-space coordinates and translation axes, instead of cell parameters. The files are backwards-compatible, but offer additional fields on write.

Technically, this is WIP, since I haven’t added read-support - just want to make sure there are no complaints or issues before I do that.

psavery:
Yeah, there is some benefit in explicitly providing the unit cell matrix. Some calculations do depend on the orientation of the unit cell.

But if we have the unit cell explicitly, we won't need the A, B, C, alpha, beta, gamma for the unit cell (it is information that can be derived from the unit cell matrix and therefore repetitive).

Also, I don't think we would necessarily need both cartesian and fractional coordinates, because one can convert from one to the other as well. We could just choose one to stick with.

cryos:
I don't like the duplication aspect, but Chemical JSON is designed to have optional fields. I have been thinking about this for a while, could we pass in a JSON string with settings for readers/writers? They could be useful for commands, i.e. just give elements, give elements and proton number, explicit Cartesian, etc.

The default writer could remain pretty minimal, but some options could affect behavior.

avogadro/io/cjsonformat.cpp
```diff
@@ -380,6 +380,20 @@ bool CjsonFormat::write(std::ostream& file, const Molecule& molecule)
...
+    vectors.push_back(molecule.unitCell()->aVector().z());
+    
+    vectors.push_back(molecule.unitCell()->bVector().x());
+    vectors.push_back(molecule.unitCell()->bVector().y());
+    vectors.push_back(molecule.unitCell()->bVector().z());
+
+    vectors.push_back(molecule.unitCell()->cVector().x());
+    vectors.push_back(molecule.unitCell()->cVector().y());
+    vectors.push_back(molecule.unitCell()->cVector().z());
+    root["cell vectors"] = vectors;
```
cryos: I have been using camel case for new keys, they are easier to use in all languages, i.e. cellVectors, I was hoping avoid any new keys with spaces in them.

cryos:
Would a text container work, and for CJSON it would be JSON, for other formats it could be other things. Yes, I had been feeling the need for options for a while @ghutchis. I will see if I can put together an example, it could also be requested as a simple JSON string and the Open Babel plugin would then turn it into command line arguments, other readers unpack as they choose.

avogadro/io/cjsonformat.cpp
ghutchis: I’m coming back to this now. I’m happy to use cellVectors as the key. I’m not sure what you mean by “vectors” for the storage? Do you want to store them as [ [ax, ay, az], [bx, by, bz], [cx, cy, cz] ] instead? That’s fine, but it’s different from how we store atomic coordinates - thus I went with an array of 9 values.

cryos: I meant the key name, that was all. I actually want to use this in our server platform now too. I was suggesting storing the whole thing in the "unit cell" key as "vectors" as the "unit cell" already makes it clear that it is defining the molecule's unit cell. I think the way you stored it makes sense.

ghutchis: Okay, I’ll get that done later this afternoon, thanks.

ghutchis:
I’m a bit unsure on how to proceed here. If I drop the cell vectors into “unit cell” then it’s no longer a dictionary. I can do that, but it won’t be backward compatible.

I think I should just have “unit cell” contain “cellVectors” in addition to “a”, “b”, “c”, etc.

ghutchis:
Example:
```
  "unit cell": {
    "a": 3.0,
    "alpha": 90.0,
    "b": 3.0,
    "beta": 90.0,
    "c": 3.0,
    "cellVectors": [
      3.0,
      0.0,
      0.0,
      -1.4999999999999993,
      2.5980762113533165,
      0.0,
      1.8369701987210297e-16,
      3.18172571617472e-16,
      3.0
    ],
    "gamma": 119.99999999999999
  }
```

cryos:
> I’m a bit unsure on how to proceed here. If I drop the cell vectors into “unit cell” then it’s no longer a dictionary. I can do that, but it won’t be backward compatible.

Not sure what you mean, referring to Python dictionary? I am happy to do the work, but I think that what is produced should be set by an option, i.e. default to saving what we had before, and if the option is set then switch to Cartesian and your new cell vectors. I don't think it is useful to save both by default.

I want to start using this, and it is a relatively small change. I really want to avoid duplicating data unless that is requested in an option for the format.

ghutchis:
Honestly, I don’t think the fractional coordinates are that useful. Is anything external using them? As far as formats, mostly CIF stores in fractional - nearly everything else uses real-space Cartesians.

What I meant by “dictionary” was the JSON - guess I should have said “object.” I definitely don’t see the point of using an option - most people will want the unit cell vectors - I had several requests at the UGM for easy access to them. I also think that dropping a, b, c, alpha, beta, gamma isn’t that big a deal - it takes up only a few extra bytes.

But if you have a better solution, by all means, it doesn’t matter much. I’d just insist that real-space Cartesians should be the default with unit cell axes. Almost all of the time, that’s what people want IMHO.

cryos:
I don't have strong feelings on the defaults, and went with what the people who had worked with materials more in the past did. I do have strong feelings about not duplicating data, and trying to remain minimal, for larger molecules storing the Cartesian and fractional coordinates would be quite a cost.

There seems to be enough of a split that offering both would be reasonable, but switching to make Cartesian coordinates the default. We want to use this format to store data in MongoDB, and to transmit over RESTful APIs, so reducing duplication is always a concern, but we didn't relish the idea of writing TypeScript to deal with fractional coordinates.

cryos (approved):
I will push a follow PR later, I think we can merge what we have now, and do that later.

PR #370, opened 2018-08-27
Initial MMTF Support

frodofine:
This commit introduces read support for the MMTF file format and is a WIP.

The msgpack library is quite large, so I am wondering if it should be GZipped and extracted during the build (this is done by other libraries).

Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.


cryos:
Really high level comment, I think the MMTF library is too big for thirdparty, and should be added to our superbuild. We put a few small libraries in thirdparty, but anything of this kind of size we wouldn't. Does it stick with C++11 as its minimum, I see some higher C++ versions, but maybe they are optional? I can help with the third party, but we would need to edit out the mmtf from the history on this branch too (happy to help there too).

frodofine:
Can you explain the 'superbuild'? I think you are correct and MMTF/msgpack are too big.

The msgpack library queries the compiler version and determines what version of C++ to use. The lowest version it supports is C++03. The version can be forced if needed, however.

I'll probably squash out the current version anyway.

cryos:
We have a repository to build Avogadro and its dependencies. We build glew, libmsym, etc in there, and pass the location of the install prefix to Avogadro.

https://github.com/OpenChemistry/openchemistry

We use CMake and external project, it makes it simpler to bump as we can simply track upstream, it will either download a tarball which is generally preferred, or can pull from git etc. We require C++11, so that is what we want to target, if it optionally uses newer that is fine.

PR #371, opened 2018-08-27
Initial MMTF support

frodofine:
I'm not sure what happened to the original PR, but here is a rebased version. It requires the changes in openchemistry project to be merged before the CI will pass.

Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.


cryos:
This looks good, it works but there only seems to be one frame and the unit cell doesn't look right. Screenshot below, if this is expected then things look good to me!
image

ghutchis:
I don't think it's an animation formation - I'm pretty confident it should just be one frame.

I'd check the unit cell against the PDB rendering in Avogadro v1 for the same PDB code.

cryos:
I was honestly asking @frodofine if both were expected, like I said if so I think this is ready to go.

frodofine:
That's normal. The crystal cells in the PDB are notoriously bad (the atoms of the protein are rarely inside of the PBC). PyMOL and other renders have similar issues.

PR #372, opened 2018-08-28
File format options

cryos:


avogadro/io/cjsonformat.cpp
```diff
@@ -363,14 +363,22 @@ bool CjsonFormat::read(std::istream& file, Molecule& molecule)
 
 bool CjsonFormat::write(std::ostream& file, const Molecule& molecule)
 {
+  json opts;
+  if (!options().empty())
+    opts = json::parse(options());
```
psavery: I think it is worth considering to, in general, wrap nlohmann json::parse() statements in a try/catch block. nlohmann json throws an exception when there is a parse error (and as far as I know, there is no way to switch it off).

cryos: You can, I forgot to pass in the optional arguments, see the parse documentation, nullptr as the call back and then false to disable exceptions.

psavery: Oh okay. Sounds good. Good to know!

Issue #335, opened 2018-07-27
Per-atom color or render options


ghutchis:
**Is your feature request related to a problem? Please describe.**

Many times you wish to add contrast to a particular atom or subset of interest (e.g., the Fe atom in a heme protein). In Avo2, one cannot change rendering options on an individual atom or selected subset.

**Additional context**

For example, in this image from Avo1, the HEM residues are rendered with Van der Waals while the remainder of the protein is rendered with wireframe and cartoon options:
![2hhb](https://user-images.githubusercontent.com/41128/43335944-1c2da948-919e-11e8-897d-0812df88c04d.png)

ghutchis:
This is actually supported - just needs to become more obvious in the UI.

PR #376, opened 2018-08-29
Custom atom color support

frodofine:
This is a WIP.

The purpose of this PR is to add specialized coloring options for atoms in Avogadro. Currently, this can only be done when the atom is added through the Editor plugin, but before this is merged I plan on adding support for changing the atom color with the Select plugin.

I have a quick question about how Colors are supported in Avogadro. `Elements::Color` returns an `unsigned char*`. Should this be changed to a `Vector3ub` or even an `unsigned int` as is done with `QRgb`?

Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

### DIFF
--- a/avogadro/core/atom.h
+++ b/avogadro/core/atom.h
@@ -156,6 +156,14 @@ public:
   /** @} */
 
   /**
+   * The color of this atom
+   * @{
+   */
+  void setColor(Vector3ub color);
+  Vector3ub color() const;
+  /** @} */
+
+  /**
    * Is the atom selected.
    * {@
    */
@@ -309,6 +317,18 @@ template <class Molecule_T>
 signed char AtomTemplate<Molecule_T>::formalCharge() const
 {
   return m_molecule->formalCharge(m_index);
+}
+
+template <class Molecule_T>
+void AtomTemplate<Molecule_T>::setColor(Vector3ub color)
+{
+  m_molecule->setColor(m_index, std::move(color));
+}
+
+template <class Molecule_T>
+Vector3ub AtomTemplate<Molecule_T>::color() const
+{
+  return m_molecule->color(m_index);
 }
 
 template <class Molecule_T>
--- a/avogadro/core/molecule.cpp
+++ b/avogadro/core/molecule.cpp
@@ -40,7 +40,7 @@ Molecule::Molecule(const Molecule& other)
     m_atomicNumbers(other.atomicNumbers()), m_positions2d(other.m_positions2d),
     m_positions3d(other.m_positions3d), m_coordinates3d(other.m_coordinates3d),
     m_timesteps(other.m_timesteps), m_hybridizations(other.m_hybridizations),
-    m_formalCharges(other.m_formalCharges),
+    m_formalCharges(other.m_formalCharges), m_colors(other.m_colors),
     m_vibrationFrequencies(other.m_vibrationFrequencies),
     m_vibrationIntensities(other.m_vibrationIntensities),
     m_vibrationLx(other.m_vibrationLx), m_bondPairs(other.m_bondPairs),
@@ -75,6 +75,7 @@ Molecule::Molecule(Molecule&& other) noexcept
     m_timesteps(std::move(other.m_timesteps)),
     m_hybridizations(std::move(other.m_hybridizations)),
     m_formalCharges(std::move(other.m_formalCharges)),
+    m_colors(std::move(other.m_colors)),
     m_vibrationFrequencies(std::move(other.m_vibrationFrequencies)),
     m_vibrationIntensities(std::move(other.m_vibrationIntensities)),
     m_vibrationLx(std::move(other.m_vibrationLx)),
@@ -105,6 +106,7 @@ Molecule& Molecule::operator=(const Molecule& other)
     m_timesteps = other.m_timesteps;
     m_hybridizations = other.m_hybridizations;
     m_formalCharges = other.m_formalCharges;
+    m_colors = other.m_colors,
     m_vibrationFrequencies = other.m_vibrationFrequencies;
     m_vibrationIntensities = other.m_vibrationIntensities;
     m_vibrationLx = other.m_vibrationLx;
@@ -152,6 +154,7 @@ Molecule& Molecule::operator=(Molecule&& other) noexcept
     m_timesteps = std::move(other.m_timesteps);
     m_hybridizations = std::move(other.m_hybridizations);
     m_formalCharges = std::move(other.m_formalCharges);
+    m_colors = std::move(other.m_colors);
     m_vibrationFrequencies = std::move(other.m_vibrationFrequencies);
     m_vibrationIntensities = std::move(other.m_vibrationIntensities);
     m_vibrationLx = std::move(other.m_vibrationLx);
@@ -244,6 +247,16 @@ Array<signed char>& Molecule::formalCharges()
 const Array<signed char>& Molecule::formalCharges() const
 {
   return m_formalCharges;
+}
+
+Array<Vector3ub>& Molecule::colors()
+{
+  return m_colors;
+}
+
+const Array<Vector3ub>& Molecule::colors() const
+{
+  return m_colors;
 }
 
 Array<Vector2>& Molecule::atomPositions2d()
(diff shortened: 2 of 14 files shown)


avogadro/core/atom.h
```diff
@@ -155,6 +155,15 @@ class AtomTemplate
   signed char formalCharge() const;
   /** @} */
 
+  /**
+   * The color of this atom
+   * @{
+   */
+  void setColor(Vector3ub charge);
```
cryos: I think you mean color

cryos:
I want to maintain support for coloring by atomic number, and not setting a color in the default path. If you want to set colors then that can happen but I think it needs to be done lazily. With some minor adjustment what you have done can accomplish this.

Leave the default path for adding atoms alone, and continue to not set any color. Change the rendering code to query the atom color, if the color array has not been set up then instead of returning black as you do now, look up the color based on the atomic number.

We also then have a path to save this out - if custom colors aren't used, we know we don't need to save a color array. If they are, then in formats with support we can save the custom colors.

avogadro/core/molecule.h
```diff
@@ -680,6 +710,31 @@ inline bool Molecule::setFormalCharge(Index atomId, signed char charge)
...
+    m_colors = colors;
+    return true;
+  }
+  return false;
+}
+
+inline bool Molecule::setColor(Index atomId, Vector3ub charge)
+{
+  if (atomId < atomCount()) {
+    if (atomId >= m_formalCharges.size())
```
cryos: Why are you using formatCharges size? Should be atomCount.

avogadro/core/molecule.h
```diff
@@ -680,6 +710,31 @@ inline bool Molecule::setFormalCharge(Index atomId, signed char charge)
...
+      m_colors.resize(atomCount(), {0, 0, 0});
```
cryos: Here, it might be better to iterate through added colors, and default to the standard RGBs for the element.

avogadro/core/molecule.h
```diff
@@ -680,6 +710,31 @@ inline bool Molecule::setFormalCharge(Index atomId, signed char charge)
...
+inline bool Molecule::setColors(const Core::Array<Vector3ub>& colors)
+{
+  if (colors.size() == atomCount()) {
+    m_colors = colors;
+    return true;
+  }
+  return false;
+}
+
+inline bool Molecule::setColor(Index atomId, Vector3ub charge)
```
cryos: You mean color I think, not charge.

avogadro/core/molecule.h
```diff
@@ -680,6 +710,31 @@ inline bool Molecule::setFormalCharge(Index atomId, signed char charge)
   return false;
 }
 
+inline Vector3ub Molecule::color(Index atomId) const
+{
+  return atomId < m_colors.size() ? m_colors[atomId] : Vector3ub(0, 0, 0);
```
cryos: Here is where I think we make the change to look up the standard color for the atomic number if the array is too short. This would let us color by atomic number as we have by default, and change in custom colors lazily for people who want to.

avogadro/core/molecule.cpp
```diff
@@ -354,6 +367,8 @@ bool Molecule::removeAtom(Index index)
       m_hybridizations[index] = m_hybridizations.back();
     if (m_formalCharges.size() == m_atomicNumbers.size())
       m_formalCharges[index] = m_formalCharges.back();
+    if (m_colors.size() == m_colors.size())
```
psavery: I think you meant `if (m_colors.size() == m_atomicNumbers.size())` here

avogadro/core/molecule.cpp
```diff
@@ -377,6 +392,8 @@ bool Molecule::removeAtom(Index index)
     m_hybridizations.pop_back();
   if (m_formalCharges.size() == m_atomicNumbers.size())
     m_formalCharges.pop_back();
+  if (m_colors.size() == m_colors.size())
```
psavery: I think you meant `if (m_colors.size() == m_atomicNumbers.size())` here (same as above)

ghutchis:
I think the feature is great - but can you explain why you'd want it as part of the editor? It seems like a more common feature is "select a bunch of atoms and set a particular color"..

ghutchis:
At the moment, I don't have time to do code review - seems like @cryos and @psavery have made good suggestions.

avogadro/core/molecule.h
```diff
@@ -182,6 +182,35 @@ class AVOGADROCORE_EXPORT Molecule
...
+   * (0,0,0) if @a atomId is invalid.
+   */
+  Vector3ub color(Index atomId) const;
+
+  /**
+   * Replace the current array of colors.
+   * @param charges The new color array. Must be of length atomCount().
+   * @return True on success, false otherwise.
+   */
+  bool setColors(const Core::Array<Vector3ub>& charges);
```
psavery: Probably meant `colors` here as well.

avogadro/qtgui/hydrogentools.cpp
```diff
@@ -123,6 +123,10 @@ void HydrogenTools::adjustHydrogens(RWMolecule& molecule, Adjustment adjustment)
            it != itEnd; ++it) {
         RWAtom newH(molecule.addAtom(1));
         newH.setPosition3d(*it);
+
+        const unsigned char* color = Core::Elements::color(1);
```
psavery: When we start using the default atomic colors (instead of black) for when the color isn't set, we probably won't need this section and a few other sections.

avogadro/qtgui/hydrogentools.cpp
```diff
@@ -192,6 +196,10 @@ void HydrogenTools::adjustHydrogens(RWAtom& atom, Adjustment adjustment)
          it != itEnd; ++it) {
       RWAtom newH(molecule->addAtom(1));
       newH.setPosition3d(*it);
+
+      const unsigned char* color = Core::Elements::color(1);
```
psavery: Same here.

avogadro/qtgui/rwmolecule.cpp
```diff
@@ -133,6 +134,7 @@ class AddAtomCommand : public RWMolecule::UndoCommand
   {
     assert(atomicNumbers().size() == m_atomId);
     atomicNumbers().push_back(m_atomicNumber);
+    colors().push_back(Vector3ub(0, 0, 0));
```
psavery: Same here.

avogadro/qtgui/rwmolecule.cpp
```diff
@@ -231,6 +234,7 @@ class RemoveAtomCommand : public RWMolecule::UndoCommand
     if (positions3d().size() == atomicNumbers().size())
       positions3d().resize(movedId, Vector3::Zero());
     atomicNumbers().resize(movedId, 0);
+    colors().resize(movedId, Vector3ub(0, 0, 0));
```
psavery: Same here.

frodofine:
Sorry for the multiple pushes, I made a mistake and needed to revert something and it got ugly.

I've broken this commit into three parts. The first changes the `Core::Molecule` and `Core::Atom` classes to have an additional array that holds the current color of a given atom. The second changes the rendering code so that this array is used to set a given color. Finally, the third modifies the SelectionTool Plugin to given the user the ability to change the color.

The current way to change an Atom's color is to select it with this tool. Then, the user clicks the push button in the lefthand tool to select a new color and if they select a new color and apply it, the scene is updated. Let me know if there is anything I should change here, I'm not quite sure this is the best way to make this interface.

cryos:
I am seeing some very strange behavior here when applying custom colors, need to debug it further. It works the first time, then after that it seems somewhat random what atoms change color.

frodofine:
I can look into that in a bit.

ghutchis:
Just for reference, this would partly fix #335

cryos:
@frodofine just checking into this, if you don't have time I can look into what is causing the issue. I think this is a great addition and would like to get it ready to be merged.

frodofine:
Unfortunately, I don't have much time now as I'm finishing up some papers before grant season starts. I'll have time again around the beginning of July at the earliest.

PR #378, opened 2018-08-30
OBFileFormat: Allow processing of extra args

psavery:

Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

### DIFF
--- a/avogadro/qtplugins/openbabel/obfileformat.cpp
+++ b/avogadro/qtplugins/openbabel/obfileformat.cpp
@@ -20,9 +20,13 @@
 
 #include <avogadro/io/cmlformat.h>
 
+#include <nlohmann/json.hpp>
+
 #include <QtCore/QCoreApplication>
 #include <QtCore/QFileInfo>
 #include <QtCore/QTimer>
+
+using json = nlohmann::json;
 
 namespace Avogadro {
 namespace QtPlugins {
@@ -90,6 +94,12 @@ OBFileFormat::~OBFileFormat()
 
 bool OBFileFormat::read(std::istream& in, Core::Molecule& molecule)
 {
+  json opts;
+  if (!options().empty())
+    opts = json::parse(options(), nullptr, false);
+  else
+    opts = json::object();
+
   // Allow blocking until the read is completed.
   OBProcess proc;
   ProcessListener listener;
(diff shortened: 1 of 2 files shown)


psavery:
I haven't tested this, but how does this look for processing extra args via Open Babel?

avogadro/qtplugins/openbabel/obfileformat.cpp
```diff
@@ -112,6 +122,15 @@ bool OBFileFormat::read(std::istream& in, Core::Molecule& molecule)
   if (formats2D.contains(QString::fromStdString(m_fileExtensions.front())))
     options << "--gen3d";
 
+  // Check if we have extra arguments for open babel
+  json extraArgs = opts.value("OBExtraArgs", json::object());
```
ghutchis: Why is this OB specific? Shouldn’t we have one key/value for arguments to pass to file formats, regardless of type? Even if the key name is documented, this seems un-necessarily opaque.

I’d prefer to have something like “formatArguments”

psavery: Well, so far in our usage of the format options, each format has it's own keys, and each key represents a particular option for that format. See [here](https://github.com/OpenChemistry/avogadrolibs/blob/master/avogadro/io/cjsonformat.cpp#L376) and [here](https://github.com/OpenChemistry/avogadrolibs/blob/master/avogadro/io/xyzformat.cpp#L154) for examples. 

So, for the xyz format, the `perceiveBonds` key has a boolean value (true or false depending on whether we should perceive bonds or not).

For Open Babel, I pictured the json options looking like this:

```
{
  "OBExtraArgs": [
    "-xw",
    "-xz"
  ]
}
```

The type of the value for the `OBExtraArgs` key would be an array of strings. Each string gets forwarded directly as a command line argument to `obabel` (The example above is an example for the POSCAR format. Passing both of these arguments will change the way the POSCAR output is ordered).

I named it `OBExtraArgs` because I don't think we could use the values for that key for any other format. They are command line arguments for Open Babel.

ghutchis: It might be fine to have particular keys for internal formats. But consider that there are two XYZ readers and two POSCAR readers now:
- Avogadro internal
- Open Babel

I understand the syntax for the JSON and that it’d be an array of strings. But picture it from a new script user - they want to write a command that uses OB’s FASTA support to generate DNA coordinates.
- How would they know to use OBExtraArgs
- Why is the key `OBExtraArgs` when for XYZ there’s a `perceiveBonds`

So my main point is that this should have a better name like `formatArguments`. I’m open to something better, but `OBExtraArgs` is IMHO a bit hacky.

cryos: Why not just use a key called arguments? At the end of the day whoever is writing this needs to know what the arguments are, but there is no need to encode ob into the key name. Where they can be standardized for internal formats they should be, I think we can safely use the arguments keyword for passing arguments to command line tools. If internal formats reuse that in the future they can do some validation/graceful handling.

psavery: Okay, sounds good! I amended the commit and changed the key to "arguments".

ghutchis:
Looks great, thanks!

ghutchis:
AppVeyor seems to be failing for everything - due to timeouts.

PR #387, opened 2018-09-15
Add a default Mac build to the Travis script.

ghutchis:

Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -51,6 +51,7 @@ matrix:
       env: TASKS="ctest clang-3.9"
 
     - os: osx
+      env: TASKS="Mac clang"
 
 install:
   - . ./scripts/travis/install.sh
--- a/.travis.yml
+++ b/.travis.yml
@@ -51,7 +51,6 @@ matrix:
       env: TASKS="ctest clang-3.9"
 
     - os: osx
-      env: MATRIX_EVAL="CC=clang && CXX=clang++"
 
 install:
   - . ./scripts/travis/install.sh
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,12 +7,16 @@ language: cpp
 sudo: true
 
 matrix:
+  allow_failures:
+    - os: osx
+
   include:
     - language: python
       sudo: false
       python: 2.7
       env: TASKS="clang-format"
 
+      os: linux
       dist: trusty
       addons:
         apt:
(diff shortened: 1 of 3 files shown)


ghutchis:
At the moment, the Mac build is set as an “allowed failure” but it’d be nice to have Mac packages uploaded via Travis.

ghutchis:
The Mac build now works, although it does not yet run `ctest`

scripts/travis/build.sh
```diff
@@ -29,7 +30,23 @@ else
     -DUSE_SYSTEM_PCRE=OFF \
     -DUSE_SYSTEM_ZLIB=ON \
     ..
-  make -j2
+  else
+    # osx
+    export CC=clang
+    export CXX=clang++
+    export CMAKE_PREFIX_PATH=/usr/local/Cellar/qt/5.11.1/lib/cmake
```
cryos: Does this work as expected? The prefix path would normally reference the install prefix, which I would expect to be /usr/local/Cellar/qt/5.11.1, the Qt5_DIR has to go deeper into lib usually. No objections if it works as expected.

.travis.yml
```diff
@@ -41,8 +47,11 @@ matrix:
         - MATRIX_EVAL="CC=gcc-6 && CXX=g++-6"
 
     - compiler: clang
+      os: linux
       env: TASKS="ctest clang-3.9"
 
+    - os: osx
+
```
cryos: Adding a TASKS="macOS C++" or similar would make the matrix easier to read.

scripts/travis/build.sh
ghutchis: Yup, it builds: 
https://travis-ci.org/OpenChemistry/avogadrolibs/jobs/429095892

.travis.yml
ghutchis: Sure.

scripts/travis/build.sh
ghutchis: Ideally, the Travis builds would be useful for AppImage and Mac continuous / nightly binary packages.

cryos: I saw it built, I just wondered if the variable was being used/affected it - poorly worded question. It isn't how it is intended to be used, and so I was surprised it would work, but CMake can surprise me at times.

PR #409, opened 2018-11-09
Fix signal jobUpdated

yurivict:
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

### DIFF
--- a/avogadro/molequeue/molequeuedialog.cpp
+++ b/avogadro/molequeue/molequeuedialog.cpp
@@ -122,11 +122,11 @@ MoleQueueDialog::SubmitStatus MoleQueueDialog::submitJob(
         progress.setCancelButton(nullptr);
 
         // Job completed -- overwrite job template with updated job details.
-        connect(&dlg.widget(), SIGNAL(jobUpdated(MoleQueue::JobObject)),
-                &dlg.widget(), SLOT(setJobTemplate(MoleQueue::JobObject)));
+        connect(&dlg.widget(), SIGNAL(jobUpdated(const ::MoleQueue::JobObject&)),
+                &dlg.widget(), SLOT(setJobTemplate(const ::MoleQueue::JobObject&)));
         QList<MetaMethod> lookupSignal;
         lookupSignal << MetaMethod(&dlg.widget(),
-                                   SIGNAL(jobUpdated(MoleQueue::JobObject)));
+                                   SIGNAL(jobUpdated(const ::MoleQueue::JobObject&)));
         dlg.widget().requestJobLookup();
         if (!dlg.waitForSignal(lookupSignal)) {
           progress.hide();


yurivict:
It breaks in the runtime. Signals and their receptors are expected to have the exact same signature, otherwise there are errors in the runtime.

cryos:
That is not the case though, please see [this article on signal-slot signature normalization](https://marcmutz.wordpress.com/effective-qt/prefer-to-use-normalised-signalslot-signatures/).

yurivict:
It doesn't work for whatever reason. Also, I work with Qt5 for years and never felt that Qt was flexible on signature mismatches.

Also, this article is from 2011. Things might have changed. It is this article from 2011 vs. the fact of runtime failures in Avogadro. Which one should be believed?

cryos:
Things haven't changed, I just selected a random article, normalized signatures were not abandoned. I will take a look, but when I can't reproduce an error I generally believe my own eyes. We use normalized signal-slot connections all over the place, and so would see many more failures than this. I have been developing in Qt since 2006 or so, across three major versions regularly on Windows, macOS, and Linux so feel fairly confident in my knowledge of the signal-slot mechanism, along with the new compile-time signatures introduced more recently.

yurivict:
Avogadro fails here without this patch.

There are no particular search hits for ```Qt normalized signatures``` in the last year.

Also there is now the new syntax of connecting signals: https://wiki.qt.io/New_Signal_Slot_Syntax

```SIGNAL``` and ```SLOT``` macros are now officially obsolete. Consequently, the signal matching rules are now different.

cryos:
> Avogadro fails here without this patch.

Where? I am not seeing a failure. Describe the steps I can use to reproduce this failure please.

> There are no particular search hits for Qt normalized signatures in the last year.

It is real, we use it extensively, see the [Qt documentation used by the code connecting signals and slots](http://doc.qt.io/qt-5/qmetaobject.html#normalizedSignature).

> Also there is now the new syntax of connecting signals: https://wiki.qt.io/New_Signal_Slot_Syntax

Like I said, we will move to this, I agree that it is an improvement, but that is not what you are doing here.

> SIGNAL and SLOT macros are now officially obsolete. Consequently, the signal matching rules are now different.

I searched your wiki page for where it stated they were obsolete, and I didn't see it. I agree that they are becoming obsolete, and I am motivated to move to compile time failure as that is much easier to detect early. The matching rules are different, but they are not what you are using, and your patch doesn't make sense to me. You can obviously patch these locally for BSD,

yurivict:
> Where? I am not seeing a failure. Describe the steps I can use to reproduce this failure please.

Go to ```Quantum```/```GAMESS``` and perform any computation.
Observe the error message printed to stdout that this signal doesn't exist in this class.

ghutchis:
I can’t reproduce the error on Mac. I’m willing to believe that for some compilers, the signal/slot signature is slightly different and would cause the problem.

I’d definitely be willing to merge this if you convert the patch to use the new PMF syntax - that would be great, e.g.using clazy:
https://github.com/KDE/clazy/blob/master/docs/checks/README-old-style-connect.md

That should prevent the problem from re-occurring since it’s a compile-time error, not a user-time error.

PR #419, opened 2019-01-03
Bring VTK volume rendering and surfaces to Avogadro

cryos:
This is a pretty large commit, I had the intention of breaking it up
more, but it is all highly related and hard to bring in piece-by-piece.
There are a few things to note, such as the fact that VTK now requires
an OpenGL 3.3 core profile, making it incompatible with the Avogadro
rendering. This means that we must defer all rendering to VTK in the VTK
render window, and copy our data into VTK data structures.

The vtkImageData format also uses a different ordering (often referred
to as "Fortran ordering", whereas our Cube class uses the "C ordering"
used in most places (C, C++, Python, etc). So cubes must be reordered
as they are copied over.

The color opacity widget is ported from Tomviz, with some adaptations to
only use VTK, and some work to create a color map that looks good for
typical molecular orbitals. This is needed to see the populations of
values in cubes, and modify the color opacity map for volume rendering.

The widget and the view both update as new cubes are calculated, and
trigger new renders. There is still some work to do, and the active
objects work needs the corresponding additions in the application to
ensure we can see what is currently active. It is very cool to see MOs
pop in, and to benefit from the speed of the flying edges algorithm.
This also reuses cubes, which the standard approach should do too.

### DIFF
--- a/avogadro/qtopengl/activeobjects.cpp
+++ b/avogadro/qtopengl/activeobjects.cpp
@@ -1,11 +1,11 @@
-/******************************************************************************
-  This source file is part of the Avogadro project.
-  This source code is released under the 3-Clause BSD License, (see "LICENSE").
-******************************************************************************/
+/* This source file is part of the Avogadro project.
+   It is released under the 3-Clause BSD License, see "LICENSE". */
 
 #include "activeobjects.h"
 
 #include "glwidget.h"
+
+#include <avogadro/qtgui/molecule.h>
 
 namespace Avogadro {
 namespace QtOpenGL {
@@ -24,9 +24,39 @@ GLWidget* ActiveObjects::activeGLWidget() const
   return m_glWidget;
 }
 
+QWidget* ActiveObjects::activeWidget() const
+{
+  return m_widget;
+}
+
+QtGui::Molecule* ActiveObjects::activeMolecule() const
+{
+  return m_molecule;
+}
+
 void ActiveObjects::setActiveGLWidget(GLWidget* glWidget)
 {
-  m_glWidget = glWidget;
+  if (m_glWidget != glWidget) {
+    m_glWidget = glWidget;
+    emit activeGLWidgetChanged(m_glWidget);
+    setActiveWidget(glWidget);
+  }
+}
+
+void ActiveObjects::setActiveWidget(QWidget* widget)
+{
+  if (m_widget != widget) {
+    m_widget = widget;
+    emit activeWidgetChanged(widget);
+  }
+}
+
+void ActiveObjects::setActiveMolecule(QtGui::Molecule* molecule)
+{
+  if (m_molecule != molecule) {
+    m_molecule = molecule;
+    emit activeMoleculeChanged(molecule);
+  }
 }
 
 } // namespace QtOpenGL
--- a/avogadro/qtopengl/activeobjects.cpp
+++ b/avogadro/qtopengl/activeobjects.cpp
@@ -37,6 +37,7 @@ QtGui::Molecule* ActiveObjects::activeMolecule() const
 void ActiveObjects::setActiveGLWidget(GLWidget* glWidget)
 {
   if (m_glWidget != glWidget) {
+    m_widget = nullptr;
     m_glWidget = glWidget;
     emit activeGLWidgetChanged(m_glWidget);
     setActiveWidget(glWidget);
@@ -46,6 +47,7 @@ void ActiveObjects::setActiveGLWidget(GLWidget* glWidget)
 void ActiveObjects::setActiveWidget(QWidget* widget)
 {
   if (m_widget != widget) {
+    m_glWidget = nullptr;
     m_widget = widget;
     emit activeWidgetChanged(widget);
   }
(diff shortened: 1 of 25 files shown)


psavery:
@cryos I tried this out. It looks really nice to do the volume rendering!

I do get a seg fault when I try to do these steps, though (backtrace attached):

1. Open Avogadro2.
2. Click "close" next to "split horizontal" and "split vertical" buttons to close the Avogadro2 render window.
3. Click "VTK" to open a VTK render window.
4. Open up cn9-homo.cube (from avogadrodata) with the Avogadro: Gaussian Cube reader (the volume rendering then appears and looks very nice).
5. Click "Extensions" -> "Edit Color Opacity Map..." (crash).

[backtrace.txt](https://github.com/OpenChemistry/avogadrolibs/files/2725253/backtrace.txt)

cryos:
If you give it another try you should see it works much better now. The update was more aggressive than I had intended, it also broke for other files if you created cubes, with a VTK window open, but had not opened the edit dialog. That data looks nice, I am hoping to put a blog post together soon and will use that as one example. Thanks for taking a look @psavery

psavery:
Unfortunately, it looks like the problem I was having might be deeper in the code. When I try to open up new views (either "3D View" or "VTK"), they tend to have some issues (like my draw tool doesn't work in new 3D views), and I still get the crashing issue above...

Is there a way to start Avogadro2 with a VTK view instead of with a 3D view?

cryos:
I think you may be confusing existing bugs with new ones too, I was just looking at draw tool doesn't work in new view split, and it is because there is no active molecule. It has been that way for a while, and for me selecting an active molecule fixes the issue. This should be improved, but not linked to the addition of volume rendering. It is a general issue I had noted with new view creation - no active molecule until one is chosen feels confusing.

psavery:
@cryos Ah yes, I see. That is indeed a separate issue. I didn't realize a molecule needed to be selected after opening up a new "3D View".

I wanted to try out the color opacity map editor, though, and I still get the crash mentioned when following the steps above. Attached is the new backtrace. I have updated both my avogadrolibs and openchemistry repos to the latest.

[new-backtrace.txt](https://github.com/OpenChemistry/avogadrolibs/files/2734182/new-backtrace.txt)

You can also see from the attached image where I tried to open cn9-homo.cube (from avogadrodata) and create the surfaces in a regular "3D View" (left) and a "VTK" view (right). The volume rendering does look very nice! But my color opacity map doesn't seem to be populated with data. Perhaps I'm using it wrong?

img

cryos:
I don't know what you are doing. I have tested this on my machine (Arch Linux), and two macOS machines (Mojave and High Sierra I think). I made sure that the histogram updates, I was wondering if it was because you opened the editor before loading the cube, but that works fine here. I am going to guess there is a sequence you are hitting I didn't account for, did you try opening the cube file, then making sure it is active in the view before opening the opacity editor?

psavery:
Yeah, I just checked that it is open. I have tried several different sequences of opening the file and then the color opacity map editor, but I seem to always get the same result (the blank color opacity map editor). I just re-cloned and re-compiled from scratch, making sure all the versions are right, but I still seem to have the issue...

It might just be an issue specific to my computer (gcc and Qt version posted below)... I'll try it out on my Mac within the next few days.

Here is an example of steps for me:

1. Run `rm ~/.config/OpenChemistry` to remove configurations.
2. Click `Open` and navigate to `cn9-homo.cube`. Open it and choose `Avogadro: GaussianCube`.
3. Click `Split Horizontal`.
4. Click `VTK` to open a new VTK render window.
5. In navigation mode (default tool), click on the blank VTK render window to select it.
6. Then, click on the `Cube file generated by...` under the `Molecules` section to make `cn9-homo.cube` active in the VTK render window. The volume rendering will appear.
7. Click `Extensions`->`Edit Color Opacity Map...`
8. The color opacity map editor appears to be blank. Image pasted below.

image

My arch linux computer was updated today. Here are the software specs:

gcc (GCC) 8.2.1 20181127
Qt 5.12.0
cmake 3.13.2

openchemistry sha: 04e44b9fa75037102cbb28c47753d6a2b9aa7943
avogadrolibs sha: a15929c3da1425810964c655626aa0b26fb53951
avogadroapp sha: f9d93869f710c08c24151136af8c40eae5b47491
VTK sha: ad0b261f7b761c9de9eba615d8af99e8b6c9bb2a

I use xfce4 for the desktop environment and window manager.

Graphics card: Nvidia GeForce GTX 950M

cryos:
It is simpler than that, I was staging this pull request, but there is a small avogadroapp PR that I hadn't created. I think if you checkout https://github.com/OpenChemistry/avogadroapp/pull/102 then you should see things work better. I want to get to a place where active object tracking is handled better to avoid setting things on every loaded module every time something changes - the vtkGLWidget is needed so that it can find the color map, image data, etc.

psavery:
I tried it out with that branch of avogadroapp, and it works really well now! It is really cool! Thanks so much for working on this! You can see my CO below.

co

I think it'd be nice if the "isovalue" and "opacity" were disabled unless "Enable surface rendering" is checked, but I can submit a PR for that later...

Looks good to me!

cryos:
I think it is time this got merged, and we can obviously iterate on this in master. I fixed the clang-format warnings, and will let the CI build to confirm all still looks OK before merging.

PR #424, opened 2019-09-23
Interface with MongoChem server

psavery:
This adds a basic interface with MongoChem server. It is located at "Extensions"->"MongoChem". Currently, it can find public molecules in the database as shown below.

example

There are several hard-coded parts, including the server address to "localhost:8080". But these can be fixed in the future. Additionally, a QTableWidget is used rather than a model/view approach, which is something that should probably be fixed in the future as well...

@cryos


ghutchis:
While it's not perfect by any means, you might also want to look at the PQR plugin:
https://github.com/OpenChemistry/avogadrolibs/blob/master/avogadro/qtplugins/importpqr/pqrwidget.cpp

I don't know if there's merit to integrating some of these efforts. One case we considered was "look up the thing I just drew" in the server (i.e., exact InChI match).

cryos:
I was just looking at that @ghutchis. This plugin is intended to start with non-authenticated, and then add editing/upload capabilities. I think we should iterate towards some common dialogs for search, preview, etc. I think short term it would be good to experiment in a dedicated plugin before we try and combine things. I would also like to improve the search of other online databases where error reporting is minimal such as PDB, and the structure resolver service.

avogadro/qtplugins/mongochem/mongochemwidget.ui
```diff
@@ -0,0 +1,59 @@
...
+    <width>620</width>
+    <height>410</height>
+   </size>
+  </property>
+  <property name="windowTitle">
+   <string>Form</string>
+  </property>
+  <layout class="QGridLayout" name="gridLayout">
+   <item row="0" column="0">
+    <widget class="QTableWidget" name="table_molecules">
```
cryos: Prefer camel case for names, i.e. tableMolecules

avogadro/qtplugins/mongochem/mongochemwidget.ui
```diff
@@ -0,0 +1,59 @@
...
+     </column>
+     <column>
+      <property name="text">
+       <string>InChIKey</string>
+      </property>
+     </column>
+    </widget>
+   </item>
+   <item row="1" column="0">
+    <widget class="QPushButton" name="push_search">
```
cryos: As above

psavery:
There's now a config button in the bottom right corner that opens a dialog where the user can set the url and the api key. The line edit for the api key uses a password echo mode. It is then stored as a plain text variable in the program.

The api key is currently not being used for anything. But soon, after the user accepts the config, the api key will be used for authentication, and the resulting token will be stored and used for the http requests.

update

config

psavery:
Molecules can now be downloaded by selecting a row and clicking "Download" at the bottom.

img

cryos:
I think this looks close, when entering the server address we should append the /api/v1 for the user, so they can just paste in the URL, i.e. http://localhost:8080 or https://data.openchemistry.org, it might be good to have a little debug code in to tell the user we are polling for a result on the command line. If we can make it remember the shorter URL and the API key then I  think this is in a place where  it should be merged.

There is a lot of code to submit, the fact it works is great and we can consider possible API updates later.

psavery:
@cryos Thank you! And sounds good, I will work on those.

After the geometry branches in mongochemserver get merged, I'm planning to have Avogadro2 always upload its molecule as a geometry for the calculation, and then run the calculation on that geometry. That way, it will always perform the calculation on whatever is drawn (rather than what is already in the database).

psavery:
Done and done.

cryos:
Let's add the QSettings save for URL and API key (I don't see that), and then merge what we have. Agreed on the geometry, but what we have here is already useful. There are quite a few other things such as list of outstanding jobs, persisting across open/close of laptop/reboot for polling/monitoring, etc. I think we should separate some of these additional things into future PRs if we have time.

psavery:
Done! Thanks Marcus!

cryos:
There are a few minor failures due to clang-format issues that crept in, once fixed I will merge this in. Thanks Patrick.

psavery:
Fixed.

PR #425, opened 2019-09-30
Area selection

cryos:
This now works for area selection, with some keyboard modifiers for growing a selection and toggling it. Some more improvements can be made to look and feel, but I think this will be a big improvement for most people.


cryos:
I added support for transparent selection spheres now, so:
avo-sel1
avosel2

ghutchis:
@cryos - looks like clang-format is needed.

cryos:
Fixed now, thanks @ghutchis

PR #440, opened 2020-02-12
add support for a JSON list of selected atoms

ghutchis:
Based on discussions:
https://discuss.avogadro.cc/t/command-scripts-json-documentation/3067

Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

### DIFF
--- a/avogadro/qtgui/interfacescript.cpp
+++ b/avogadro/qtgui/interfacescript.cpp
@@ -458,6 +458,14 @@ bool InterfaceScript::insertMolecule(QJsonObject& json,
   if (m_moleculeExtension == QLatin1String("None"))
     return true;
 
+  // insert the selected atoms
+  QJsonArray selectedList;
+  for (Index i = 0; i < mol.atomCount(); ++i) {
+    if (mol.atomSelected(i))
+      selectedList.append( (qint64)i );
+  }
+  json.insert("selectedatoms", selectedList);
+
   Io::FileFormatManager& formats = Io::FileFormatManager::instance();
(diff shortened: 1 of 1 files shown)


avogadro/qtgui/interfacescript.cpp
```diff
@@ -458,6 +458,14 @@ bool InterfaceScript::insertMolecule(QJsonObject& json,
   if (m_moleculeExtension == QLatin1String("None"))
     return true;
 
+  // insert the selected atoms
+  QJsonArray selectedList;
+  for (Index i = 0; i < mol.atomCount(); ++i) {
+    if (mol.atomSelected(i))
+      selectedList.append((qint64)i);
```
cryos: Can we avoid C-style casts, prefer `static_cast<qint64>(i)`, one question - the CJSON already has a selected atom array, but this is to support atom selections augmenting other formats?

cryos: I think from reading the related discussion this is indeed a request to have access to selection in JSON for any format, for CJSON it seems duplicative but I can see why implementations might need another format for convenience, but still want selection lists.

ghutchis: Happy to fix the cast.

Yes, the discussion made sense - for a variety of codes/scripts they would want SDF, XYZ, PDB .. some other format .. but still want selected atoms, constraints, etc.

This seems like a reasonable workaround for a clear need from the scripts.

cryos (approved):
Sounds good, thanks.

Issue #221, opened 2018-03-15
"Reset View" Command


ghutchis:
Avogadro v1 had a feature to reset the view - computing the best-fit plane through the molecule. This also included an animation as the view twisted into position and zoom.

Another useful feature is the Reset View to Axes - rather than calculate the best-fit plane, the view resets to display the XY plane.

ghutchis:
I have an initial template here, but don't have time to finish it:
https://gist.github.com/ghutchis/d53469f38887074eb63017bdaf590c0a

serk12:
Hi, I'm almost done with this issue but I'm facing some problems with the animation. 
I'm using a QTimer to update the camera view model but it doesn't update the scene, any solution?

cryos:
You might have to show code, but only the camera should need to be changed. I am not clear on what you mean by it doesn't update the scene. Maybe you need a call to render to actually show the updated camera position?

serk12:
Ups sorry. Yes, here is the code https://github.com/serk12/avogadrolibs/blob/issue-221/avogadro/qtplugins/resetview/resetview.cpp
To be more exactly the problem is here:
`  QTimer::singleShot(alpha*1000, this, [this, interpolation] (){ m_camera->setModelView(interpolation); });`
My theory is that the Action threads dies before this happen and it doesn't call the update render, but I don't know how to call it.

ghutchis:
@serk12 - sorry I missed your comment - at the first step, let's skip the animation and make sure the camera is set correctly and re-renders. If so, please submit a PR - the animation is secondary.

When we implemented this in Avo1, we did it the opposite way:
https://github.com/cryos/avogadro/blob/79d22168cb8c5a874189bebf16f7270871f3c469/avogadro/src/mainwindow.cpp#L2753

In that code, the class kept a permanent QTimer that would periodically call the interpolation method - and in turn update the camera.

As far as needing to call for an explicit re-render,  I'm not sure. The code in `avogadro/qtplugins/navigator/navigator.cpp` doesn't do that, but that's called through Qt events so it's inherently part of the render loop anyway.

ghutchis:
Thanks to @serk12 - this is now merged.

PR #450, opened 2020-03-25
Reset view (camera)

serk12:
Hi, following the recommendations and tips from @ghutchis I did a first approach to the reset View and Centering functions in the issue #221 .

- Centering: It calculates a oriented bounding box and selects the best vectors to get the most relevant plane, next is displaced at the centroid and it calculates how far should be.

- Reset view: the same as centering but, it uses a simple bounding box to get the centroid and the default axis.


ghutchis:
one minor thing is the sign-off on the commits:
https://github.com/OpenChemistry/avogadrolibs/pull/450/checks?check_run_id=533595913

ghutchis:
I'll take a look at this later this afternoon (Pittsburgh time)

ghutchis:
There's definitely a redraw needed.

I think you need to `override`   virtual void setActiveWidget(QWidget* widget);
Then you should be able to properly call GLWidget::requestUpdate()
https://github.com/OpenChemistry/avogadrolibs/blob/master/avogadro/qtopengl/glwidget.h#L171

This is basically what the `navigate` tool does:
```
emit updateRequested();
```

ghutchis:
You should also run `clang-format` to keep the formatting consistent.

ghutchis:
This is actually a very nice solution - using signals - but  I'm not seeing the redraw. Does it work for you?

cryos:
I added the signal, and it is connected for all tools in `qtopengl/glwidget.cpp`, I haven't had a chance to build this yet. I did want to say at a high level there are lots of tiny commits, and it would be good to rewrite this branch to squash most (all) of them together, and write a more concise commit message. The whole point of the Developer Certificate or Origin is that your real legal name must be used, see [here](https://blog.chef.io/introducing-developer-certificate-of-origin/) for an example of the reasoning, so the commits do not comply with DCO unless they have real names - we need to know who is contributing code, and get their sign off under their name.

avogadro/qtplugins/resetview/resetview.cpp
```diff
@@ -0,0 +1,225 @@
...
+                         m_camera->projectionType());
+
+  Vector3f eye = -mid + (Vector3f::UnitZ() * -1.0f * d);
+  goal->translate(eye);
+  animationCamera(goal, animate);
+}
+
+void ResetView::animationCamera(Eigen::Affine3f* goal, bool animate)
+{
+  if (animate) {
```
cryos: I think you should simplify this for now, rather than leave an empty `if` block, looks confusing.

avogadro/qtplugins/resetview/resetview.cpp
```diff
@@ -0,0 +1,225 @@
...
+  Vector3f mid = (max.cast<float>() + min.cast<float>()) / 2.0f;
+  float d = getZDistance(m_camera->projection(), max.x() - min.x(),
+                         m_camera->projectionType());
+
+  Vector3f eye = -mid + (Vector3f::UnitZ() * -1.0f * d);
+  goal->translate(eye);
+  animationCamera(goal, animate);
+}
+
+void ResetView::animationCamera(Eigen::Affine3f* goal, bool animate)
```
cryos: Why not pass by const reference here?

avogadro/qtplugins/resetview/resetview.cpp
```diff
@@ -0,0 +1,225 @@
...
+  Vector3 min, max;
+
+  getBB(mols, min, max);
+
+  Eigen::Matrix3f linearGoal;
+  linearGoal.row(0) = Vector3f::UnitX();
+  linearGoal.row(1) = Vector3f::UnitY();
+  linearGoal.row(2) = Vector3f::UnitZ();
+  // calculate the translation matrix
+  Eigen::Affine3f* goal = new Eigen::Affine3f(linearGoal);
```
cryos: Why `new`, where is this deleted?

cryos:
In my testing locally the render happened as expected. I am not sure about using "Center" as it wasn't obvious - we can obviously tweak that. I can try and get it build on a macOS machine, but I take it the navigate tool works @ghutchis but you aren't getting the render here?

ghutchis:
@cryos - I was referring to using the signal / slot - I would have just called ::update() manually.

It doesn't redraw for me. The signal is emitted, but after adding some debugging, it's never received by the GLWidget.

As best as I can figure out, setActiveWidget() is never called. I added some debugging to that method and never see the message.
```
void ResetView::setActiveWidget(QWidget* widget)
{
  m_glWidget = widget;
  qDebug() << " set widget: " << widget->metaObject()->className();
  connect(this, SIGNAL(updateRequested()), m_glWidget, SLOT(requestUpdate()));
}
```

avogadro/qtplugins/resetview/resetview.cpp
serk12: Yes, you are right, I let it because I still have hope to do the animation but yes, for the PR is annoying.

avogadro/qtplugins/resetview/resetview.cpp
serk12: There is no need, true. This and the previous have no sense since I changed how the default camera is created

serk12:
Okey, first of all thanks for the feedback, I'm still a bit rocky in PR.

Also thanks, for the suggestion about how to handle this many little commits. Next time I will try to do less commits or a develop branch for myself an another for the PR and sign-off.

About the repaint, for me it works since the beginning, but only does the repaint at the end of the action (I'm in Linux so maybe there is something that it's not working in macOS), and because this the animation is not working and this make sense because `setActiveWidget` is never called. So it needs to be call, but I don't know if this is what you already did in 
> I added the signal, and it is connected for all tools in qtopengl/glwidget.cpp

or more or less where to do it (in the factory? the `MainWindow`? in the `ResetView` makes no sense)...

ghutchis:
Don't worry too much about the small commits - it's also possible to 'squash and merge' in the PR.

I guess my question for @serk12 is whether setActiveWidget() is actually called for you on Linux? If so, I'll check into what's different on Mac.

cryos:
@ghutchis this just works for me on Linux, I haven't dug into it much yet. I am guessing something isn't quite right on macOS, but surprised as I don't see any Linux versus macOS differences jumping out at me. I added the signal so that tools didn't have to get the widget, connect to it etc. This should all be happening magically for all tools - I can look into it further but the design is that the tool developer never connects any signal-slot pairs in the tool.

ghutchis:
@cryos - this code isn't a Tool it's an Extension class. Tools work fine (Navigator) - but I don't see any code hooking up Extensions. The code in this PR should work okay if something calls setActiveWidget to this class, but that doesn't seem to be happening. If I look at the codebase, I'm not sure what would call setActiveWidget for each Extension.

cryos:
It is rendering as expected no Linux for me. I can try and take a better look tomorrow, I had missed that on the first look.

serk12:
I couldn't find the Linux vs Mac repaint problem, but I found that there is no interaction between `MultiViewWidget::activeWidgetChanged` and `ExtensionPlugin::setActiveWidget`. So the simplest solution would be to add in `MainWindow::viewActivated` this:
`foreach (ExtensionPlugin* extension, m_extensions) {
      extension->setScene(&glWidget->renderer().scene());
      extension->setCamera(&glWidget->renderer().camera());
      extension->setActiveWidget(widget);//<-this line
    }`
With this the repaint works and also the animation, (and probably the mac version, due quarantine I have no Mac with me right now), I can push the animation version with some fixes I found if you want

ghutchis:
Yes, that's exactly the fix needed for `avogadroapp` - I'll actually go submit a PR for that.

@cryos - given that fix - functional testing is fine. I'll look at the code itself later today.

ghutchis:
Wow, that was weird - don't know why merging the `avogadroapp` PR closed this one -sorry.

ghutchis:
The animation component definitely works on my Mac now. If anything it's a bit too fast. 😄

ghutchis:
This looks good to me.  I'm going to merge.

PR #458, opened 2020-04-24
Mutex

cryos:
Add shared mutexes, move to C++17.


ghutchis:
The Travis build is a clang-format issue. The Ubuntu build seems to be an issue linking with pthread.

ghutchis:
Looks like circleci has an issue with C++17:
> Target "avogadroiopython" requires the language dialect "CXX17" , but CMake
>   does not know the compile flags to use to enable it.

cryos:
We need to look at CircleCI as it didn't package the wheel for the last release either. I think it need s few upgrades including CMake to a version that knows about C++17. Hopefully we can still make a binary wheel, if I need to for now we could look at a build flag to disable C++17 for the wheel which only packages core and io right now.

ghutchis:
I've had fairly good luck with the GitHub workflows, e.g. https://help.github.com/en/actions/language-and-framework-guides/using-python-with-github-actions

I'd be happy to help work towards a release workflow that auto-generates everything from a tag.

ghutchis:
Happy to merge this once the builds are done.

cryos:
This looks good - all builds are coming back clean. I will merge now.

PR #463, opened 2020-04-27
Get rid of bits/stdc++ headers - non-portable GCC header

ghutchis:
See https://stackoverflow.com/a/31816096/131896

Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

### DIFF
--- a/avogadro/qtplugins/coloropacitymap/vtkChartHistogram.h
+++ b/avogadro/qtplugins/coloropacitymap/vtkChartHistogram.h
@@ -4,8 +4,6 @@
 #ifndef tomvizvtkChartHistogram_h
 #define tomvizvtkChartHistogram_h
 
-#include <bits/stdc++.h>
-
 #include <vtkChartXY.h>
 
 #include <vtkNew.h>
--- a/avogadro/qtplugins/coloropacitymap/vtkChartHistogramColorOpacityEditor.cpp
+++ b/avogadro/qtplugins/coloropacitymap/vtkChartHistogramColorOpacityEditor.cpp
@@ -3,8 +3,6 @@
 
 #include "vtkChartHistogramColorOpacityEditor.h"
 
-#include <bits/stdc++.h>
-
 #include <vtkAxis.h>
 #include <vtkChart.h>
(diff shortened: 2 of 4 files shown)


cryos:
This fails for me locally when turning C++17 flag on, which is why I added it in the first place. Looking on the web the documentation is inconsistent but it definitely doesn't compile here with your branch and C++17 enabled:
```
In file included from /home/marcus/build/openchemistry/prefix/include/vtk-8.2/vtkRect.h:32,
                 from /home/marcus/build/openchemistry/prefix/include/vtk-8.2/vtkAxis.h:58,
                 from /home/marcus/src/openchemistry/avogadrolibs/avogadro/vtk/vtkplot.cpp:20:
/home/marcus/build/openchemistry/prefix/include/vtk-8.2/vtkMath.h: In static member function ‘static T vtkMath::ClampValue(const T&, const T&, const T&)’:
/home/marcus/build/openchemistry/prefix/include/vtk-8.2/vtkMath.h:1516:15: error: ‘clamp’ is not a member of ‘std’
 1516 |   return std::clamp(value, min, max);
      |               ^~~~~
ninja: build stopped: subcommand failed.
```
This is GCC 9.3.0, while Ubuntu is much older.

cryos:
OK, I chose poorly - if we have a recent compiler that complies with C++17 properly then std::clamp comes from the STL `algorithm` header, I should have added that. C++ master also fixed that issue. Easiest would be to put algorithm in the deleted spots, harder would be update but likely worth doing.

ghutchis:
I think it's building from VTK 9 right now from the `openchemistry` repo, which is why the automatic tests passed.

If you want to add <algorithm> that's fine with me.

cryos (approved):
Everything build and works if I move the VTK SHA forward - I am going to do that as I was a little sad about patching our code anyway...

cryos:
It is odd that it built, but I just realized the CI is not building/enabling VTK so that makes sense.

ghutchis:
@cryos - didn't realize that - i'll move the CI builds to enable more features for testing

Issue #436, opened 2019-12-31
avogadrolibs-1.92.1 fails to build: undefined reference to `gwavi_open'


pacho2:
**Avogadro version: (please complete the following information from the About box):**
 - Avogadrolibs: 1.92.1
 - Qt: 5.13.2

**Desktop version: (please complete the following information):**
 - OS: Gentoo Linux
 - Compiler : gcc-9.2.0

Fails to build with:
```
[ 77%] Building CXX object avogadro/qtplugins/surfaces/CMakeFiles/Surfaces.dir/SurfacesPlugin.cpp.o
[...]

PR #472, opened 2020-05-02
Missing plugin links

cryos:
This fixes the shared plugin build for me locally.


cryos:
This should fix #436 along with a few I didn't see bug reports for.

Issue #317, opened 2018-07-18
RPMLint warning: /usr/lib/libAvogadroQtPlugins.so.1.91.0 calls exit(), which is strongly discouraged


fusion809:
**Avogadro version: (please complete the following information from the About box):**
 - Avogadrolibs: latest commit, 1560 (02eeab3).
 - Qt: 5.11.1.

**Desktop version: (please complete the following information):**
 - OS: openSUSE Tumbleweed.
 - Version: 20180714.
 - Compiler: gcc 8.1.1. 

**Describe the bug**
I'm not much of a programmer, but the openSUSE Build Service's RPMLint program reports the warning (as opposed to error):

```bash
[  327s] avogadro2-libs.x86_64: W: shared-lib-calls-exit /usr/lib/libAvogadroQtPlugins.so.1.91.0 exit@GLIBC_2.2.5
[  327s] This library package calls exit() or _exit(), probably in a non-fork()
[  327s] context. Doing so from a library is strongly discouraged - when a library
[  327s] function calls exit(), it prevents the calling program from handling the
[  327s] error, reporting it to the user, closing files properly, and cleaning up any
[  327s] state that the program has. It is preferred for the library to return an
[  327s] actual error code and let the calling program decide how to handle the
[  327s] situation.
[  327s] 
```

so from what I understand of this warning /usr/lib/libAvogadroQtPlugins.so.1.91.0 calling exit() makes handling errors or bugs more challenging. This isn't a bug, _per se_, it's merely a warning that if rectified might make your jobs easier and aid future bug reporters in describing exactly what is wrong. 

**To Reproduce**
1. Clone my OBS repository: https://build.opensuse.org/package/show/home:fusion809/avogadro2-libs with (assuming `osc` is installed):

```bash
osc co home:fusion809 avogadro2-libs
```

2. Run:

```bash
osc build openSUSE_Tumbleweed x86_64 --noverify
```

3. Look near the end of the log and lo and behold you should see this warning. 

**Additional context**
My full build log, although I don't think it'll be helpful here, is [here](https://gist.github.com/91514232d199d4c88e9aa674f01bca46). 

Thanks for your time and I hope this helps advance this fine project,
Brenton

ghutchis:
This has now been fixed in #475

PR #475, opened 2020-05-08
Remove exit calls from qtaim code

ghutchis:
Not always great options, but these are error paths

Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

### DIFF
--- a/avogadro/qtplugins/qtaim/qtaimodeintegrator.cpp
+++ b/avogadro/qtplugins/qtaim/qtaimodeintegrator.cpp
@@ -129,7 +129,6 @@ QVector3D QTAIMODEIntegrator::integrate(QVector3D x0y0z0)
       break;
     default:
       qDebug() << "Catastrophic: No ODE parameters for this property.";
-      exit(1);
       break;
   }
 
@@ -257,7 +256,6 @@ void QTAIMODEIntegrator::r8_f(qreal t, qreal y[], qreal yp[])
       break;
     default:
       qDebug() << "Catastrophic: No ODE parameters for this property.";
-      exit(1);
       break;
   }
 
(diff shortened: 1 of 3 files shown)


ghutchis:
Fixes #317

avogadro/qtplugins/qtaim/qtaimodeintegrator.cpp
```diff
@@ -796,10 +792,10 @@ qint64 QTAIMODEIntegrator::r8_rkf45(qint64 neqn, qreal y[], qreal yp[],
       } else if (kflag == 4) {
         nfe = 0;
       } else if (kflag == 5 && abserr == 0.0) {
-        exit(1);
+        // error
```
cryos, line 795: Is is safe to do this? The code is pretty unfriendly in general, so I guess just marking it as error is OK. It would be good to rewrite some of this, but it is a decent undertaking. I don't object, as I think all of this code was pretty fragile to begin with, and was never adapted to modern C++/our general code style.

ghutchis: What I'm hoping, is that the Python implementations in `cclib` and `Horton` will be able to do all of this externally, and we can jettison the C++ code in favor of 'install the Horton Python plugin scripts to run QTAIM analysis'.

cryos: I didn't realize there were new implementations that might become available, that sounds preferable for sure! I will pay less attention to this, happy to rubber stamp it, I think this code has enough other issues I am concerned about that this shouldn't add any additional concern :-)

PR #486, opened 2020-08-30
Fixes #483 by forcing a C locale when building the cml file

### LINKED ISSUE #483: Wrong coordinates when using OpenBabel exporter
**Avogadro version: (please complete the following information from the About box):**
 - Avogadrolibs: 1.93.0-28-g780d634
 - Qt: 5.12.18

**Desktop version: (please complete the following information):**
 - OS: Ubuntu 20.04
 - Version: kernel 5.4.0-40-generic, libc6 2.31-0ubuntu9
 - Compiler: gcc/cpp (Ubuntu 9.3.0-10ubuntu2) 9.3.0

**Describe the bug**
When exporting a molecule using File->Export->Molecule..., the coordinates are completely messed up. 
In order to export the molecule, I had to first click on Save in the file selector, then click Cancel, because the file selector was shown again (but I was a bug report for it).
**To Reproduce**
1. Draw a molecule (eg: methane)
1. Go to File->Export->Molecule...
2. Select a file name, use mol2 extension, or xyz but choose OpenBabel exporter; then click on "Save", the file chooser will show again (which is another bug, but already reported), then click "Cancel"
3. Open file in Avogadro2 - or in another editor - the coordinates will be wrong

**Expected behavior**
The coordinates should match the ones which were drawn. But they are totally wrong.

**Screenshots**
Molecule as drawn
Screenshot_20200711_124209
Molecule loaded from exported file
Screenshot_20200711_124312

**Additional context**
The content of the mol2 file
```
@<TRIPOS>MOLECULE
*****
 6 5 0 0 0
SMALL
GASTEIGER

@<TRIPOS>ATOM
      1 C          -3.0000    1.0000   -3.0000 C.3     1  UNL1        0.0330
      2 H          -3.0000    2.0000   -0.0000 H       1  UNL1        0.0521
      3 H          -3.0000    1.0000    1.0000 H       1  UNL1        0.0521
      4 H          -3.0000    0.0000   -0.0000 H       1  UNL1        0.0521
      5 O          -2.0000    1.0000   -0.0000 O.3     1  UNL1       -0.3982
      6 H          -2.0000    2.0000    0.0000 H       1  UNL1        0.2090
@<TRIPOS>BOND
     1     1     2    1
     2     1     3    1
     3     1     4    1
     4     1     5    1
     5     5     6    1
```

luancarvalhomartins:
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

### DIFF
--- a/avogadro/io/cmlformat.cpp
+++ b/avogadro/io/cmlformat.cpp
@@ -13,7 +13,6 @@
   limitations under the License.
 
 ******************************************************************************/
-
 #include "cmlformat.h"
 
 #ifdef AVO_USE_HDF5
@@ -32,6 +31,7 @@
 #include <bitset>
 #include <cmath>
 #include <map>
+#include <clocale>
 #include <sstream>
 #include <streambuf>
 
@@ -431,8 +431,9 @@ bool CmlFormat::read(std::istream& file, Core::Molecule& mol)
 
 bool CmlFormat::write(std::ostream& out, const Core::Molecule& mol)
 {
+  char* old_locale = std::setlocale(LC_NUMERIC, NULL);
+  std::setlocale(LC_NUMERIC, "C");
   xml_document document;
-
   // Add a custom declaration node.
   xml_node declaration = document.prepend_child(pugi::node_declaration);
   declaration.append_attribute("version") = "1.0";
@@ -623,7 +624,7 @@ bool CmlFormat::write(std::ostream& out, const Core::Molecule& mol)
 #endif
 
   document.save(out, "  ");
-
+  std::setlocale(LC_NUMERIC, old_locale);
   return true;
 }
 


luancarvalhomartins:
Sorry, I inadvertently closed the issue. The PR is quite trivial.

avogadro/io/cmlformat.cpp
```diff
@@ -431,8 +431,9 @@ bool CmlFormat::read(std::istream& file, Core::Molecule& mol)
 
 bool CmlFormat::write(std::ostream& out, const Core::Molecule& mol)
 {
+  char* old_locale = std::setlocale(LC_NUMERIC, NULL);
+  std::setlocale(LC_NUMERIC, "C");
```
cryos, line 435: This is global, and Avogadro is multi-threaded. I was hoping that we could use imbue to set the locale on the stream for example, see https://stdcxx.apache.org/doc/stdlibug/24-3.html and https://pugixml.org/docs/manual.html for some background. It may turn out that C functions are used in places and there is no choice, it would be good to explore.

ghutchis:
This makes a lot more sense (e.g., why it worked for me, but not many others) - do we also want to set the locale on read, or is that un-necessary?

cryos:
The bigger problem is that as far as I can tell, and searching on this, setlocale is per process, and so changes the C locale for all of Avogadro, worse still the call is not thread safe, and our reading/writing is multithreaded. Ideally we want to use libraries that do not use the C locale, and instead imbue streams with the correct locale. This will fix the immediate issue, it is likely to introduce other issues. How bad is it as an application to just set `LC_NUMERIC` to C at startup, and our visual elements hopefully do the right thing but possibly do not...

ghutchis:
@luancarvalhomartins - apologies that I haven't had much time for coding. Does #549 work to solve this? It should work only on the output stream.

ghutchis:
@cryos - the locale needs to be set before XML formatting happens, so it can't just be done on the stream. (Maybe there's a different XML library, but pugixml needs the locale set.)

avogadro/io/cmlformat.cpp
ghutchis: See my comments below - this seems to be the only solution with pugixml.

luancarvalhomartins:
> 
> 
> @luancarvalhomartins - apologies that I haven't had much time for coding. Does #549 work to solve this? It should work only on the output stream.

No apologies needed. I'll give #549 a try and let you know.

ghutchis:
I had another thought. Rather than let the XML code format the numbers, we can use a `stringstream` and use `imbue()` to create the  strings ourselves.

I'm going to go hack this now...

ghutchis:
Okay, #549 definitely works to solve this one and doesn't set the global locale.

I'm  going to close this PR - but thank you **very much** for pushing this.

PR #488, opened 2020-09-06
Allow use of HDF5 >= 1.12.0

berquist:
The HDF5 API changed for the major release 1.12.0, but rather than change the code, this uses an [application-level compilation flag](https://confluence.hdfgroup.org/display/HDF5/API+Compatibility+Macros) to tell newer library versions to use the older API.


ghutchis:
@berquist - thanks, this looks fine for me. Can you merge in master (with the build fixes)?

Also, we ask for signed commits for DCO (which should also force the re-build):
```
git commit --amend --signoff
git push --force-with-lease origin
```

ghutchis:
Thanks @berquist - looks like a useful solution for now.

Issue #439, opened 2020-02-10
export molecule fails silently: returns back to save dialog


drew-parsons:
**Avogadro version: 1.93.0**
 - Avogadrolibs: 1.93.0 (debian 1.93.0-1)
 - Qt: 5.12.5

**Desktop version: (please complete the following information):**
 - OS: Linux (Debian)
 - Version Linux 5.4.13 on debian unstable
 - Compiler: g++ 9.2.1

**Describe the bug**

The Export to Molecule functionality is failing (trying to save to file, e.g. to .cml, or .tmol or .xyz).
The save dialog box pops up, but when a file and format is selected, the save dialog box pops up again instead of generating the molecular coordinate file.

There is some instability in the Export to Molecule functionality (obviously).  The bug is reproducible every time, but in the middle of attempting to export, it sometimes does write a file anyway (while still popping back to the Save dialog).  These "successful" writes are not reproducible, i.e. export is only randomly successful.

**To Reproduce**
1. Create a molecule (e.g. click on initial screen to create methane)
2. Go to File->Export->Molecule in the menu
3. When save dialog comes up, type "methane.cml" (or any other relevant suffix)
4. Dialog box pops up to ask "Multiple writers found that can write this format. Which should be used?"
Select "Avogadro: CML".  (or select "OpenBabel: Chemical Markup Language", the failure is the same)
5. The Save dialog pops up again with no error message, and the coordinate file has not been written.

**Expected behavior**
After step 4, a coordinate file (methane.cml) should have been written, and the save dialog box should not reappear.

RizzerOnGitHub:
I suspect the problem might be related to how the obabel process is spawned to generate the files.

RizzerOnGitHub:
I can see some pattern in the successful write: write seems to proceeds if I cancel the second Save dialog box (e.g. press the  Esc button).  So the problem is, why is it returning back to that Save dialog?  Perhaps it is a bug in avogadroapp rather than avogadrolib?

cryos:
I can reproduce, we moved to latest Open Babel and this may be one I missed. I will see what I can do to get this fixed up. Thanks for reporting, and for the great description.

anubh4v:
There is another thing I noticed. When save dialog pops up again, if you click on cancel the file gets saved successfully.
![ezgif com-video-to-gif1](https://user-images.githubusercontent.com/52047994/105838132-aed8c480-5ff5-11eb-9300-4eb82a675b81.gif)

[...]

PR #517, opened 2021-01-29
Fixes bug #439

anubh4v:
Issue was with the do while loop in line 123 of fileformatdialog.cpp. Condition was true instead of false.

Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.


anubh4v:
@cryos please look into this.

ghutchis:
I'm not sure I understand. One commit changes a line - and the next commit changes it back?

anubh4v:
I am trying to replace "true" with "false" in the original file.

ghutchis:
Sorry for the delayed review. I think I'd be happier if `QString fileName` was moved out of the loop and the termination would be:
```
  } while (fileName.isEmpty());
```

It's functionally the same, but I think the code would be more readable.

anubh4v:
If `QString fileName` is moved outside the loop then we won't be able to change the format when `continue` is encountered in the loop. Since you need to change the fileName extension to change the format.

ghutchis:
I mean something like this:

```
  QString fileName;
  do { // jump point for continue statements on retry
    fileName = QFileDialog::getSaveFileName(parentWidget, caption, dir, realFilter);

    if (fileName.isEmpty()) // user cancel
      return result;
.. .
  } while (fileName.isEmpty());
```

ghutchis:
Thanks - just waiting for Mac / Win builds to complete (which they should)

ghutchis:
Sorry, one last minor thing before merge - can you sign off on your commits?
(This indicates that these changes are yours.)
https://github.com/OpenChemistry/avogadrolibs/pull/517/checks?check_run_id=2253882029

ghutchis:
Thanks for the fix and your patience on this one. I have more flexible time to contribute to Avogadro for the next few months, so future pull requests should get merged faster.

PR #520, opened 2021-02-12
Simplify wheel building

psavery:
This moves wheel building over to use [cibuildwheel](https://github.com/joerick/cibuildwheel) and github actions. This simplifies the wheel building logic with a lot of shared script and code paths between platforms.

On new tag pushes, the built wheels will automatically be pushed to PyPI.

The following wheels were tested:

* Linux: 3.8 - 3.9
* Mac: 3.5 - 3.9
* Windows: 3.8 - 3.9


ghutchis:
I'll take a look at the other build failures - I've been busy with deadlines, but want to make sure we get regular builds / releases again.

ghutchis:
I think if you push again, it should fix the build_cmake failures - it was a problem with the openchemistry cmake files.

psavery:
@ghutchis Sounds good, I am simplifying one or two things. But it should be ready soon.

psavery:
I think this is ready to go!

ghutchis:
I'll check on the Windows build later - I don't think that's a serious problem.

Do you know if we still need the Travis build or can we disable that after merging this? (I know it was to build wheels..)

ghutchis:
Looks good - the Windows break is on my side - should be fixed soon.

Issue #585, opened 2021-05-22
Merge RPC client code to avoid MoleQueue dependency


ghutchis:
As discussed in a Zoom call, right now the input generators depend on BUILD_MOLEQUEUE.

It would be great to migrate / copy the RPC client classes to `avogadrolibs` to eliminate the dependency (e.g., you don't need to have MoleQueue to run the input generator classes).

PR #596, opened 2021-05-28
Import molequeue client

cryos:


ghutchis (approved):
I still think we should switch everything to nlohmann JSON, but this looks good.

ghutchis:
This fixe #585

Issue #614, opened 2021-06-08
Selecting atoms tools fails 


serk12:
**Avogadro version:**
 - Avogadrolibs: 1.90.0 (master)
 - Qt: 5.12.8

**Desktop version:**
 - OS: ubuntu 20.04

**Describe the bug**

- The selecting tool have a strange behavior, I think it's when selecting more than half a bond.
- The selecting tool don't select if you move from bottom-left to top-right and vice versa. 

**To Reproduce**
1. Create a simple ethane molecule
2. Change to selection (don't move the camera)
3. Select the left part to 60-70% the bond

**Expected behavior**
Select only the left part

**Screenshots**

https://user-images.githubusercontent.com/10197352/121248032-3804df80-c8a3-11eb-8427-cdf08fbe24cc.mp4

ghutchis:
`master HEAD` at the moment is 1.94, but it's worth investigating. I found a few selection bugs in the last few weeks.

serk12:
No I have a normal 1080p screen. I also think it's something with the mapping, if you move the camera the bug behaves completely different, like bottom-left to top-right works but top-right to bottom-left don't, etc...

serk12:
I've downloaded the Windows version and nothing from the selecting tool does work (I use windows 10 home).

ghutchis:
@serk12 - do you want to try the Windows version now? It should be:
https://github.com/OpenChemistry/avogadrolibs/suites/2960459395/artifacts/66696306

Please let me know if we should close this.

serk12:
Yes you can close it. I was testing the [old 1.91.0 version](https://www.openchemistry.org/downloads/).

ghutchis:
Yes, that version didn't have the "rubber band selection" implemented yet.

Issue #452, opened 2020-03-31
Selection tool 'Apply color' => extension


ghutchis:
The selection tool has code to apply custom colors to a set of selected atoms.

This is cool, but it's not easily discoverable - you have to tap on the semi-invisible button next to the "apply color" label.
<img width="230" alt="Screen Shot 2020-03-31 at 11 24 37 AM" src="https://user-images.githubusercontent.com/41128/78044098-37d58f80-7342-11ea-9904-a6e594cbf92c.png">

I think the code should migrate to an Extension that operates on a selection and is a menu item, for example under the "View" menu, e.g "Apply custom color". (This could be extended to apply colors based on the Avo1 color plugins .. partial charges, etc.)

The good news is that the rendering code already supports custom colors.
<img width="405" alt="Screen Shot 2020-03-31 at 11 26 10 AM" src="https://user-images.githubusercontent.com/41128/78044289-74a18680-7342-11ea-943e-cd3f9c6b19b7.png">

ghutchis:
Avo1 color schemes:
- partial charges
- distance
- atom index
- residue
- color by SMARTS (easier is to add 'select by SMARTS' and custom color to a selection)

Krzmbrzl:
Adding a frame around the color box would also make it more visible

PR #615, opened 2021-06-09
Fixes to the selection tools

serk12:
I've fixed the two errors from #614, and the enhancement from #452.

- for the bottom-left to top-right: the points where flipped.
- the random selection: the selection computed the bound IDs as atom IDs creating unexpected behaviors.
- Removed the flat property from `selectiontoolwidget.ui` the solution is simple and visual, but if you don't like it I can try a more complex solution. 

Also I added a double click selection, and correct the draw function from disable the camera move.

### DIFF
--- a/avogadro/qtplugins/selectiontool/selectiontool.cpp
+++ b/avogadro/qtplugins/selectiontool/selectiontool.cpp
@@ -39,6 +39,10 @@
 #include <QtGui/QMouseEvent>
 #include <QtWidgets/QAction>
 
+#include <iostream>
+#include <queue>
+#include <set>
+
 using Avogadro::Core::Array;
 using Avogadro::Core::Atom;
 using Avogadro::QtGui::Molecule;
@@ -54,7 +58,7 @@ SelectionTool::SelectionTool(QObject* parent_)
   : QtGui::ToolPlugin(parent_), m_activateAction(new QAction(this)),
     m_molecule(nullptr), m_renderer(nullptr),
     m_toolWidget(new SelectionToolWidget(qobject_cast<QWidget*>(parent_))),
-    m_drawSelectionBox(false)
+    m_drawSelectionBox(false), m_doubleClick(false), m_initSelectionBox(false)
 {
   m_activateAction->setText(tr("Selection"));
   m_activateAction->setIcon(QIcon(":/icons/selectiontool.png"));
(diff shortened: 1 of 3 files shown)


avogadro/qtplugins/selectiontool/selectiontool.cpp
```diff
@@ -76,84 +80,76 @@ QUndoCommand* SelectionTool::mousePressEvent(QMouseEvent* e)
...
+  m_initSelectionBox = true;
   m_start = Vector2(e->pos().x(), e->pos().y());
   m_end = m_start;
-  Identifier hit = m_renderer->hit(e->pos().x(), e->pos().y());
-
-  // If an atom is clicked, accept the event, but don't add it to the atom list
-  // until the button is released (this way the user can cancel the click by
-  // moving off the atom, and the click won't get passed to the default tool).
-  if (hit.type == Rendering::AtomType)
-    e->accept();
```
serk12, line 87: don't know why there is a double accept, so I've removed one.

avogadro/qtplugins/selectiontool/selectiontool.cpp
```diff
@@ -39,6 +39,10 @@
 #include <QtGui/QMouseEvent>
 #include <QtWidgets/QAction>
 
+#include <iostream>
```
ghutchis, line 42: Did we need this? If you want to debug in any of the Qt classes:
```
#include <QDebug>

..
qDebug() << " debugging message ";
```

avogadro/qtplugins/selectiontool/selectiontool.cpp
```diff
@@ -226,30 +225,76 @@ void SelectionTool::applyColor(Vector3ub color)
...
+void SelectionTool::selectLinkedMolecule(QMouseEvent* e, Index atom)
+{
+  std::queue<Index> toSelect;
+  std::set<Index> done;
+  toSelect.push(atom);
+  while (!toSelect.empty()) {
+    atom = toSelect.front();
+    toSelect.pop();
+    selectAtom(e, atom);
+    auto bonds = m_molecule->bonds(atom);
```
ghutchis, line 237: This is probably a useful iterator "get connected component." I'm working on some iterators over the angles and torsions, so I'll look at adding a range iterator for "visit all atoms connected to this" and some similar things.

avogadro/qtplugins/selectiontool/selectiontool.cpp
```diff
@@ -72,97 +76,92 @@ QWidget* SelectionTool::toolWidget() const
...
 QUndoCommand* SelectionTool::mouseReleaseEvent(QMouseEvent* e)
 {
   // If the click is released on an atom, add it to the list
-  if (e->button() != Qt::LeftButton || !m_renderer)
+  if (e->button() != Qt::LeftButton || !m_renderer || m_doubleClick) {
+    m_doubleClick = false;
     return nullptr;
-
+  }
+  shouldClean(e);
```
ghutchis, line 99: If you click on a single atom and then click again, I think the expectation is that it selects and then deselects.

As far as I can tell, this clears the selection .. so later on, it doesn't know that the atom was previously selected.

See below.

avogadro/qtplugins/selectiontool/selectiontool.cpp
```diff
@@ -72,97 +76,92 @@ QWidget* SelectionTool::toolWidget() const
...
-    } else {
-      // If the shift modifier is not pressed clear the previous selection.
-      if (!(e->modifiers() & Qt::ShiftModifier)) {
-        clearAtoms();
-      }
-      for (auto it = hits.begin(); it != hits.end(); ++it) {
-        addAtom(*it);
+    m_initSelectionBox = false;
+    auto hits = m_renderer->hits(start.x(), start.y(), end.x(), end.y());
+    for (auto it = hits.begin(); it != hits.end(); ++it) {
```
ghutchis, line 112: Can we switch this to the new `for (auto it: hits)` loop syntax? I'm hoping to submit a match to migrate loops soon.

avogadro/qtplugins/selectiontool/selectiontool.cpp
```diff
@@ -72,97 +76,92 @@ QWidget* SelectionTool::toolWidget() const
...
-        toggleAtom(*it);
-      }
-    } else {
-      // If the shift modifier is not pressed clear the previous selection.
-      if (!(e->modifiers() & Qt::ShiftModifier)) {
-        clearAtoms();
-      }
-      for (auto it = hits.begin(); it != hits.end(); ++it) {
-        addAtom(*it);
+    m_initSelectionBox = false;
```
ghutchis, line 110: I think this is where you need to call `shouldClean()`

avogadro/qtplugins/selectiontool/selectiontool.cpp
```diff
@@ -72,97 +76,92 @@ QWidget* SelectionTool::toolWidget() const
...
   } else {
     // Single click
     m_start = Vector2(e->pos().x(), e->pos().y());
     m_end = m_start;
     Identifier hit = m_renderer->hit(e->pos().x(), e->pos().y());
-
     // Now add the atom on release.
     if (hit.type == Rendering::AtomType) {
-      toggleAtom(hit);
+      toggleAtom(hit.index);
```
ghutchis, line 124: By moving the `shouldClean()` this should work for single atom toggle-on / toggle-off

avogadro/qtplugins/selectiontool/selectiontool.cpp
```diff
@@ -72,97 +76,92 @@ QWidget* SelectionTool::toolWidget() const
...
+    Vector2 select = Vector2(e->pos().x(), e->pos().y());
+    Identifier hit = m_renderer->hit(select.x(), select.y());
+    // Reset the atom list
+    if (!hit.isValid()) {
+      clearAtoms();
+    } else {
+      shouldClean(e);
+      m_drawSelectionBox = false;
+      selectAtom(e, hit.index);
+      selectLinkedMolecule(e, hit.index);
```
ghutchis, line 148: This is really a great idea - very intuitive. Thanks!

avogadro/qtplugins/selectiontool/selectiontool.cpp
serk12: Range-based don't support iterators, for this case, as far as I know, not important but is something to keep in mind.

avogadro/qtplugins/selectiontool/selectiontool.cpp
serk12: Ups sorry.

avogadro/qtplugins/selectiontool/selectiontool.cpp
serk12: Yes. Funny thing is that now double click followed by ctrl click does a complementary selection.

avogadro/qtplugins/selectiontool/selectiontool.cpp
serk12: Note this is `O(atoms*bonds*log(atoms))` and for large structures may be slow. maybe a dinamic solution using a structure ID and pointers should be considerated.

ghutchis (approved):

PR #620, opened 2021-06-18
connected group class + move overlayaxes to view menu

serk12:
I've added a `UniqueGroup` class which is a bidirectional map between a ID and a set of IDs, to calculate on streaming the connectivity from multiple groups.
Also I've moved the reference axes inside the view menu because it's more intuitive than Display Types. 

The PR is not complete there is two errors. 
The first one is when I try to get the graph in the `selectiontool.h:240` using `m_molecule->graph()` the program crashes. My theory is that `RWMolecule` class is not talking with `Core::Molecule` and not updating or setting dirty to true. This ends up in crashing the program.

The second one is a minnor thing, but the all the tools, except navigation, clears the `GroupNode` removing the overlay axes, My solution would be connect the overlay plugin with a update signal, but I don't know if it exists or where to put it.


ghutchis:
It would be great to separate the display axes bits from the graph bits, since they're logically unrelated.
(I'm happy moving the axes to the View menu.)

avogadro/core/molecule.cpp
```diff
@@ -31,7 +31,7 @@ namespace Avogadro {
 namespace Core {
 
 Molecule::Molecule()
-  : m_graphDirty(false), m_basisSet(nullptr), m_unitCell(nullptr)
+  : m_graphDirty(true), m_basisSet(nullptr), m_unitCell(nullptr)
```
ghutchis, line 34: If it's an empty molecule, by definition the graph can't be dirty.

avogadro/core/connectedgroup.cpp
```diff
@@ -0,0 +1,161 @@
+/******************************************************************************
+  This source file is part of the Avogadro project.
+  This source code is released under the 3-Clause BSD License, (see "LICENSE").
+******************************************************************************/
+
+#include "connectedgroup.h"
+
+#include <algorithm>
+#include <cassert>
```
ghutchis, line 9: No asserts. If you want to add in something that prints warnings / errors to stdout or stderr, that's fine.

avogadro/core/connectedgroup.cpp
```diff
@@ -0,0 +1,161 @@
+/******************************************************************************
+  This source file is part of the Avogadro project.
+  This source code is released under the 3-Clause BSD License, (see "LICENSE").
+******************************************************************************/
+
+#include "connectedgroup.h"
```
ghutchis, line 6: I guess my main question is philosophical. Do you need this as a separate class rather than adding / modifying the graph class? Some of this seems like it duplicates work that the graph class already does.

avogadro/core/connectedgroup.cpp
```diff
@@ -0,0 +1,161 @@
...
+}
+
+void ConnectedGroup::removeConnection(size_t a, std::vector<size_t> a_neighbors,
+                                      size_t b, std::vector<size_t> b_neighbors)
+{
+  assert(m_elementToGroup.find(a) != m_elementToGroup.end());
+  assert(m_elementToGroup.find(b) != m_elementToGroup.end());
+  assert(m_elementToGroup[a] == m_elementToGroup[b]);
+
+  bool stillCOnected = false;
```
ghutchis, line 107: is this a typo? I'd normally use `stillConnected` here.

avogadro/core/graph.cpp
```diff
@@ -57,18 +67,20 @@ bool Graph::isEmpty() const
 void Graph::clear()
 {
   setSize(0);
+  connectedGroup->clear();
 }
 
 size_t Graph::addVertex()
 {
+  connectedGroup->addElement(size() + 1);
```
ghutchis, line 75: These make me think it's better to just add the ConnectedGroup code directly to the graph class.

Is there a reason you'd have a ConnectedGroup without a Graph? (Seems like "no")
Is there a reason you'd have a Graph without a ConnectedGroup (As written, this is "no")

ghutchis (changes requested):
I'd like to see the axes change split into a separate PR. I'm also not quite sure why it's a new class rather than just new methods in the Graph class - which seems to make more sense to me.

avogadro/core/connectedgroup.cpp
ghutchis: Asserts can crash the UI, which is extremely confusing to an end-user.

ghutchis:
I think this should get RWMolecule to update the graph state:

[rwgraph.txt](https://github.com/OpenChemistry/avogadrolibs/files/6683019/rwgraph.txt)

avogadro/core/connectedgroup.cpp
serk12: I will use this class to control the groups in the layering system. Maybe it can be inside the graph file. But this has more use outside the graph idea.

avogadro/core/molecule.cpp
serk12: This was just to double check the RWMolecule problem, but yeas by definition should be false.

ghutchis:
Closing to have 2 separate PR...

PR #648, opened 2021-07-05
Refactor Molecules class 

serk12:
I've refactored the `RWMolecule`, `QtGui::Molecule` and `Core::Molecule` so the Molecule graph is dynamic and also compute the connectivity between nodes.

Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.


ghutchis:
The build failures are from the `pybind11` Python bindings. Looks like most boil down to:
```
2021-07-05T15:39:52.3786938Z        .def("add_atom", &Molecule::addAtom, "Add an atom")
```

I think the problem is that you've now added the overload, `pybind` can't figure out which method to bind:
https://pybind11.readthedocs.io/en/stable/classes.html#overloaded-methods

So you'll just need to tweak `python/core.cpp`:
https://github.com/OpenChemistry/avogadrolibs/blob/master/python/core.cpp#L46

Probably users would want to access both methods...

ghutchis (changes requested):
Overall, it's good, but I'd really like to keep more inline functions and combine the MoleculeGraph and Molecule classes - I'm not sure I see the point of separating them.

avogadro/core/moleculegraph.h
```diff
@@ -0,0 +1,194 @@
+/******************************************************************************
```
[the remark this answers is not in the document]
serk12: Yes, I did this to ensure that `m_graph` is always sync with `m_bondPairs` and `m_atomicNumbers` and separate this part from Molecule class was a first step to ensure it.  if you prefer I can merge it with molecule again.

avogadro/core/molecule.cpp
```diff
@@ -21,7 +21,13 @@ namespace Core {
 
 using std::swap;
 
+<<<<<<< Updated upstream
```
ghutchis: This looks like a merge conflict

avogadro/core/molecule.cpp
```diff
@@ -270,7 +292,12 @@ Molecule::AtomType Molecule::addAtom(unsigned char number, Vector3 position3d)
 
 void Molecule::swapBond(Index a, Index b)
 {
+<<<<<<< Updated upstream
```
ghutchis: Again, looks like a merge conflict. Does  it compile?

avogadro/core/molecule.cpp
```diff
@@ -346,9 +426,35 @@ Molecule::BondType Molecule::addBond(const AtomType& a, const AtomType& b,
...
+  return n * aproxLog;
+}
+
 bool Molecule::removeBond(Index index)
 {
-  return Core::MoleculeGraph::removeBond(index);
+  if (index >= bondCount())
+    return false;
+  if (!m_graphDirty) {
+    // mark dirty the graph O(n) only if is more effitien than remove an edge
```
ghutchis, line 429: "efficient"

avogadro/core/molecule.cpp
```diff
@@ -737,5 +847,203 @@ Residue& Molecule::residue(Index index)
...
+  std::map<unsigned char, size_t> composition;
+  for (Array<unsigned char>::const_iterator it = m_atomicNumbers.begin(),
+                                            itEnd = m_atomicNumbers.end();
+       it != itEnd; ++it) {
+    ++composition[*it];
+  }
+  return composition;
+}
+
+Index Molecule::findBond(Index atomId1, Index atomId2) const
```
ghutchis, line 935: I think we can omit "find" and just call it Molecule::bond(Index, index) - it would match Molecule::atom(index) which already exists.

avogadro/core/molecule.cpp
```diff
@@ -737,5 +847,203 @@ Residue& Molecule::residue(Index index)
...
+    if (bond.first == atom || bond.second == atom) {
+      removeBond(i);
+    } else {
+      ++i;
+    }
+  }
+  return true;
+}
+
+Array<std::pair<Index, Index>> Molecule::getBonds(Index index) const
```
ghutchis, line 964: This could probably use Molecule::getAtomBonds(index) to clarify it's  the  index to the atom, not a bond index.

ghutchis (commented):
Looks pretty good, although it seems like there are some merge conflicts?

ghutchis:
The "Wheel" failures can be fixed with this patch:
```
diff --git a/python/core.cpp b/python/core.cpp
index ae4e9901..f105d38e 100644
--- a/python/core.cpp
+++ b/python/core.cpp
@@ -43,7 +43,7 @@ PYBIND11_MODULE(core, m)
 
   py::class_<Molecule>(m, "Molecule")
     .def(py::init<>())
-    .def("add_atom", &Molecule::addAtom, "Add an atom")
+    .def("add_atom", static_cast<Atom (Molecule::*)(unsigned char)>(&Molecule::addAtom), "Add an atom")
     .def("atom_count", atomCount0, "The number of atoms")
     .def("atom_count", atomCount1,
          "The number of atoms with the supplied atomic number")
```

avogadro/core/molecule.cpp
serk12: There is already a `BondType bond(Index atomId1, Index atomId2) const` I cannot overload the `bond` function. Any alternative name?

avogadro/core/molecule.h
```diff
@@ -883,54 +734,6 @@ inline bool Molecule::isSelectionEmpty() const
   return true;
 }
 
-inline std::pair<Index, Index> Molecule::bondPair(Index bondId) const
```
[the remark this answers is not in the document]
serk12: I think I've "inlined" all of them not sure if I missed one.

avogadro/core/moleculegraph.h
```diff
@@ -0,0 +1,194 @@
...
+  void addBonds(const Array<std::pair<Index, Index>>& bonds,
+                const Array<unsigned char>& orders);
+
+  // chenge the bond index position
+  void swapBond(Index a, Index b);
+  // channge the Atom index position
+  void swapAtom(Index a, Index b);
+
+protected:
+  // update base info first, next derivated
```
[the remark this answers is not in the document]
serk12: Now I've moved all of them to `core/molecule.h` but the idea is keeping the graph sync with the arrays. So everything except that is keep in protected/public and this three elements can only be modified by the class (only const reference getter allowed).

avogadro/core/molecule.cpp
serk12: The same but returns the index. I just realize I can use `bond` and extract the index.

ghutchis:
Everything looks good from my side. @cryos if you're busy, I'm just going to go ahead and merge this on Wednesday.

ghutchis (approved):

PR #676, opened 2021-07-15
Reverse direction of zoom with scrolling

### LINKED ISSUE #670: Zooming with mouse wheel/touchpad is reversed
**Avogadro version: 1.94.0-17-gb2b873b**
 - Avogadrolibs: 1.94.0-94-g2d86f8c
 - Qt: 5.12.8

**Desktop version: (please complete the following information):**
 - OS and Version: Linux Mint 20.2 (based on Ubuntu 20.04
 - Compiler: g++ 9.3.0

**Describe the bug**
On default settings, scrolling on a mouse wheel (or track pad) scrolls opposite from the OS settings. I tried reversing scroll direction through the OS and it was still opposite from what I would expect.

**To Reproduce**
Load some atoms and then scroll up/down.

**Expected behavior**
Normally, if the user has "Reverse scrolling" (Windows default), scrolling down would zoom out, but Avogadro zooms in. Similarly, "Natural scrolling" (Mac default) is also reversed from what is expected.

**Additional context**
While I believe the current issue to be a bug, this could also be a feature request to put an **option** in to allow for the scrolling settings to be different from the OS settings.

I was able to get the desired scrolling direction by negating the zoom factor below to be `-e->delta() * 0.1`.

https://github.com/OpenChemistry/avogadrolibs/blob/2d86f8cdb657b7fc5ce900b0f36fb6f93e6a1665/avogadro/qtplugins/navigator/navigator.cpp#L143

mquevill:
This affects both mouse and touchpad. It also replaces the obsolete `.delta()` function with `.angleDelta()`. The pixel-level function `.pixelDelta()` does not have direct meaning when performing zoom. The conversion between the two delta functions is 1/8, hence the change in scaling to match the pixel-level control speed in Qt.

Fixes #670.

Signed-off-by: Michael Quevillon <mquevill@nd.edu>

Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

### DIFF
--- a/avogadro/qtplugins/navigator/navigator.cpp
+++ b/avogadro/qtplugins/navigator/navigator.cpp
@@ -139,8 +139,18 @@ QUndoCommand* Navigator::mouseDoubleClickEvent(QMouseEvent* e)
 QUndoCommand* Navigator::wheelEvent(QWheelEvent* e)
 {
   /// @todo Use scale for orthographic projections
-  // Zoom
-  zoom(m_renderer->scene().center(), -e->angleDelta() * 0.125);
+  // Amount to zoom
+  float d = 0.0f;
+
+  QPoint numPixels = e->pixelDelta();
+  QPoint numDegrees = e->angleDelta() * 0.125;
+
+  if (!numPixels.isNull())
+    d = numPixels.y();
+  else if (!numDegrees.isNull())
+    d = numDegrees.y();
+
+  zoom(m_renderer->scene().center(), -d);
 
   e->accept();
   emit updateRequested();
--- a/avogadro/qtplugins/navigator/navigator.cpp
+++ b/avogadro/qtplugins/navigator/navigator.cpp
@@ -140,7 +140,7 @@ QUndoCommand* Navigator::wheelEvent(QWheelEvent* e)
 {
   /// @todo Use scale for orthographic projections
   // Zoom
-  zoom(m_renderer->scene().center(), e->delta() * 0.1);
+  zoom(m_renderer->scene().center(), -e->angleDelta() * 0.125);
 
   e->accept();
   emit updateRequested();


mquevill:
I cannot test if the pinch gesture (on Mac, for example) is affected by this change. I think that case is covered by something other than QWheelEvent.

ghutchis:
Cool, I'll take a "spin" tonight. Thanks!

ghutchis:
Okay, I changed to fix the compile to use `angleDelta().y()`

On my MacBook trackpad:
- scroll down = zoom in
- scroll up = zoom out

Pinch-to-zoom no longer works.

I'd really like to see you post about this and get feedback: https://discuss.avogadro.cc/

The code change is minimal, but I'd like to know if other people feel the current behavior is an issue. I suspect it also needs `#ifdef` for platforms, because on my Mac, the current behavior is correct.

mquevill:
I posted on the forums as well as updating the pull request. As previously mentioned, there are both `.angleDelta()` and `.pixelDelta()` that replace the obsolete `.delta()`. I rewrote it in the recommended way from the Qt documentation (https://doc.qt.io/qt-5/qwheelevent.html#angleDelta). I am thinking that the pinch-to-zoom gesture might only send the changes via `.pixelDelta()`, whereas two-finger touchpad scrolling sends both pixels and angle.

Alternatively, pinch-to-zoom may need to be captured by a QGestureEvent instead of a QWheelEvent. Unfortunately, I don't have any devices that support pinch-to-zoom to test.

ghutchis:
Considering there hasn't been a bunch of feedback, my suggestion is that we add this as an option / checkbox in the Navigator tool. Maybe "Reverse direction of zoom" for wording? (I'm open to suggestions..)

I'd be happy to help set up the widget - at the moment the main thing would be to add a class member variable that changes the sign in that code.

ghutchis:
Somehow I forgot to push this - and for some reason I can't amend this pull request. So I'm closing in favor of #778 which includes all of this. Let me know what you think there.

Issue #701, opened 2021-07-27
Build failure for Mac wheel


ghutchis:
There is now a consistent build failure for Python wheels on Mac. I can't make sense of the failure, since it seems to be using Python 3.6 just fine.

Presumably there's now a problem finding the dylib?

[build-wheel.txt](https://github.com/OpenChemistry/avogadrolibs/files/6881823/build-wheel.txt)

@psavery - if you can take a look, I'm happy to do the fix... Perhaps @cryos may have ideas too?

psavery:
Hi @ghutchis, I should be able to look into this soon.

I wonder if it might be caused by a new version of [`delocate`](https://github.com/matthew-brett/delocate). I see previous CI runs that succeeded were using `delocate==0.8.2`, whereas the new ones that are failing are using `delocate==0.9.0`.

ghutchis:
Makes sense. The ChangeLog is here .. looks like `--ignore-missing-dependencies` might be useful...
https://github.com/matthew-brett/delocate/blob/master/Changelog

PR #704, opened 2021-07-27
Try downgrading cibuildwheel as a quick fix...

psavery:
This is an attempt to be a quick fix of #701. A better fix will be
to figure out why the new version `delocate` is failing, and fix
that so that we can upgrade `cibuildwheel`.


ghutchis:
Looks like it fixes that build issue for now.

ghutchis:
I think this is a good workaround, thanks!

PR #709, opened 2021-07-30
Layer system

serk12:
Signed-off-by: Marc Prat Masó <marc.prat.maso@estudiantat.upc.edu>

Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

Some examples:


avogadro/qtgui/layermodel.cpp
```diff
@@ -0,0 +1,195 @@
...
+        }
+        case Qt::ForegroundRole:
+          if (layer == static_cast<int>(getMoleculeLayer().activeLayer()))
+            return QVariant(QColor(Qt::red));
+          else
+            return QVariant(QColor(Qt::black));
+        default:
+          return QVariant();
+      }
+    } else if (idx.column() == (QTTY_COLUMNS - 5)) {
```
ghutchis, line 85: I'd declare an enum  -- either in the header or here. That way you can have `(idx.column()  == LayerVisible)` and it makes it much easier to change / add  columns in the future.

e.g., `propertymodel.*`
https://github.com/OpenChemistry/avogadrolibs/pull/711/files

ghutchis (changes requested):
Generally looks good - definitely needs some doc comments in headers. Also seems like a `.png` crept in.

Somewhere we need to document the source of the icons (e.g., README)

avogadro/core/layer.cpp
```diff
@@ -0,0 +1,120 @@
...
+  m_atomAndLayers.clear();
+  m_activeLayer = m_maxLayer = 0;
+}
+
+size_t Layer::activeLayer() const
+{
+  return m_activeLayer;
+}
+
+size_t Layer::maxLayer() const
```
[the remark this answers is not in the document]
serk12: `MaxLayer` is the known maximum, so it starts at 0. If you prefer I could rewrite it, but it's not a counting variable.

avogadro/core/layermanager.h
```diff
@@ -0,0 +1,121 @@
...
+/**
+ * @class LayerData layermanager.h <avogadro/core/layermanager.h>
+ * @brief Interface to store layer data structure.
+ */
+struct LayerData
+{
+  LayerData(std::string save = "") { load(save); }
+
+  /** save custom data, base save should never be called */
+  virtual std::string save()
```
serk12, line 33: I'm not happy with this, is there anything in `Avogadro:IO` I could extend to? or any better idea?

avogadro/core/layer.cpp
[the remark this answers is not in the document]
ghutchis: Okay. Still seems like there should be a `layerCount()` method to parallel atoms, bonds, residues, etc, but I guess that's `maxLayer() + 1`

avogadro/core/layermanager.h
ghutchis: I mean, this is basically `serialize()` and `deserialize()` - you're not actually saving, you're just generating a string that represents the current layer configuration.

You can be sure the base method isn't called by making it a pure virtual method.

serk12: I can't use pure virtual method. I create `LayerData` objects in `cjsonformat.cpp` and then it's serialized in `pluginlayermanager.h` where I know the class type. But the renaming gives me some comfort .

avogadro/core/layermanager.h
```diff
@@ -0,0 +1,121 @@
...
+ * @brief Interface to store layer data structure.
+ */
+struct LayerData
+{
+  LayerData(std::string save = "") { deserialize(save); }
+
+  /** save custom data, base save should never be called */
+  virtual std::string serialize()
+  {
+    assert(true);
```
ghutchis, line 35: Can we drop the `assert` if it's always true?

ghutchis:
The Mac build failure needs `#include <string>` in `layermanager.h`

avogadro/core/layermanager.h
```diff
@@ -0,0 +1,121 @@
...
+
+#include "avogadrocore.h"
+
+#include "array.h"
+#include "layer.h"
+
+#include <cassert>
+#include <map>
+#include <memory>
+#include <vector>
```
ghutchis, line 17: `#include <string>`

PR #819, opened 2022-01-22
Fix erratic editing behavior

aerkiaga:
This fixes the regression described at https://github.com/OpenChemistry/avogadroapp/issues/239. This issue appeared with https://github.com/OpenChemistry/avogadrolibs/commit/d2dcfaead5080597c877246a7253c483f7668c3c, due to a small change in `avogadro/qtgui/rwmolecule_undo.h` wherein `RWMolecule::addAtom()` was made to be passed `m_atomId` rather than `m_atomUid`, when it should expect a UID.

The consequence was that, when two carbons were bonded and their hydrogens adjusted, the hydrogens on one of the carbons would be removed and not be added back, due to the passed UIDs actually being their IDs, which happened to be equal to the removed hydrogens' UIDs.

Then, trying to add atoms around this molecule with 5 atoms would pass atom ID 5 as UID, this being the UID of one of the existing hydrogens, which would simply get moved to the clicked location.

Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

### DIFF
--- a/avogadro/qtgui/rwmolecule_undo.h
+++ b/avogadro/qtgui/rwmolecule_undo.h
@@ -89,7 +89,7 @@ public:
   {
     assert(m_molecule.atomCount() == m_atomId);
     if (m_usingPositions)
-      m_molecule.addAtom(m_atomicNumber, Vector3::Zero(), m_atomId);
+      m_molecule.addAtom(m_atomicNumber, Vector3::Zero(), m_atomUid);
     else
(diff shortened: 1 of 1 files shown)


ghutchis:
This definitely seems to fix some bugs, so I'm going to merge - and I'll sort out the Ubuntu 20.04 build issue.

PR #825, opened 2022-02-23
Fix Manipulation and Bond Centric Manipulation tools

aerkiaga:
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

The issue fixed in this Pull Request is described at [this Debian package bug](https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=1005006). Simply, the Manipulation and Bond Centric Manipulation tools don't work.

The problem seems to arise from an Avogadro::Core::Molecule& being cast to a pointer and then stored. For some reason, this gives a different value as just using it in an expression without storing it first. I must admit I don't understand this, and it might be some special feature of the language I'm not aware of, or a compiler bug.

However, by the code one can deduce that both approaches should be expected to give the same result, so it's safe to simply replace the stored variable with an expression, and this workaround fully restores both tools.

### DIFF
--- a/avogadro/qtplugins/bondcentrictool/bondcentrictool.cpp
+++ b/avogadro/qtplugins/bondcentrictool/bondcentrictool.cpp
@@ -351,8 +351,7 @@ QUndoCommand* BondCentricTool::mousePressEvent(QMouseEvent* e)
   Rendering::Identifier ident = m_renderer->hit(e->pos().x(), e->pos().y());
 
   // If no hits, return. Also ensure that the hit molecule is the one we expect.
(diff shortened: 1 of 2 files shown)


aerkiaga:
Note: the bug occurs both with GCC 9.3.0-17ubuntu1~20.04 and Clang 10.0.0-4ubuntu1, so it's probably not a compiler bug.

ghutchis:
Looks good to me. Initial testing seems to confirm it fixes the bug. It's probably worth going through the other tools to see if there are similar patterns.

PR #829, opened 2022-03-09
Faster bond perception

aerkiaga:
This Pull Requests implements a new bond perception algorithm that is O(n) in the average case, down from O(n²). Among other probable improvements, this represents 5 - 15 times faster PDB import, making the Avogadro PDB path faster than Open Babel.

While the previous code tested every possible pair of atoms in a molecule, this new version first puts atoms in a 3D array of cubic bins with side equal to twice the maximum covalent radius plus tolerance, then only checks each atom against other atoms inside its bin and those in a 3D Moore neighborhood. For valid molecules, the number of operations performed must be, on average, linear to the number of atoms.

Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

### DIFF
--- a/avogadro/core/molecule.cpp
+++ b/avogadro/core/molecule.cpp
@@ -840,7 +840,7 @@ void Molecule::perceiveBondsSimple(const double tolerance, const double min)
   // note that the "worst case" here would need to be an invalid molecule
   for (Index i = 0; i < atomCount(); i++) {
     Vector3 ipos = m_positions3d[i];
-    const std::array<size_t, 3>& bin_index = atoms_bin[i];
+    const std::array<size_t, 3> &bin_index = atoms_bin[i];
     for (Index xi = std::max(0, signed(bin_index[0]) - 1);
         xi < std::min(bin_count[0], bin_index[0] + 2); xi++) {
       for (Index yi = std::max(0, signed(bin_index[1]) - 1);
--- a/avogadro/core/molecule.cpp
+++ b/avogadro/core/molecule.cpp
@@ -783,39 +783,90 @@ void Molecule::setVibrationLx(const Array<Array<Vector3>>& lx)
   m_vibrationLx = lx;
 }
 
-// bond perception code ported from VTK's vtkSimpleBondPerceiver class
 void Molecule::perceiveBondsSimple(const double tolerance, const double min)
 {
   // check for coordinates
-  if (m_positions3d.size() != atomCount())
+  if (m_positions3d.size() != atomCount() || m_positions3d.size() < 2)
     return;
 
   // cache atomic radii
   std::vector<double> radii(atomCount());
+  double max_radius = 0.0;
   for (size_t i = 0; i < radii.size(); i++) {
     radii[i] = Elements::radiusCovalent(atomicNumber(i));
     if (radii[i] <= 0.0)
       radii[i] = 2.0;
+    if (radii[i] > max_radius)
+      max_radius = radii[i];
+  }
+
+  // find molecule bounding box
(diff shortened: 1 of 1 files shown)


avogadro/core/molecule.cpp
```diff
@@ -803,7 +803,7 @@ void Molecule::perceiveBondsSimple(const double tolerance, const double min)
   // find molecule bounding box
   Vector3 min_pos = m_positions3d[0];
   Vector3 max_pos = m_positions3d[0];
-  for (Index i = 1; i < atomCount(); i++) {
+  for (Index i = 0; i < atomCount(); i++) {
```
ghutchis, line 806: You don't need to start the loop at 0 because you already grabbed the position of atom zero in the lines above.

avogadro/core/molecule.cpp
```diff
@@ -825,27 +825,27 @@ void Molecule::perceiveBondsSimple(const double tolerance, const double min)
...
+    atoms_bin[i] = bin_index;
   }
 
   // check for bonds
   // O(n) average-case, O(n^2) worst-case
   // note that the "worst case" here would need to be an invalid molecule
   for (Index i = 0; i < atomCount(); i++) {
     Vector3 ipos = m_positions3d[i];
-    std::array<size_t, 3> bin_index = atoms_bin[i];
-    for (Index xi = std::max(size_t(0), bin_index[0] - 1);
```
ghutchis: The `const` is good, but why did you change the rest of this? It should stay `size_t` rather than a `signed` - I'm getting compiler warnings from `clang`

ghutchis (changes requested):
I think my overall question is whether it adds overhead to smaller molecules. In my experience with Open Babel's implementation, this only showed significant speed boosts in large proteins.

That's not to say the change isn't welcome, but it would be great to try some benchmarking on small things like benzene and caffeine in addition to the large PDB files.

avogadro/core/molecule.cpp
aerkiaga: The problem is, `size_t` is an unsigned type, so `std::max(size_t(0), bin_index[0] - 1)` will produce weird results if `bin_index[0]` equals 0. Another option would be  `std::max(size_t(1), bin_index[0]) - 1`, which avoids the cast. I'll write that instead.

aerkiaga:
What use-cases do you have in mind for small molecules? With the Qt interface, those operatios are instantaneous anyway, and the major bottleneck is not this. And for things like applying a script to a library of compounds, the overhead could not even materialize: for small molecules, adding bonds takes a large share of the time, so while the extra loop is noticeable for e.g. benzene, any larger molecule (drugs, maybe even caffeine) would offset this with having to test less bonds. I'll benchmark it anyway.

ghutchis:
I'm not particularly concerned with benzene - because as you said, it's nearly instantaneous - but want to have a rough idea of the trade-off. 

Handling larger structures (proteins, materials, etc.) is clearly the bigger problem, so the new implementation is a win.

aerkiaga:
Okay, the benefit seems to be detectable up to relatively small molecules:
image

The smallest one I can actually see a spped-up with (without instrumentation) is about 800 atoms. Also, since each atom within ~3 bond lengths of others can save time, and just a few of them compensate for the extra pass, I'd say the overhead wouldn't appear for any molecule over 30 atoms.

Branching before an operation is a good consideration though. I'll add that too if you find it beneficial.

ghutchis:
If the overhead isn't significant for below 30 atoms, it makes no sense to add the branch. On most recent computers, the time required for 30 atoms isn't worth discussing.

Happy to merge.

PR #831, opened 2022-03-09
Formal charges

aerkiaga:
This Pull Requests improves Avogadro's support for formal charges. First of all, it implements reading them from CML, which also extends to other formats (e.g. SMILES, which is internally converted to CML). Those charges are then passed down the class hierarchy. Secondly, since PDB files can contain quaternary ammonium (e.g. acetylcholine, various drug ligands) or tertiary sulfonium (S-adenosylmethionine), or even quaternary arsonium (https://www.rcsb.org/structure/5NXY) groups, and those cannot be eliminated by hydrogen adjustment, specific code was added to detect those.

Then, the code was made more convergent with other paths so as to allow the various optimization PRs to act on it.

Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

### DIFF
--- a/avogadro/core/molecule.cpp
+++ b/avogadro/core/molecule.cpp
@@ -827,52 +827,6 @@ void Molecule::perceiveBondsFromResidueData()
   }
 }
 
-void Molecule::perceiveSubstitutedCations()
-{
-  for (Index i = 0; i < atomCount(); i++) {
-    unsigned char requiredBondCount(0);
-    switch(atomicNumber(i)) {
-      case 7:
-      case 15:
-      case 33:
-      case 51:
-        requiredBondCount = 4;
-        break;
-      case 8:
-      case 16:
-      case 34:
-      case 52:
-        requiredBondCount = 3;
-    }
-    if (!requiredBondCount)
-      continue;
-
-    unsigned char bondCount(0);
-    Index j = 0;
-    for (const auto &bond : bonds(i)) {
-      unsigned char otherAtomicNumber(0);
-      Index index1(bond.atom1().index());
-      Index index2(bond.atom2().index());
-      if (index1 == i) {
-        otherAtomicNumber = atomicNumber(index2);
-        bondCount += bond.order();
-      } else if (index2 == i) {
-        otherAtomicNumber = atomicNumber(index1);
-        bondCount += bond.order();
-      }
-      if (otherAtomicNumber && otherAtomicNumber != 6) {
-        bondCount = 0;
-        break;
-      }
-      j++;
-    }
-
-    if (bondCount == requiredBondCount) {
-      setFormalCharge(i, 1);
-    }
-  }
-}
-
 int Molecule::coordinate3dCount()
 {
   return static_cast<int>(m_coordinates3d.size());
--- a/avogadro/core/molecule.cpp
+++ b/avogadro/core/molecule.cpp
@@ -835,11 +835,13 @@ void Molecule::perceiveSubstitutedCations()
       case 7:
       case 15:
       case 33:
+      case 51:
         requiredBondCount = 4;
         break;
       case 8:
       case 16:
-      case 32:
+      case 34:
+      case 52:
         requiredBondCount = 3;
     }
     if (!requiredBondCount)
@@ -847,7 +849,7 @@ void Molecule::perceiveSubstitutedCations()
 
     unsigned char bondCount(0);
     Index j = 0;
-    for (auto &bond : bonds(i)) {
+    for (const auto &bond : bonds(i)) {
       unsigned char otherAtomicNumber(0);
       Index index1(bond.atom1().index());
       Index index2(bond.atom2().index());
--- a/avogadro/core/molecule.cpp
+++ b/avogadro/core/molecule.cpp
@@ -847,7 +847,7 @@ void Molecule::perceiveSubstitutedCations()
 
     unsigned char bondCount(0);
     Index j = 0;
-    for (auto& bond : bonds(i)) {
+    for (auto &bond : bonds(i)) {
       unsigned char otherAtomicNumber(0);
       Index index1(bond.atom1().index());
       Index index2(bond.atom2().index());
--- a/avogadro/core/molecule.cpp
+++ b/avogadro/core/molecule.cpp
@@ -847,16 +847,16 @@ void Molecule::perceiveSubstitutedCations()
 
     unsigned char bondCount(0);
     Index j = 0;
-    for (auto& pair : m_bondPairs) {
+    for (auto& bond : bonds(i)) {
       unsigned char otherAtomicNumber(0);
-      if (pair.first == pair.second) {
-      }
-      if (pair.first == i) {
-        otherAtomicNumber = atomicNumber(pair.second);
-        bondCount += m_bondOrders[j];
-      } else if (pair.second == i) {
-        otherAtomicNumber = atomicNumber(pair.first);
-        bondCount += m_bondOrders[j];
+      Index index1(bond.atom1().index());
+      Index index2(bond.atom2().index());
+      if (index1 == i) {
+        otherAtomicNumber = atomicNumber(index2);
+        bondCount += bond.order();
+      } else if (index2 == i) {
+        otherAtomicNumber = atomicNumber(index1);
+        bondCount += bond.order();
       }
       if (otherAtomicNumber && otherAtomicNumber != 6) {
         bondCount = 0;
--- a/avogadro/core/molecule.cpp
+++ b/avogadro/core/molecule.cpp
@@ -827,6 +827,50 @@ void Molecule::perceiveBondsFromResidueData()
   }
 }
 
+void Molecule::perceiveSubstitutedCations()
+{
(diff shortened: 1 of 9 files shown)


ghutchis:
The other part of this would be writing to CJSON since we're now using that preferentially over CML.

avogadro/core/molecule.cpp
```diff
@@ -827,6 +827,50 @@ void Molecule::perceiveBondsFromResidueData()
...
 
+void Molecule::perceiveSubstitutedCations()
+{
+  for (Index i = 0; i < atomCount(); i++) {
+    unsigned char requiredBondCount(0);
+    switch(atomicNumber(i)) {
+      case 7:
+      case 15:
+      case 33:
+        requiredBondCount = 4;
```
ghutchis, line 838: Might as well add 51 here as well.

avogadro/core/molecule.cpp
```diff
@@ -827,6 +827,50 @@ void Molecule::perceiveBondsFromResidueData()
...
+        break;
+      case 8:
+      case 16:
+      case 32:
+        requiredBondCount = 3;
```
ghutchis, line 843: Similarly, I'd add 52 here too.

avogadro/core/molecule.cpp
```diff
@@ -827,6 +827,50 @@ void Molecule::perceiveBondsFromResidueData()
...
+    }
+    if (!requiredBondCount)
+      continue;
+
+    unsigned char bondCount(0);
+    Index j = 0;
+    for (auto &bond : bonds(i)) {
```
ghutchis, line 850: `const auto`?

avogadro/core/molecule.cpp
```diff
@@ -827,6 +827,50 @@ void Molecule::perceiveBondsFromResidueData()
   }
 }
 
+void Molecule::perceiveSubstitutedCations()
```
ghutchis, line 830: Would this be better suited in `pdbformat` itself? Many formats (SDF, for example) don't need this.

If you wanted a more general form, say `perceiveFormalChargeFromValence()` it might make sense here .. check for both cations and anions .. something like `mdlvalence_p.h` but inverted - formal charges from the bond order.

ghutchis (commented):
I think the question would be more on the generality - whether it's worth expanding beyond the simple cations to a slightly larger formal charge perception method.

I also don't remember if I added formal charges to the native `mdlformat` code -- it looks like not.

ghutchis:
This fixes your problems for sure, but it would be great to have a slightly broader patch to address CJSON and SDF support too and ideally also handle negative formal charges.

avogadro/io/cmlformat.cpp
```diff
@@ -237,6 +237,13 @@ class CmlFormatPrivate
         }
       }
 
+      // Check formal charge.
+      attribute = node.attribute("formalCharge");
+      if (attribute) {
+        signed int formalCharge = lexicalCast<signed int>(attribute.value());
+        atom.setFormalCharge(formalCharge);
```
ghutchis, line 244: We should also make sure the formal charge is written on save for CML files.

aerkiaga:
Regarding CJSON and SDF... How are new features introduced in CJSON; can I just send a Pull Request to them? Also, I'm not familiar with SDF at all; never used it. Where could I find the specification?

ghutchis:
For CJSON, you're just adding an extension for atom formal charges, so it's backwards-compatible still.

In `CjsonFormat::read` - I'd do something like this for formal charges:
```
  json labels = atoms["labels"];
  if (labels.is_array() && labels.size() == atomCount) {
    for (size_t i = 0; i < atomCount; ++i) {
      molecule.atom(i).setLabel(labels[i]);
    }
  }
```

In `write` I'd look at something like:
```
  // labels
  json labels;
  for (size_t i = 0; i < molecule.atomCount(); ++i) {
    labels.push_back(molecule.label(i));
  }
  root["atoms"]["labels"] = labels;
```

avogadro/io/cmlformat.cpp
```diff
@@ -533,13 +540,17 @@ bool CmlFormat::write(std::ostream& out, const Core::Molecule& mol)
...
-      atomNode.append_attribute("x3") = formatNumber(floatStream,a.position3d().x()).c_str();
-      atomNode.append_attribute("y3") = formatNumber(floatStream,a.position3d().y()).c_str();
-      atomNode.append_attribute("z3") = formatNumber(floatStream,a.position3d().z()).c_str();
+      atomNode.append_attribute("x3") = formatNumber(numberStream,a.position3d().x()).c_str();
+      atomNode.append_attribute("y3") = formatNumber(numberStream,a.position3d().y()).c_str();
+      atomNode.append_attribute("z3") = formatNumber(numberStream,a.position3d().z()).c_str();
+    }
+    if(a.formalCharge() != 0) {
+      atomNode.append_attribute("formalCharge") =
+          formatNumber(numberStream, int(a.formalCharge())).c_str();
```
ghutchis, line 553: This shouldn't be needed, since it's an integer. We ran into problems with CML expecting the "C" locale and floats written as "1,0456"

aerkiaga:
How can I save a file as MDL SDF? Avogadro errors out stating that there are no writers for either MDL format...

avogadro/io/mdlformat.cpp
```diff
@@ -210,13 +248,20 @@ bool MdlFormat::write(std::ostream& out, const Core::Molecule& mol)
...
+    chargeList.push_back(std::pair(atom.index(), charge));
+    unsigned int chargeField = (charge < 0)?
+        ((charge >= -3)? 4 - charge : 0) :
+        ((charge <= 3)? charge : 0);
     out << setw(10) << std::right << std::fixed << setprecision(4)
         << atom.position3d().x() << setw(10) << atom.position3d().y()
         << setw(10) << atom.position3d().z() << " " << setw(3) << std::left
-        << Elements::symbol(atom.atomicNumber())
-        << "  0  0  0  0  0  0  0  0  0  0  0  0\n";
+        << Elements::symbol(atom.atomicNumber()) << "0 "
```
ghutchis, line 262: This isn't aligned correctly anymore. The symbol is fine, but the next zero should be " 0" (i.e., 2-character right-aligned) for the mass difference field.

avogadro/io/mdlformat.cpp
```diff
@@ -210,13 +248,20 @@ bool MdlFormat::write(std::ostream& out, const Core::Molecule& mol)
...
+        << setw(3) << std::right << chargeField /* for compatibility */
+        << "0  0  0  0  0  0  0  0  0  0  \n";
```
ghutchis, line 264: And this one starts with zero, so it's not 3-character right-aligned.

ghutchis (changes requested):

ghutchis:
Otherwise, this looks good. I was able to read in the test file, and adding hydrogens generated R-NH3+ and R-CO2- properly, which is great.

avogadro/io/mdlformat.cpp
```diff
@@ -226,6 +271,13 @@ bool MdlFormat::write(std::ostream& out, const Core::Molecule& mol)
         << bond.atom2().index() + 1 << setw(3) << static_cast<int>(bond.order())
         << "  0  0  0  0\n";
   }
+  // Properties block.
+  for (size_t i = 0; i < chargeList.size(); ++i) {
+    Index atomIndex = chargeList[i].first;
```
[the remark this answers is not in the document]
aerkiaga: Ok, I forgot to test that... Checking it before adding those atoms to the list now.

ghutchis:
Thanks, the SDF fix looks good. Anything else or is this good to merge?
Screen Shot 2022-03-11 at 2 51 11 PM

aerkiaga:
Well, there's that thing with halogen-substituted anions, but at this point it would only be added to PDB, and I'm not aware of any drugs or research ligands with such structures...

So... yes, good to merge!

aerkiaga:
Well, maybe I should support in CJSON? Looks small enough to me that you could merge it with that support tomorrow...

avogadro/io/cjsonformat.cpp
```diff
@@ -152,6 +152,14 @@ bool CjsonFormat::read(std::istream& file, Molecule& molecule)
     }
   }
 
+  // formal charges
+  json charges = atoms["charges"];
```
ghutchis, line 156: Sorry, I should have been explicit. Let's call it "formalCharges" because we'll probably want "partialCharges" sometime.

avogadro/io/cjsonformat.cpp
```diff
@@ -754,6 +762,13 @@ bool CjsonFormat::write(std::ostream& file, const Molecule& molecule)
   }
   root["atoms"]["labels"] = labels;
 
+  // formal charges
+  json charges;
+  for (size_t i = 0; i < molecule.atomCount(); ++i) {
+    charges.push_back(molecule.formalCharge(i));
+  }
+  root["atoms"]["charges"] = charges;
```
ghutchis, line 770: Same as above. Let's go with "formalCharges"

PR #832, opened 2022-03-09
Bond graph optimization

aerkiaga:
This Pull Request gives a big boost to many Molecule operations. It should make PDB import almost twice as fast (relative to #829), and hydrogen adjustment 15 times faster for macromolecules.

It adds a new member to `Core::Molecule`, a vector of vectors containing the bond indices associated with each particular atom in the molecule. This information could only be obtained in O8n) time, while now it should be O(1). This data is updated in parallel with `m_graph`.

Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

### DIFF
--- a/avogadro/core/connectedgroup.cpp
+++ b/avogadro/core/connectedgroup.cpp
@@ -196,6 +196,7 @@ std::vector<std::set<size_t>> ConnectedGroup::getAllGroups() const
 
 void ConnectedGroup::resetToSize(size_t n)
 {
+  m_groupToNode.resize(n);
   for (size_t i = 0; i < n; ++i) {
     m_nodeToGroup[i] = i;
     m_groupToNode[i].insert(i);
--- a/avogadro/core/graph.cpp
+++ b/avogadro/core/graph.cpp
@@ -19,6 +19,7 @@
 #include <avogadro/core/connectedgroup.h>
 
 #include <algorithm>
+#include <array>
 #include <cassert>
 #include <set>
 #include <stack>
--- a/avogadro/core/graph.cpp
+++ b/avogadro/core/graph.cpp
@@ -232,6 +232,12 @@ const std::vector<size_t>& Graph::neighbors(size_t index) const
   return m_adjacencyList[index];
 }
 
+const std::vector<size_t>& Graph::edges(size_t index) const
+{
+  assert(index < size());
+  return m_edgeMap[index];
+}
+
 size_t Graph::degree(size_t index) const
 {
   return neighbors(index).size();
--- a/avogadro/core/graph.cpp
+++ b/avogadro/core/graph.cpp
@@ -86,7 +86,7 @@ void Graph::removeVertex(size_t index)
     size_t affectedIndex = m_adjacencyList.size() - 1;
     for (size_t i = 0; i < m_adjacencyList[index].size(); i++) {
       size_t otherIndex = m_adjacencyList[index][i];
-      for (size_t j = 0; i < m_adjacencyList[otherIndex].size(); i++) {
+      for (size_t j = 0; j < m_adjacencyList[otherIndex].size(); j++) {
         if (m_adjacencyList[otherIndex][j] == affectedIndex)
           m_adjacencyList[otherIndex][j] = index;
       }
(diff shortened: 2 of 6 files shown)


ghutchis:
You don't need this. It may be useful to cache the bond network, but you can access it from `Molecule::graph()`:

`const auto nodes = graph.neighbors(b); // get bonded neighbors from b`

or 
`if graph.containsEdge(a, b) // do something …`

ghutchis:
It's possible you want to add some methods to the `Graph` class, but most of this looks like it's adding a map that should already be in the graph.

aerkiaga:
Yes, I agree that implementing this outside graph is ugly at the very least. I just wasn't sure how to implement edge metadata in a satisfactory way...

aerkiaga:
I find it slightly confusing that all "size" values refer to the number of vertices in the graph, when formally it should refer to the number of edges... But I understand that renaming everything to "order" at this point would break stuff heavily...

If the `Graph` class could handle bond indices, many other structures wouldn't be necessary. I'll try that now.

aerkiaga:
Okay, I'm having trouble understanding something about the graph. The adjacency list contains the list of neighbors for each vertex, by index. When any vertex is deleted via `removeVertex`, `std::vector::erase` is called and all indices higher than this one should become invalid. I'm sorry if I'm not reading into the code enough and missing something...

ghutchis:
It's been a while, so I'll admit I'm skimming the code quickly. There are two steps:

`removeEdges(index); // which removes any edges to that vertex for other atoms`

Then you remove the adjacency entry itself:
`m_adjacencyList.erase(m_adjacencyList.begin() + index);`

Everything after that vertex is moved down one entry in the `vector` so they remain correct.

aerkiaga:
Thanks for your help with the code, Mr Hutchison. I've finally found out why that particular code works, contrary to my expectation... and I don't like the reason.

Basically, `Core::Graph::removeVertex` is not being called even once, in any situation. `m_graphDirty` is true whenever `Core::Molecule::removeAtom` reaches the relevant `if` clause, probably because `Core::Molecule::removeBond` sets it. Adding `updateGraph();` right before that line causes Avogadro to crash all the time, either by `Assertion `index < size()' failed.` or straight up segfault, depending on the editing actions performed. Debugging reveals that `m_adjacencyList` contains many invalid entries, but Avogadro can crash even if this is not yet the case.

aerkiaga:
Should I fix this issue in a separate PR, or send everything as part of the changes made here?

ghutchis:
Let's deal with that in a separate PR. If those issues aren't flagged by the current tests, we need to add more tests.

aerkiaga:
Okay, so it's likely that moving the functionality introduced in this PR into the `Graph` class will uncover at least some of the bugs I talked about before. I'll fix those as I encounter them, and leave more thorough testing for another PR, as you stated.

ghutchis:
I'm a bit overloaded for the next day or two, but I'll get back to you soon. It seems like what you're describing is persistent indices for an atom / bond? If so, I'm not sure you'd need anything as complicated as what you describe. But I'll take a look and we can discuss.

ghutchis:
I'm waiting for the build / test to finish...

aerkiaga:
Weird... I ran all tests in my Ubuntu machine and they ran fine. Maybe I forgot to stage something? Let me check in a moment...

ghutchis:
Sorry, didn't mean to close this.

ghutchis:
If you can't find it, once the build starts, you can use https://github.com/mxschmitt/action-tmate to get access to the runner.

e.g. https://github.com/OpenChemistry/avogadrolibs/blob/eb54b42c0f5e087d75c09e0ba1097831f13b394c/.github/workflows/build_cmake.yml#L193

aerkiaga:
Oh, cool! I'm not familiar with CI, so thanks for all your help with these more technical things :)

ghutchis:
You should be able to check the live build here: https://github.com/OpenChemistry/avogadrolibs/runs/5652565414?check_suite_focus=true

aerkiaga:
Oh, I see... I didn't see the crash because I ran the tests through Valgrind, and that seems not to crash. It does detect some memory issues, but those were in `ConnectedGroup`, so I just assumed they were there before... Mistake on my part, I'll have a look right now and probably fix it tomorrow.

ghutchis:
Thanks - it's probably worth setting up a Valgrind CI build, but I haven't done that. I find MacOS also hides some memory bugs for some reason.

aerkiaga:
Ugh, I hate these inconsistent memory bugs, I agree a Valgrind CI action would help... Regarding the bug, it's definitely in `ConnectedGroup`; haven't dug much into the code, but by the looks of it I'm positive that this last workaround is actually helping and not just a hack to force the code into passing tests.

ghutchis:
Thanks for your persistence on this one.

PR #855, opened 2022-04-16
OpenGL: Detaching and deleting shaders/programs on updates

octycs:
Background:  
I had issues with Avogadro crashing after some minutes when I was viewing vibrational modes, and I quickly realized it was because Avogadro continually allocated more memory until my RAM+swap was full.

From what I understand, this is because each geometry update (for each animation frame) will trigger `GLWidget::updateScene()`, which clears the rendering `GroupNode` (https://github.com/OpenChemistry/avogadrolibs/blob/master/avogadro/qtopengl/glwidget.cpp#L82) and then rebuilds. During that clearing, the `ShaderProgram` objects get destroyed, however they are not being deleted from OpenGL.

Changing this alone however doesn't fix the issue, so I also added code to detach and delete the shader as soon as `glLinkProgram()` was called (Following [this reference](https://stackoverflow.com/a/43294730)). To avoid relinking the program, I needed to remove the `m_linked = false`. Until now, only `SphereAmbientOcclusionRenderer` has been using `ShaderProgram::detachShader()`, so I hope this change is safe. I only did this for `CylinderGeometry`, `SphereGeometry`, `LineStripGeometry` and `TextLabelBase`, because there I observed the issue to be significant and I could test it.

Doing this greatly reduces the continuous memory allocation, and I could not find any unexpected behaviour. However since I don't know the code well, there might be a different and better solution. Feel free to modify this PR.

I'v attached a NWChem log to test viewing molecular vibrations:
[mode.log](https://github.com/OpenChemistry/avogadrolibs/files/8500207/mode.log)

---

Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

### DIFF
--- a/avogadro/rendering/cylindergeometry.cpp
+++ b/avogadro/rendering/cylindergeometry.cpp
@@ -169,6 +169,11 @@ void CylinderGeometry::update()
     d->program.attachShader(d->fragmentShader);
     if (!d->program.link())
       cout << d->program.error() << endl;
+
+    d->program.detachShader(d->vertexShader);
+    d->program.detachShader(d->fragmentShader);
+    d->vertexShader.cleanup();
+    d->fragmentShader.cleanup();
   }
 }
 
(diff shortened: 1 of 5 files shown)


ghutchis:
This looks okay to me, but I'm going to wait to hear from @cryos since he did far more of the rendering code.

PR #857, opened 2022-04-19
Make Navigator transformations relative to center + translation (v2)

aerkiaga:
Prior to this PR, transformations are done relative to the center of the scene. This gives the expected behavior, save for translation: if the user pans the camera away from the starting position, rotation, zoom, etc. will act in a way that doesn't match mouse movements in any meaningful way.

This PR keeps track of the total translation vector and performs all operations relative to the center of the scene minus this vector, which effectively results in a reference point that follows the user, always placing itself in front of the camera. Behavior without panning is preserved, while in other cases it more closely matches user expectation.

The change is small but split into per-transformation changes. Please tell me if any of the transformations should not be given this treatment. Feel free to squash the remaining commits into a single one.

Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

### DIFF
--- a/avogadro/qtplugins/navigator/navigator.cpp
+++ b/avogadro/qtplugins/navigator/navigator.cpp
@@ -133,8 +133,7 @@ QUndoCommand* Navigator::mouseMoveEvent(QMouseEvent* e)
     case Translation: {
       Vector2f fromScreen(m_lastMousePosition.x(), m_lastMousePosition.y());
       Vector2f toScreen(e->localPos().x(), e->localPos().y());
-      Vector3f ref = m_renderer->scene().center() - m_translation;
-      translate(ref, fromScreen, toScreen);
+      translate(m_renderer->scene().center(), fromScreen, toScreen);
       e->accept();
       break;
     }
--- a/avogadro/qtplugins/navigator/navigator.cpp
+++ b/avogadro/qtplugins/navigator/navigator.cpp
@@ -126,8 +126,7 @@ QUndoCommand* Navigator::mouseMoveEvent(QMouseEvent* e)
   switch (m_currentAction) {
     case Rotation: {
       QPoint delta = e->pos() - m_lastMousePosition;
-      Vector3f ref = m_renderer->scene().center() - m_translation;
-      rotate(ref, delta.y(), delta.x(), 0);
+      rotate(m_renderer->scene().center(), delta.y(), delta.x(), 0);
       e->accept();
       break;
     }
@@ -141,11 +140,10 @@ QUndoCommand* Navigator::mouseMoveEvent(QMouseEvent* e)
     }
     case ZoomTilt: {
       QPoint delta = e->pos() - m_lastMousePosition;
-      Vector3f ref = m_renderer->scene().center() - m_translation;
       // Tilt
-      rotate(ref, 0, 0, delta.x());
+      rotate(m_renderer->scene().center(), 0, 0, delta.x());
       // Zoom
-      zoom(ref, delta.y());
+      zoom(m_renderer->scene().center(), delta.y());
       e->accept();
       break;
     }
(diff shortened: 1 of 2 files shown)


ghutchis:
I don't like it, to be honest. The previous code handled rotations around the center of the molecule. So if I translate the camera, I'm still rotating around the center of the molecule.

Now, if I translate, it makes a big "swing" for everything I've tried.

Can you record a video or something of how this is an improvement? I'm willing to consider it as an option, perhaps, but this feels unintuitive to me.

Willing to listen to counter-arguments.

aerkiaga:
I agree. Rotation behaves worse under this change. However, I'd still advocate for the new zoom behavior, which motivated this PR to a great degree.

aerkiaga:
https://user-images.githubusercontent.com/38228836/164799075-b8661797-87cb-4ec5-805c-e5823c8773da.mp4

https://user-images.githubusercontent.com/38228836/164799089-a12b9cb3-59b8-4316-9704-485c44a0f361.mp4

aerkiaga:
In fact, I'd like to put some more thought into which transformations work well relative to the molecule's center, and which are better like this.

ghutchis:
Sounds good - rotation is definitely worse. Not sure about tilt + zoom since I'm using my laptop right now.

aerkiaga:
My idea is that zooming should always move the camera forward, as its typical use case is focusing attention on a detail that the user is seeing.

Rotation is probably better as it's always been. As long as a dragging motion causes the view to feel displaced in a predictable way, it's intuitive, plus the center of rotation should be meaningful.

I'm not sure about the tilt-zoom. Since it serves a similar purpose to plain zoom, I'd also make it relative.

Making panning relative wasn't quite that important as a change in workflow, but I meant it to avoid some situations wherein the screen would move in unpredictable ways due to other transformations.

And finally, keyboard operations. I was unsure of whether to make them relative, and I'd say we shouldn't.

I'll revert the appropriate commits. Let's see if this is more or less usable...

aerkiaga:
I'm sorry, but I'm unable to get to a state wherein navigation just works for any situation. There always seem to be corner cases. Will need more investigation...

aerkiaga:
I believe `m_translation` needed to be rotated along with the camera. Now zoom works as intended by this PR (always towards the front). I'm not sure how rotation will behave now, though...

aerkiaga:
https://user-images.githubusercontent.com/38228836/164910555-c56ebb9a-4fec-4467-83c5-9ad826e960f5.mp4

What do you think of this? It should work as before except for zooming behavior, which always focuses whatever is pointed at by the camera, instead of scaling the molecule about its center.

ghutchis:
Did you intend to close this?

aerkiaga:
Yes, sorry if it comes across as weird. My motivation was that, for large molecules, the user may want to navigate relative to specific features (e.g. a ligand) rather than the center of the molecule. But now I realize this is probably not the best way to do it, as it doesn't play well with either small molecules or multiple open molecules. Better ideas for discussion would be:

- An implementation that uses an independent variable for every individual scene. Essentially this PR but better.
- A checkbox to enable it. Possibly enabling all transformation rather than just zooming (navigate around center vs. around focused feature).
- Just leaving it as-is and waiting for layers/ConnectedGroup and the ability to split ligands from their target to eventually sort the problem out. So every layer can have its own center.

ghutchis:
In Avogadro 1, we had a feature in which you could click on a particular atom and then rotation / zoom would be around that atom.

I could imagine the code looking for a selection, and using that for the centroid. So you'd select a ligand, then you can use that for the navigation (maybe with a modifier key or checkbox).

I can also imagine it would be useful in a larger structure to have a new Extension command for "zoom to selection" - translating / rotating and zooming for centering a selected ligand -- you select ARG26 and move to it then select chain 2 and move to that, etc.

aerkiaga:
Wonderful idea! It would only need to set the scene center to the appropriate value, then possibly zoom in a bit. Could be done easily.

PR #866, opened 2022-04-23
Improve Eigen CMake handling

nicolasfella:
Eigen ships a CMake config file itself, so the find module here is no longer needed

Use imported targets instead of include_directories

Add dependency to CMake config file

Link with PUBLIC in AvogadroCore so that consumers pick up the dependency automatically

Remove dependency where not used

Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.


ghutchis:
Sounds great, but the compile is failing back quickly:
```
CMake Error at avogadro/core/CMakeLists.txt:1 (find_package):
      By not providing "FindEigen3.cmake" in CMAKE_MODULE_PATH this project has
      asked CMake to find a package configuration file provided by "Eigen3", but
      CMake did not find one.
    
      Could not find a package configuration file provided by "Eigen3" with any
      of the following names:
    
        Eigen3Config.cmake
        eigen3-config.cmake
```

Keep in mind that this is built under https://github.com/OpenChemistry/openchemistry

But if you'd like to help clean up the CMake a bit, I'd be happy - it's never been my strong suit..

ghutchis:
Okay, this seems limited to the Python builds because they don't have Eigen installed before build. Seems solvable.

PR #868, opened 2022-04-24
Add plugin to focus or unfocus selection

aerkiaga:

Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.


avogadro/qtplugins/focus/focus.cpp
```diff
@@ -0,0 +1,111 @@
+/******************************************************************************
```
ghutchis, line 1: Drop all this. As we edit files, we're going to a new top comment:
```
/******************************************************************************
  This source file is part of the Avogadro project.
  This source code is released under the 3-Clause BSD License, (see "LICENSE").
******************************************************************************/
```

(In particular, you're editing this in 2022, not 2012 .. and you're not working for Kitware.)

avogadro/qtplugins/focus/focus.h
```diff
@@ -0,0 +1,71 @@
+/******************************************************************************
```
ghutchis, line 1: Please use the new header:
```
/******************************************************************************
  This source file is part of the Avogadro project.
  This source code is released under the 3-Clause BSD License, (see "LICENSE").
******************************************************************************/
```

ghutchis:
* I think it should be a bit more zoomed into the selection.

Load `1ayo` - select residue 10. Focus. Could probably be a bit more zoomed.

* It needs to call for a redraw after focus / unfocus:
```
    emit updateRequested();
```

* I also got it to totally make the molecule disappear. You have to be a bit careful with transforming the camera. Not sure if the camera matrix went sideways .. take a look at `resetview` for example.

avogadro/qtplugins/focus/focus.cpp
aerkiaga: Ooops... Yes, I'm sorry, I just copied the notice from a different file. I'm pretty stupid when it comes to legal code.

aerkiaga:
Every thing done. The disappearing issue was due to a missed initialization, which should be fixed now. The current behavior is looking at the focused point from the current camera position, then scaling the view to put it at a fixed distance from the target.

ghutchis:
Much better. One more minor thing. I think the zoom level should be estimated based on the radius / size of the selection.

Try this as an example. Download `1ayo` then select `ARG` (i.e., all the ARG residues). Some, but not a lot.

Focus. With this revision, you're zoomed in and can't even see the selection.

It's a minor thing, but I think if we get this right, it will be a very popular feature.

Issue #824, opened 2022-02-22
malloc crash when building crystal supercell


drew-parsons:
**Avogadro version:**
 - Avogadrolibs: 1.95.1
 - Qt: 5.15.2

**Desktop version:**
 - OS: Linux (Debian unstable)
 - Version linux 5.16.7-2
 - Compiler gcc 11.2.0

**Describe the bug**

I'm reading a  pdb file of a DNA or protein fragment  and trying to extend the crystal to a supercell (the aim is to extend along the longitudinal axis to get a sense of the DNA helix).

But after pressing OK in the "Build Supercell" dialog box (e.g. after selecting 2×2x2), avogadro crashes, writing to stderr
```
free(): invalid pointer
Aborted
```

**To Reproduce**
1. Open a pdb file, e.g. [2ztb.pdb](https://www.rcsb.org/structure/2ZTB) from https://www.rcsb.org/structure/2ZTB
1. Go to 'Crystal→Build Supercell'
2. Select '2' for A, B and C repeat.
3. Click 'OK'

**Expected behavior**

The fragment (unit cell) should be shown repeated according to the chosen supercell.

**Backtrace**

Running in gdb gets a backtrace:
```
$ gdb avogadro2
GNU gdb (Debian 10.1-2) 10.1.90.20210103-git
...
(gdb) run 2ztb.pdb
Starting program: /usr/bin/avogadro2 2ztb.pdb
...
Open Babel formats ready:  145
[New Thread 0x7fffb23fd640 (LWP 2288197)]
[New Thread 0x7fffb0e98640 (LWP 2288198)]
[Thread 0x7fffb23fd640 (LWP 2288197) exited]
free(): invalid pointer

Thread 1 "avogadro2" received signal SIGABRT, Aborted.
__GI_raise (sig=sig@entry=6) at ../sysdeps/unix/sysv/linux/raise.c:49
49	../sysdeps/unix/sysv/linux/raise.c: No such file or directory.
(gdb) bt
#0  __GI_raise (sig=sig@entry=6) at ../sysdeps/unix/sysv/linux/raise.c:49
#1  0x00007ffff6290546 in __GI_abort () at abort.c:79
#2  0x00007ffff62e7eb8 in __libc_message (action=action@entry=do_abort, fmt=fmt@entry=0x7ffff6405a78 "%s\n") at ../sysdeps/posix/libc_fatal.c:155
[...]

PR #806, opened 2021-10-05
Be careful about assign or copy from molecules with no layers

ghutchis:
Not sure if this is the perfect solution, but it fixes #800
Might also fix #752

Signed-off-by: Geoff Hutchison <geoff.hutchison@gmail.com>

Developer Certificate of Origin
Version 1.1

Copyright (C) 2004, 2006 The Linux Foundation and its contributors.
1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

Everyone is permitted to copy and distribute verbatim copies of this
license document, but changing it is not allowed.

Developer's Certificate of Origin 1.1

By making a contribution to this project, I certify that:

(a) The contribution was created in whole or in part by me and I
    have the right to submit it under the open source license
    indicated in the file; or

(b) The contribution is based upon previous work that, to the best
    of my knowledge, is covered under an appropriate open source
    license and I have the right under that license to submit that
    work with modifications, whether created in whole or in part
    by me, under the same open source license (unless I am
    permitted to submit under a different license), as indicated
    in the file; or

(c) The contribution was provided directly to me by some other
    person who certified (a), (b) or (c) and I have not modified
    it.

(d) I understand and agree that this project and the contribution
    are public and that a record of the contribution (including all
    personal information I submit with it, including my sign-off) is
    maintained indefinitely and may be redistributed consistent with
    this project or the open source license(s) involved.

PR #876, opened 2022-04-28
WIP to find the memory bug involved in #824

ghutchis:
Right now, this reverts #806 but there's still some sort
of memory corruption when assigning one molecule to another

Also, it seems like File -> Import Crystal adds atoms that
don't have layer information

Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

### DIFF
--- a/avogadro/core/layermanager.cpp
+++ b/avogadro/core/layermanager.cpp
@@ -26,6 +26,7 @@ Layer& LayerManager::getMoleculeLayer()
 
 Layer& LayerManager::getMoleculeLayer(const Molecule* mol)
 {
+  assert(mol != nullptr);
   auto it = m_molToInfo.find(mol);
   if (it == m_molToInfo.end()) {
     m_molToInfo[mol] = make_shared<MoleculeInfo>(mol);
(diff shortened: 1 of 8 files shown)


drew-parsons:
Seems to work more or less.  Doesn't give the memory crash, and I can see the crystal supercell is formed.

There are some other issues with the interface e.g. the ball&stick view gets randomly switched off, but there's no reason to think that's caused by this patch. Might just be the size of the sample I used to test (one protein, so the supercell has 1000s of atoms to track).

I'd say thumbs up for this patch, and can deal with any other problems as they come up if they recur.

ghutchis:
It's still related. What's happening is that some of the actions -- supercell is the easiest to find -- actually replace the molecule "under the hood."

At the moment, that sometimes replaces the layer information.

The patch is an 80% fix. What you describe is a consequence .. we're losing some of the layer settings .. but if it avoids a crash it's a big step in the right direction.

AFAICT this should be suitable for backporting to 1.95.1 LTS packages.

PR #878, opened 2022-04-29
Subgraph rewrite

aerkiaga:
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

---

The algorithm is simple: when a new atom is added, it's not assigned a subgraph quite yet. When two atoms are bonded, their subgraphs, if present, are merged. When a bond is broken, the subgraph it's in is marked dirty.

Now, when any information about subgraphs is queried, it will be returned if it refers to a clean subgraph. If it refers to a dirty subgraph, all its formed vertices are traversed and new subgraphs split from it. If it refers to a lone vertex, it will be assigned a subgraph.

This way, expensive operations are only performed once, or not at all, and any overhead is avoided.


aerkiaga:
Still WIP, as lone vertices still need to be computed when querying all subgraphs.

ghutchis (changes requested):
Mostly just asking for some doc strings.

ghutchis:
There's something else .. here's a test file:
[test.cjson.zip](https://github.com/OpenChemistry/avogadrolibs/files/8595102/test.cjson.zip)

If you double-click with the select tool, you can select the water ligands. I deleted a few.

Then I broke the graph by deleting a carbon atom in the middle and tried to double-click one of the fragments. It hung.

I don't have time to investigate further tonight, but wanted to report the bug - and then the need to add a unit test like the previous ConnectedGraph one.

aerkiaga:
I can try to fix it now. And regarding the testcase, I wanted to try to port the original testcase.

aerkiaga:
A little minimal, maybe. Should I add more testcase code?

ghutchis:
In my testing so far, this seems robust.

ghutchis:
I think for now, the test case is okay. I'll think through some more torture cases (e.g., patches of graphene)

PR #898, opened 2022-05-21
Render close contacts

aerkiaga:

Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

### DIFF
--- a/avogadro/qtplugins/closecontacts/closecontacts.cpp
+++ b/avogadro/qtplugins/closecontacts/closecontacts.cpp
@@ -32,7 +32,7 @@ using Rendering::LineStripGeometry;
 using Rendering::GeometryNode;
 using Rendering::GroupNode;
 
-CloseContacts::CloseContacts(QObject* p) : ScenePlugin(p)
+CloseContacts::CloseContacts(QObject *p) : ScenePlugin(p)
 {
   m_layerManager = PluginLayerManager(m_name);
   
@@ -42,7 +42,32 @@ CloseContacts::CloseContacts(QObject* p) : ScenePlugin(p)
 
 CloseContacts::~CloseContacts() {}
 
-void CloseContacts::process(const Molecule& molecule, Rendering::GroupNode& node)
+static bool checkPairNot1213(const Molecule &molecule, Index i, Index n)
+{
+  static Array<Index> bondedCache;
+  static Index lastIndex;
+  static bool lastIndexValid = false;
+
+  if (!lastIndexValid || lastIndex != i) {
+    bondedCache.clear();
+    for (const Bond *b : molecule.bonds(i))
+      bondedCache.push_back(b->atom1().index() == i ? b->atom2().index() : b->atom1().index());
+    lastIndex = i;
+    lastIndexValid = true;
+  }
+
+  for (const Bond *b : molecule.bonds(n)) {
+    Index m = (b->atom1().index() == n ? b->atom2() : b->atom1()).index();
+    if (m == i) // exclude 1-2 pairs
+      return false;
+    for (Index bn: bondedCache)
+      if (bn == m) // exclude 1-3 pairs
+        return false;
+  }
+  return true;
+}
+
+void CloseContacts::process(const Molecule &molecule, Rendering::GroupNode &node)
 {
   float radius(0.1f);
   Vector3ub color(128, 255, 64);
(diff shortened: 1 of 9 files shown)


avogadro/qtplugins/closecontacts/closecontacts.cpp
```diff
@@ -0,0 +1,124 @@
...
+}
+
+QWidget* CloseContacts::setupWidget()
+{
+  QWidget *widget = new QWidget(qobject_cast<QWidget*>(this->parent()));
+  QVBoxLayout *v = new QVBoxLayout;
+
+  // maximum distance
+  QDoubleSpinBox *spin = new QDoubleSpinBox;
+  spin->setRange(0.1, 10.0);
```
ghutchis, line 99: I think the minimum should be something like 1.5 maybe?

Also consider adding a suffix for the unit.
`spin->setSuffix(tr(" Å"));`

avogadro/qtplugins/closecontacts/closecontacts.cpp
```diff
@@ -0,0 +1,124 @@
...
+void CloseContacts::process(const Molecule& molecule, Rendering::GroupNode& node)
+{
+  float radius(0.1f);
+  Vector3ub color(128, 255, 64);
+
+  NeighborPerceiver perceiver(molecule.atomPositions3d(), m_maximumDistance);
+
+  GeometryNode* geometry = new GeometryNode;
+  node.addChild(geometry);
+  CylinderGeometry *cylinders = new CylinderGeometry;
```
ghutchis, line 54: Personally, I'd rather the wireframe / lines, like `wireframe.cpp`:
```
    LineStripGeometry* lines = new LineStripGeometry;
    geometry->addDrawable(lines);

    lines->setColor(m_color);
    float width = m_lineWidth;
    // etc.
    lines->addLineStrip(points, colors, width);
```

ghutchis: IMHO the dashed lines with configurable colors and/or width would look good. But then again, that's because it's what we did in Avo1.

avogadro/qtplugins/closecontacts/closecontacts.cpp
```diff
@@ -0,0 +1,124 @@
...
+  for (Index i = 0; i < molecule.atomCount(); ++i) {
+    Vector3 pos = molecule.atomPosition3d(i);
+    Array<Index> bonded;
+    for (const Bond *b : molecule.bonds(i))
+      bonded.push_back(b->atom1().index() == i ? b->atom2().index() : b->atom1().index());
+    for (Index n : perceiver.getNeighbors(pos)) {
+      if (n <= i)
+        continue;
+      bool go_on = false;
+      for (const Bond *b : molecule.bonds(n)) {
```
ghutchis, line 67: Here you're skipping neighbors that are bonded directly to i.

avogadro/qtplugins/closecontacts/closecontacts.cpp
```diff
@@ -0,0 +1,124 @@
...
+        Index m = (b->atom1().index() == n ? b->atom2() : b->atom1()).index();
+        if (m == i) {
+          go_on = true;
+          break;
+        }
+        for (Index bn: bonded) {
```
ghutchis, line 73: Here you're trying to remove atoms bonded to something bonded to i.

avogadro/qtplugins/closecontacts/closecontacts.cpp
```diff
@@ -0,0 +1,124 @@
...
+          if (bn == m) {
+            go_on = true;
+            break;
+          }
+        }
+      }
+      if (go_on)
+        continue;
+
+      Vector3 npos = molecule.atomPosition3d(n);
```
ghutchis, line 83: Otherwise, make sure that it's not too far.

I guess my question is on the ordering of the loops .. which one of these is more common in a bigger molecule.

But based on the neighborPerceiver using the max distance threshold .. will this statement ever happen?

avogadro/qtplugins/closecontacts/closecontacts.cpp
aerkiaga: I agree. Something that stands out from regular bonds as much as possible is the goal.

avogadro/qtplugins/closecontacts/closecontacts.cpp
aerkiaga: Should change that too. A reasonable option would be to add two methods: one that returns neighbors only within the specified distance, and another that uses the current behavior but has a more descriptive name.

aerkiaga: What about now? It's perceivably faster, possibly somewhat more readable.

ghutchis:
Seems like this has most / all of what you'd want complete.

I don't know what else you want to do on this one.. Maybe add color / line width options?

Adding chalcogen interactions should pretty easy .. add a filter if both atoms are in O, S, Se, Te (8, 16, 34, 52). I guess I'd default to using yellow / orange for that? (That could be a separate PR.)

The hydrogen-bond part is a bit trickier and I'd probably save that for a separate PR so we can discuss the filtering.

aerkiaga:
I agree. Let's add those features in further PRs, this one should be ready. Although as you said, there are a few options to customize here...

aerkiaga:
The MacOS build failure seems to be a spurious connection error, rather than a compiler panic or such.

PR #926, opened 2022-06-03
Implement basic hydrogen bonds

aerkiaga:
This is the general atom-pair algorithm I outlined, but with angle tolerances (the largest part) not implemented yet. That's equivalent to a 180 degree tolerance, but surprisingly already gives quite accurate results for the macromolecules I've tested, both with automatically placed and experimentally determined hydrogens.

Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

### DIFF
--- a/avogadro/qtplugins/atompairbonds/atompairbonds.cpp
+++ b/avogadro/qtplugins/atompairbonds/atompairbonds.cpp
@@ -1,0 +1,159 @@
+/******************************************************************************
+  This source file is part of the Avogadro project.
+  This source code is released under the 3-Clause BSD License, (see "LICENSE").
+******************************************************************************/
+
+#include "atompairbonds.h"
+
+#include <avogadro/core/array.h>
+#include <avogadro/core/bond.h>
+#include <avogadro/core/elements.h>
+#include <avogadro/core/neighborperceiver.h>
+#include <avogadro/qtgui/molecule.h>
+#include <avogadro/rendering/dashedlinegeometry.h>
+#include <avogadro/rendering/geometrynode.h>
+#include <avogadro/rendering/groupnode.h>
+
+#include <QtCore/QSettings>
+#include <QtWidgets/QDoubleSpinBox>
+#include <QtWidgets/QFormLayout>
+#include <QtWidgets/QVBoxLayout>
+#include <QtWidgets/QWidget>
+
+#include <iostream>
+
+namespace Avogadro {
+namespace QtPlugins {
+
+using Core::Array;
+using Core::Bond;
+using Core::NeighborPerceiver;
+using QtGui::Molecule;
+using QtGui::PluginLayerManager;
+using Rendering::DashedLineGeometry;
+using Rendering::GeometryNode;
+using Rendering::GroupNode;
+
+AtomPairBonds::AtomPairBonds(QObject *p) : ScenePlugin(p)
+{
+  m_layerManager = PluginLayerManager(m_name);
+  
+  QSettings settings;
+  m_angleToleranceDegrees = settings.value("atomPairBonds/angleTolerance", 20.0).toDouble();
+}
+
+AtomPairBonds::~AtomPairBonds() {}
+
+static const int HYDROGEN_BOND = 0;
+
+static const float MAX_DISTANCES[] = {2.0};
+
+static const float ABSOLUTE_MAX_DISTANCE = 2.0;
+
+static int getInteractionType(const Molecule &molecule, Index i) {
+  unsigned char inum = molecule.atomicNumber(i);
+  switch (inum) {
+    case 1: // hydrogen bond
+      for (const Bond *b : molecule.bonds(i)) {
+        Index j = (b->atom1().index() == i ? b->atom2() : b->atom1()).index();
+        unsigned char jnum = molecule.atomicNumber(j);
+        switch (jnum) {
+          case 7: case 8: case 9: // F, O, N
+            return HYDROGEN_BOND;
+        }
+      }
+      break;
+  }
+  return -1;
+}
+
+static bool checkPairDonorIsValid(const Molecule &molecule, Index n, int interactionType) {
+  unsigned char nnum = molecule.atomicNumber(n);
+  switch (interactionType) {
+    case HYDROGEN_BOND:
+      switch (nnum) {
+        case 7: case 8: case 9: // F, O, N
+          return true;
+      }
+      break;
+  }
+  return false;
+}
+
+void AtomPairBonds::process(const Molecule &molecule, Rendering::GroupNode &node)
+{
+  float radius(0.1f);
+  Vector3ub color(64, 192, 255);
+
+  NeighborPerceiver perceiver(molecule.atomPositions3d(), ABSOLUTE_MAX_DISTANCE);
+  std::vector<bool> isAtomEnabled(molecule.atomCount());
+  for (Index i = 0; i < molecule.atomCount(); ++i)
+    isAtomEnabled[i] = m_layerManager.atomEnabled(i);
+
+  GeometryNode *geometry = new GeometryNode;
+  node.addChild(geometry);
+  DashedLineGeometry *lines = new DashedLineGeometry;
+  lines->identifier().molecule = &molecule;
+  lines->identifier().type = Rendering::BondType;
+  lines->setLineWidth(2.0);
+  geometry->addDrawable(lines);
+  Array<Index> neighbors;
+  for (Index i = 0; i < molecule.atomCount(); ++i) {
+    if (!isAtomEnabled[i])
+      continue;
+    int interactionType = getInteractionType(molecule, i);
+    if (interactionType < 0)
+      continue;
+    Vector3 pos = molecule.atomPosition3d(i);
+    perceiver.getNeighborsInclusiveInPlace(neighbors, pos);
+    for (Index n : neighbors) {
+      if (!isAtomEnabled[n])
+        continue;
+      if (!checkPairDonorIsValid(molecule, n, interactionType))
+        continue;
+
+      Vector3 npos = molecule.atomPosition3d(n);
+      Vector3 distance_vector = npos - pos;
+
+      if (distance_vector.norm() > MAX_DISTANCES[interactionType])
+        continue;
+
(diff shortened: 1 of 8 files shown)


aerkiaga:
Or, said otherwise, this is half of the planned H-bond support code, but does the job for anyone needing it.

avogadro/qtplugins/atompairbonds/atompairbonds.cpp
```diff
@@ -0,0 +1,159 @@
...
+  QDoubleSpinBox *spin = new QDoubleSpinBox;
+  spin->setRange(0.0, 180.0);
+  spin->setSingleStep(1.0);
+  spin->setDecimals(0);
+  spin->setSuffix(tr(" °"));
+  spin->setValue(m_angleToleranceDegrees);
+  QObject::connect(spin, SIGNAL(valueChanged(double)), this,
+                   SLOT(setMaximumDistance(double)));
+  QFormLayout *form = new QFormLayout;
+  //form->addRow(QObject::tr("Angle tolerance:"), spin);
```
ghutchis, line 141: I'm confused - why don't you need this?

avogadro/qtplugins/atompairbonds/atompairbonds.cpp
```diff
@@ -0,0 +1,159 @@
...
+    }
+  }
+}
+
+QWidget *AtomPairBonds::setupWidget()
+{
+  QWidget *widget = new QWidget(qobject_cast<QWidget *>(this->parent()));
+  QVBoxLayout *v = new QVBoxLayout;
+
+  // maximum distance
```
ghutchis, line 131: This looks like the angle, not the distance... (But don't you need both?)

avogadro/qtplugins/atompairbonds/atompairbonds.cpp
```diff
@@ -0,0 +1,159 @@
...
+using Rendering::DashedLineGeometry;
+using Rendering::GeometryNode;
+using Rendering::GroupNode;
+
+AtomPairBonds::AtomPairBonds(QObject *p) : ScenePlugin(p)
+{
+  m_layerManager = PluginLayerManager(m_name);
+  
+  QSettings settings;
+  m_angleToleranceDegrees = settings.value("atomPairBonds/angleTolerance", 20.0).toDouble();
```
ghutchis, line 42: Don't you want both an angle and distance threshold?

avogadro/qtplugins/atompairbonds/atompairbonds.h
```diff
@@ -0,0 +1,55 @@
...
+  Q_OBJECT
+
+public:
+  explicit AtomPairBonds(QObject* parent = nullptr);
+  ~AtomPairBonds() override;
+
+  void process(const QtGui::Molecule& molecule,
+               Rendering::GroupNode& node) override;
+
+  QString name() const override { return tr(m_name.c_str()); }
```
ghutchis, line 29: This doesn't work. You need to use the string here, or the translation scripts won't find it to add to the template.

avogadro/qtplugins/atompairbonds/atompairbonds.cpp
ghutchis: It also seems pretty easy to add the color and line width as settings here, even if the widget doesn't implement the UI for them yet.

avogadro/qtplugins/atompairbonds/atompairbonds.cpp
```diff
@@ -0,0 +1,159 @@
...
+          return true;
+      }
+      break;
+  }
+  return false;
+}
+
+void AtomPairBonds::process(const Molecule &molecule, Rendering::GroupNode &node)
+{
+  float radius(0.1f);
```
ghutchis, line 85: Is this used at all?

avogadro/qtplugins/atompairbonds/atompairbonds.cpp
```diff
@@ -0,0 +1,159 @@
...
+{
+  m_layerManager = PluginLayerManager(m_name);
+  
+  QSettings settings;
+  m_angleToleranceDegrees = settings.value("atomPairBonds/angleTolerance", 20.0).toDouble();
+}
+
+AtomPairBonds::~AtomPairBonds() {}
+
+static const int HYDROGEN_BOND = 0;
```
ghutchis, line 47: I think you'll want an enum instead...

ghutchis (changes requested):

avogadro/qtplugins/atompairbonds/atompairbonds.cpp
aerkiaga: It has no effect at the moment and was slightly bugging me. But I guess it doesn't hurt to have it...

avogadro/qtplugins/atompairbonds/atompairbonds.cpp
aerkiaga: Yes, sorry. I've basically copied most of my earlier Close Contact code, deleted the detection function and rewritten it, so there are still some issues with naming...

avogadro/qtplugins/atompairbonds/atompairbonds.cpp
aerkiaga: Angle and distance, huh... Well, it could be more useful than an angle alone, especially if the distances don't change much across bond types.

avogadro/qtplugins/atompairbonds/atompairbonds.cpp
ghutchis: Well, it has no effect because you've set the `SLOT(setMaximumDistance(double))` which doesn't exist. Presumably you want `setAngleTolerance(double)`

You can use the newer syntax:
```
connect(spin, &valueChanged, this, &setMaximumDistance);
```

This should catch problems at compile time.

avogadro/qtplugins/atompairbonds/atompairbonds.cpp
```diff
@@ -0,0 +1,192 @@
...
+  NONE = -1,
+  HYDROGEN_BOND = 0
+};
+
+static enum InteractionTypes getInteractionType(const Molecule &molecule, Index i) {
+  unsigned char inum = molecule.atomicNumber(i);
+  switch (inum) {
+    case 1: // hydrogen bond
+      for (const Bond *b : molecule.bonds(i)) {
+        Index j = (b->atom1().index() == i ? b->atom2() : b->atom1()).index();
```
ghutchis, line 64: I tend to agree with the MISRA rule here.. I don't really like the ? operator -- it's hard to read. Can you just change this to an if/else syntax?

avogadro/qtplugins/atompairbonds/atompairbonds.cpp
```diff
@@ -0,0 +1,192 @@
...
+          return true;
+      }
+      break;
+  }
+  return false;
+}
+
+static int checkAtomPairNotBonded(const Molecule &molecule, Index i, Index n) {
+  Array<const Bond *> bonds = molecule.bonds(i);
+  return std::all_of(bonds.begin(), bonds.end(), [i, n](const Bond *b)
```
ghutchis, line 91: Please comment what this is supposed to be doing here.

avogadro/qtplugins/atompairbonds/atompairbonds.cpp
```diff
@@ -0,0 +1,192 @@
...
+  std::vector<bool> isAtomEnabled(molecule.atomCount());
+  for (Index i = 0; i < molecule.atomCount(); ++i)
+    isAtomEnabled[i] = m_layerManager.atomEnabled(i);
+
+  GeometryNode *geometry = new GeometryNode;
+  node.addChild(geometry);
+  DashedLineGeometry *lines = new DashedLineGeometry;
+  lines->identifier().molecule = &molecule;
+  lines->identifier().type = Rendering::BondType;
+  lines->setLineWidth(2.0);
```
ghutchis, line 110: Maybe m_lineWidths[0] instead?

avogadro/qtplugins/atompairbonds/atompairbonds.h
```diff
@@ -0,0 +1,60 @@
...
+  Q_OBJECT
+
+public:
+  explicit AtomPairBonds(QObject* parent = nullptr);
+  ~AtomPairBonds() override;
+
+  void process(const QtGui::Molecule& molecule,
+               Rendering::GroupNode& node) override;
+
+  QString name() const override { return tr("Atom-Pair Bonds"); }
```
ghutchis, line 30: I'm tempted to call it "Hydrogen Bonds" or "Non-Covalent" maybe.

ghutchis:
Code review aside, initial play seems to indicate it works fairly well. I'm sure some might want more careful filtering, but this is definitely a good place to start.

avogadro/qtplugins/atompairbonds/atompairbonds.h
aerkiaga: "Non-covalent" seems good. The idea is that the same plugin should handle halogen and chalcogen bonds too. All of these have in common the fact they involve an atom and an electron pair from another atom, unlike salt bridges etc., but to my knowledge no one calls them "Atom-pair bonds", so I wholeheartedly agree.

Should I change all references to atom-pair interactions in the code, or just the name?

avogadro/qtplugins/atompairbonds/atompairbonds.h
```diff
@@ -27,11 +27,11 @@ class AtomPairBonds : public QtGui::ScenePlugin
   void process(const QtGui::Molecule& molecule,
                Rendering::GroupNode& node) override;
 
-  QString name() const override { return tr("Atom-Pair Bonds"); }
+  QString name() const override { return tr("Non-Covalent Bonds"); }
```
ghutchis, line 30: Let's just stick with "Non-Covalent" and leave off the bonds. Lots of debates about what is a bond and what is not a bond. 😄

ghutchis:
I'm conflicted about the code-scanning bot. If I could just figure out how to turn off "MISRA" rules, I'd be pretty happy.

aerkiaga:
At least I see the number of detected issues is going down :)

PR #930, opened 2022-06-06
GAMESS-US and NWChem output reader sanity check

TiborGY:
The GAMESS-US and NWChem-log output readers now raise an error if they could not read any atomic coordinates from the output. Also make the return value nodiscard to force callers to check if the read actually succeeded, as these functions also alter the state of a Molecule object if the read succeeds differently than in case of a failure.

Partially resolves issue 2. in https://discuss.avogadro.cc/t/molpro-import-is-still-regressed-in-avogadro2-1-96-0-compared-to-avogadro1/3855

Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

### DIFF
--- a/avogadro/quantumio/gamessus.cpp
+++ b/avogadro/quantumio/gamessus.cpp
@@ -60,7 +60,7 @@ std::vector<std::string> GAMESSUSOutput::mimeTypes() const
   return std::vector<std::string>();
 }
 
-bool GAMESSUSOutput::read(std::istream& in, Core::Molecule& molecule)
+[[nodiscard]] bool GAMESSUSOutput::read(std::istream& in, Core::Molecule& molecule)
 {
   // Read the log file line by line, most sections are terminated by an empty
   // line, so they should be retained.
@@ -99,7 +99,11 @@ bool GAMESSUSOutput::read(std::istream& in, Core::Molecule& molecule)
       readEigenvectors(in);
     }
   }
-
+  if (!atomsRead){
+    appendError("Could not find any atomic coordinates! Are you sure this is a GAMESS-US output file?");
+    return false;
+  }
+  
   // f functions and beyond need to be reordered
   reorderMOs();
 
--- a/avogadro/quantumio/gamessus.cpp
+++ b/avogadro/quantumio/gamessus.cpp
@@ -60,7 +60,7 @@ std::vector<std::string> GAMESSUSOutput::mimeTypes() const
   return std::vector<std::string>();
 }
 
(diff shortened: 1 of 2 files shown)


avogadro/quantumio/nwchemlog.cpp
```diff
@@ -56,12 +56,16 @@ std::vector<std::string> NWChemLog::mimeTypes() const
   return std::vector<std::string>();
 }
 
-bool NWChemLog::read(std::istream& in, Core::Molecule& molecule)
+[[nodiscard]] bool NWChemLog::read(std::istream& in, Core::Molecule& molecule)
```
ghutchis, line 59: For now, we're sticking to C++ 11, thanks.

avogadro/quantumio/gamessus.cpp
```diff
@@ -60,7 +60,7 @@ std::vector<std::string> GAMESSUSOutput::mimeTypes() const
   return std::vector<std::string>();
 }
 
-bool GAMESSUSOutput::read(std::istream& in, Core::Molecule& molecule)
+[[nodiscard]] bool GAMESSUSOutput::read(std::istream& in, Core::Molecule& molecule)
```
ghutchis, line 63: For now, please stick to C++11, thanks.

ghutchis:
Thanks, I think this is a good step, I'll comment on the forum in a bit.

avogadro/quantumio/nwchemlog.cpp
TiborGY: Contributing.MD says C++17 is OK to use, but alright.

TiborGY: Nodiscard removed.

avogadro/quantumio/gamessus.cpp
TiborGY: Nodiscard removed.

TiborGY:
I have tested the Win64 build from the CI, and the warning message is triggered as expected when trying to read a Molpro output with either of these readers.

ghutchis:
Thanks - does it go back to the "choose a reader" dialog?

TiborGY:
> Thanks - does it go back to the "choose a reader" dialog?

Nope. As far as I can tell it cancels the entire "file open procedure".

ghutchis:
Okay, I'll take a look later tonight. Thanks.

ghutchis:
One other minor thing. We don't require copyright assignments, but we do ask contributors to sign off their commits with full names:

https://github.com/OpenChemistry/avogadrolibs/pull/930/checks?check_run_id=6761031592

Thanks!

PR #972, opened 2022-06-23
Implement Solvent Excluded Surfaces

aerkiaga:
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

### DIFF
--- a/avogadro/qtplugins/surfaces/surfaces.cpp
+++ b/avogadro/qtplugins/surfaces/surfaces.cpp
@@ -40,6 +40,7 @@ namespace {
 #include <avogadro/quantumio/nwchemlog.h>
 
 #include <QtConcurrent/QtConcurrentMap>
+#include <QtConcurrent/QtConcurrentRun>
 #include <QtCore/QBuffer>
 #include <QtCore/QCoreApplication>
 #include <QtCore/QDebug>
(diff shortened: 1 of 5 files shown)


ghutchis:
It seems like this is part of #972 -- I'm just going to review there.

avogadro/qtplugins/surfaces/surfaces.cpp
```diff
@@ -183,69 +186,149 @@ void Surfaces::calculateSurface()
...
+      (*atoms)[i].second = Core::Elements::radiusVDW(m_molecule->atomicNumber(i)) + probeRadius;
+      if ((*atoms)[i].second > max_radius)
+        max_radius = (*atoms)[i].second;
     }
 
-    m_cube->setValue(in(0), in(1), in(2), minDistance);
+    double padding = max_radius + probeRadius;
+    m_cube->setLimits(*m_molecule, m_dialog->resolution(), padding);
+    Vector3i size = m_cube->dimensions();
+    for (int z = 0; z < size(2); z++)
```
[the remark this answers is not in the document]
ghutchis: A related method in `Cube` could be useful for filling regions (e.g., the low resolution algorithm). Instead of the triple loop here, such a method could use a double loop with `std::fill` on the underlying `m_data` vector.

avogadro/qtplugins/surfaces/surfaces.cpp
```diff
@@ -183,69 +186,149 @@ void Surfaces::calculateSurface()
...
+
+    // then, for each atom, set cubes around it up to a certain radius
+    QFuture innerFuture = QtConcurrent::map(*atoms, [=](std::pair<Vector3, double> &in) {
+      double startPosZ = in.first(2) - in.second;
+      double endPosZ = in.first(2) + in.second;
+      int startIndexZ = (startPosZ - min(2)) / res;
+      int endIndexZ = (endPosZ - min(2)) / res + 1;
+      for (int indexZ = startIndexZ; indexZ < endIndexZ; indexZ++) {
+        double posZ = indexZ * res + min(2);
+        double radiusZsq = pow(in.second, 2) - pow(posZ - in.first(2), 2);
```
ghutchis, line 232: I'm pretty sure that `in.second**in.second` is faster than `std:pow()`

Sometimes I've defined a macro:
`#define SQUARE(x) ((x)*(x))`

avogadro/qtplugins/surfaces/surfaces.cpp
```diff
@@ -183,69 +186,149 @@ void Surfaces::calculateSurface()
...
+        if (radiusZsq < 0.0)
+          continue;
+        double radiusZ = sqrt(radiusZsq);
+        double startPosY = in.first(1) - radiusZ;
+        double endPosY = in.first(1) + radiusZ;
+        int startIndexY = (startPosY - min(1)) / res;
+        int endIndexY = (endPosY - min(1)) / res + 1;
+        for (int indexY = startIndexY; indexY < endIndexY; indexY++) {
+          double posY = indexY * res + min(1);
+          double lengthZYsq = pow(radiusZ, 2) - pow(posY - in.first(1), 2);
```
ghutchis, line 242: See above.

avogadro/qtplugins/surfaces/surfaces.cpp
```diff
@@ -183,69 +186,149 @@ void Surfaces::calculateSurface()
...
+          if (lengthZYsq < 0.0)
+            continue;
+          double lengthZY = sqrt(lengthZYsq);
+          double startPosX = in.first(0) - lengthZY;
+          double endPosX = in.first(0) + lengthZY;
+          int startIndexX = (startPosX - min(0)) / res;
+          int endIndexX = (endPosX - min(0)) / res + 1;
+          for (int indexX = startIndexX; indexX < endIndexX; indexX++) {
```
ghutchis, line 250: I don't remember whether the `Cube` class stores strides in x, y, z.. (And it *definitely* needs documentation / comments on that.)

But this could also be a case for a `Cube::fill()` if, like I assume, changes in `x` values are contiguous.

If instead, it's ordered differently, ignore this comment.
(I don't have time tonight to skim the code..)

avogadro/qtplugins/surfaces/surfaces.h
```diff
@@ -94,7 +95,9 @@ private slots:
 
   Core::Cube* m_cube = nullptr;
   std::vector<Core::Cube*> m_cubes;
-  QFutureWatcher<void> m_cubesWatcher;
+  QFutureWatcher<void> m_performEDTStepWatcher;
```
ghutchis, line 98: I would add comments on what each of these do. Otherwise, you find yourself coming back 2 months or 2 years later and have to re-read your code.

I tend to *over-comment* things. 😄

ghutchis (changes requested):
A few things. I'd check out the underlying `Cube` class and consider whether you can replace some of these loops with `fill()` methods. Certainly there should be a default `fill()` (no parameters) that would be much faster than your loop.

These are minor optimizations, but probably still useful for larger / higher-resolution surfaces.

avogadro/qtplugins/surfaces/surfaces.cpp
aerkiaga: I believe the fastest way to do this on all compilers is to define an inline function, as in some cases a macro could lead to its arguments being evaluated twice, and `pow(x, 2)` is optimized into a register multiply on GCC (`-O1`) and Clang (`-O1`), but apparently not MSVC.

avogadro/qtplugins/surfaces/surfaces.cpp
aerkiaga: It looks like `Cube` stores values contiguous in `z`, so I could reorder those loops and use an overloaded `Cube::fill()` as you mention.

avogadro/qtplugins/surfaces/surfaces.h
aerkiaga: Done!

ghutchis:
I think this is a good step.

PR #1019, opened 2022-07-12
Set Version info for cmake project

usta:
Add missing Version info part into CMake project

Signed-off-by: Ömer Fadıl USTA <omerusta@gmail.com>

Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,6 +1,12 @@
 cmake_minimum_required(VERSION 3.3 FATAL_ERROR)
+# Set up our version.
+set(AvogadroLibs_VERSION_MAJOR   "1")
+set(AvogadroLibs_VERSION_MINOR   "96")
+set(AvogadroLibs_VERSION_PATCH   "0")
+set(AvogadroLibs_VERSION
+  "${AvogadroLibs_VERSION_MAJOR}.${AvogadroLibs_VERSION_MINOR}.${AvogadroLibs_VERSION_PATCH}")
 
-project(AvogadroLibs)
+project(AvogadroLibs VERSION ${AvogadroLibs_VERSION})
 
 set(CMAKE_MODULE_PATH ${CMAKE_CURRENT_SOURCE_DIR}/cmake)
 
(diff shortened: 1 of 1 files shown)


ghutchis:
Can you clarify the exact warning? 

If you'd like to take a crack at `cmake` warnings, there are complaints for `spglib` and `libarchive` that I haven't had time to track down, e.g.

```
CMake Warning (dev) at /opt/homebrew/Cellar/cmake/3.23.2/share/cmake/Modules/FindPackageHandleStandardArgs.cmake:438 (message):
  The package name passed to `find_package_handle_standard_args` (SPGLIB)
  does not match the name of the calling package (Spglib).  This can lead to
  problems in calling code that expects `find_package` result variables
  (e.g., `_FOUND`) to follow a certain pattern.
Call Stack (most recent call first):
  cmake/FindSpglib.cmake:15 (find_package_handle_standard_args)
  avogadro/core/CMakeLists.txt:7 (find_package)
```

The files to tweak are in `cmake/FindSpglib.cmake:7` and `cmake/FindLibArchive.cmake:17`

usta:
@ghutchis the warning which you mentioned is related to name of the package it is searching. I mean
for example if i is coded like SpgLib or SPGLIB and if you use it like find_package(Spglib) it gives warning
like this but this is not related to this PR/MR

I have added Version tag in this MR/PR to overcome this warning :
`
CMake Deprecation Warning at CMakeLists.txt:1 (cmake_minimum_required):
  Compatibility with CMake < 2.8.12 will be removed from a future version of
  CMake.
  Update the VERSION argument <min> value or use a ...<max> suffix to tell
  CMake that the project does not need compatibility with older versions.
`
which doesnt affected but still this patch is a good addition according to current cmake projects
so it might not fix anything [for now] but let it solve upcomminng versions' problems

ghutchis:
The warning you see is from the `msym` dependency which we use, which declares a 2.8 minimum CMake version.
https://github.com/mcodev31/libmsym/blob/master/CMakeLists.txt

You can either contribute a patch to libmsym, or modify the bits here to patch the CMakeLists (or better, both):
https://github.com/OpenChemistry/openchemistry/blob/master/cmake/External_libmsym.cmake

There are some bits about the patch step here:
https://cmake.org/cmake/help/latest/module/ExternalProject.html

And there's an example for `libxml2` in which we previously had to patch on Windows:
https://github.com/OpenChemistry/openchemistry/blob/1290d91e142de0b56f7a81a3c1765d5937986ab4/cmake/External_libxml2.cmake#L29

(Basically, we can just copy a new CMakeLists file into place.)

ghutchis:
I'm going to close this. I appreciate the contribution, but it isn't needed and we already set the versioning.

Thanks, and I hope you'll contribute more in the future.

PR #1020, opened 2022-07-13
Workaround charge models locking up or not running

aerkiaga:

Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

### DIFF
--- a/avogadro/core/molecule.cpp
+++ b/avogadro/core/molecule.cpp
@@ -258,6 +258,7 @@ MatrixX Molecule::partialCharges(const std::string& type) const
     return search->second; // value from the map
   } else {
     MatrixX charges(atomCount(), 1);
+    charges.fill(0.0);
     return charges;
   }
 }
--- a/avogadro/qtplugins/openbabel/obcharges.cpp
+++ b/avogadro/qtplugins/openbabel/obcharges.cpp
@@ -198,6 +198,10 @@ MatrixX OBCharges::partialCharges(Core::Molecule& molecule) const
     charges(i, 0) = output[i];
   }
 
+  // workaround failed runs causing the code to freeze
+  if (abs(charges(0, 0)) < 0.00001)
+    charges(0, 0) = 0.0001;
+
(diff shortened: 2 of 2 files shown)


avogadro/core/molecule.cpp
```diff
@@ -258,6 +258,7 @@ MatrixX Molecule::partialCharges(const std::string& type) const
     return search->second; // value from the map
   } else {
     MatrixX charges(atomCount(), 1);
+    charges.fill(0.0);
```
ghutchis, line 261: I *think* you should be able to do:
```
MatrixX charges(atomCount(), 1, 0.0);
```

Otherwise, this is fine.

avogadro/qtplugins/openbabel/obcharges.cpp
```diff
@@ -198,6 +198,10 @@ MatrixX OBCharges::partialCharges(Core::Molecule& molecule) const
     charges(i, 0) = output[i];
   }
 
+  // workaround failed runs causing the code to freeze
+  if (abs(charges(0, 0)) < 0.00001)
```
ghutchis, line 202: This part, I'm not sure I understand.

aerkiaga: Yes, exactly. In that case it would just freeze, this gets it to finish as soon as possible.

avogadro/core/molecule.cpp
aerkiaga: Tried it now. Seems to only work for sized vectors.

avogadro/qtplugins/openbabel/obcharges.cpp
aerkiaga: I'm sure there are better ways to do it, but this one is quite simple.

ghutchis: Okay, I'll merge for now while I think about a better solution.

PR #1021, opened 2022-07-15
Initial salt bridge implementation

aerkiaga:
This displays interactions between oppositely charged atoms, as well as same-charge ("bad contacts"). Charged residues with no real formal charge in the molecule are also interpreted as charged.

Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

### DIFF
--- a/avogadro/qtplugins/closecontacts/closecontacts.cpp
+++ b/avogadro/qtplugins/closecontacts/closecontacts.cpp
@@ -9,6 +9,7 @@
 #include <avogadro/core/bond.h>
 #include <avogadro/core/elements.h>
 #include <avogadro/core/neighborperceiver.h>
+#include <avogadro/core/residue.h>
 #include <avogadro/qtgui/molecule.h>
 #include <avogadro/rendering/dashedlinegeometry.h>
 #include <avogadro/rendering/geometrynode.h>
@@ -70,6 +71,7 @@ void CloseContacts::process(const Molecule &molecule, Rendering::GroupNode &node
 {
   Vector3ub color(128, 128, 128);
 
+  //Add general contacts
   NeighborPerceiver perceiver(molecule.atomPositions3d(), m_maximumDistance);
   std::vector<bool> isAtomEnabled(molecule.atomCount());
   for (Index i = 0; i < molecule.atomCount(); ++i)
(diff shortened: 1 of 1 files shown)


avogadro/qtplugins/closecontacts/closecontacts.cpp
```diff
@@ -102,6 +104,91 @@ void CloseContacts::process(const Molecule &molecule, Rendering::GroupNode &node
...
+      residues.push_back(Index(0) - 1);
+    }
+  }
+
+  // Add predicted charged atoms from residues
+  for (const auto &r: molecule.residues()) {
+    for (const auto &a: r.residueAtoms()) {
+      if (molecule.formalCharge(a.index()) != 0)
+        continue;
+      bool gammaOxygen = false;
```
ghutchis, line 127: Before you do this, don't you want to check the atom name to ignore "N" "CA" "C" and "O" atoms?

(It might be worth having `Residue::isBackboneAtom()` and `Residue::isSideChainAtom()` methods.

avogadro/qtplugins/closecontacts/closecontacts.cpp
```diff
@@ -102,6 +104,91 @@ void CloseContacts::process(const Molecule &molecule, Rendering::GroupNode &node
...
+        continue;
+      positions.push_back(molecule.atomPosition3d(i));
+      charges.push_back(molecule.formalCharge(i));
+      residues.push_back(Index(0) - 1);
+    }
+  }
+
+  // Add predicted charged atoms from residues
+  for (const auto &r: molecule.residues()) {
+    for (const auto &a: r.residueAtoms()) {
```
ghutchis, line 124: Do you want to filter by residue? For example, this isn't important for `ALA` or `GLY` right?

avogadro/qtplugins/closecontacts/closecontacts.cpp
```diff
@@ -102,6 +104,91 @@ void CloseContacts::process(const Molecule &molecule, Rendering::GroupNode &node
...
+      if (molecule.formalCharge(a.index()) != 0)
+        continue;
+      bool gammaOxygen = false;
+      bool doubleBond = false;
+      bool alphaDoubleBond = false;
+      auto bonds = molecule.getAtomBonds(a.index());
+      auto orders = molecule.getAtomOrders(a.index());
+      for (Index i = 0; i < bonds.size(); i++) {
```
ghutchis, line 132: Please add some comments about what you're doing here. You're trying to find particular atoms, right?

Do you want to look at residue atom names? `Residue::atomNameMap()` for example.

Check out LigandExpo:
- http://ligand-expo.rcsb.org/ld-search.html
- e.g. http://ligand-expo.rcsb.org/reports/H/HIS/HIS_D3L3.gif

Now some files will come in with problematic labels, but definitely check atom names first, because they'll save a lot of time.

ghutchis (changes requested):
Some of this seems really complicated when you can check the atom names...

ghutchis:
I don't think any of this code is *bad* but I think you should check the atom names (and/or add some methods to `Residue` for example to get the name of a particular atom) to skip this logic if it's available.

aerkiaga:
I've replaced all the code with one that checks for atom names.

ghutchis:
I think a few of those should be fractional charges, but for the salt bridge visualization, the magnitude doesn't matter as far as I can tell, just + / - vs. + / +…

So for now, it looks good to me. We'll make sure it compiles cleanly and merge.

PR #1123, opened 2022-08-22
Screen space shading

aerkiaga:
At the moment, it just applies SSAO and edge detection to spheres, cylinders and opaque meshes, so as to test them. Some changes still pending:
* Extend it to other geometry types.
* Fix bugs, in particular SSAO working weird in perspective projection.
* Add dialog to configure rendering.

Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.


ghutchis:
On Mac, for some reason, the build now can't find GLEW. Not sure if that's this patch, because it doesn't seem to affect any of that.

avogadro/rendering/solid_first_fs.glsl
```diff
@@ -0,0 +1,97 @@
...
+  vec3 r = vec3(xdelta, ydelta, 2.0 * sampleStep);
+  return normalize(r);
+}
+
+float lerp(float a, float b, float f)
+{
+    return a + f * (b - a);
+}
+
+const vec2 SSAOkernel[16] = vec2[16](
```
ghutchis, line 46: I'm a little confused by the small magic numbers that are then multiplied by 10.0 in the code?

Wouldn't it be easier to just multiply this all by 10.0 here, e.g.
`vec2(-5.3e-3, 5.0e-3)` etc.?

avogadro/rendering/solid_first_fs.glsl
```diff
@@ -0,0 +1,97 @@
...
+float computeEdgeLuminosity(vec3 normal)
+{
+  return max(0.0, pow(normal.z - 0.1, 1.0 / 3.0));
+}
+
+void main()
+{
+  vec3 normal = getNormalAt(UV);
+
+  float luminosity = 1.0;
```
ghutchis, line 91: For what it's worth, I've been trying 1.15 or 1.2, which seem a bit closer to the previous illumination.

avogadro/rendering/solidpipeline.cpp
```diff
@@ -0,0 +1,132 @@
...
+  glBindTexture(GL_TEXTURE_2D, *texDepth);
+  glTexParameteri(GL_TEXTURE_2D, GL_TEXTURE_MAG_FILTER, GL_NEAREST);
+  glTexParameteri(GL_TEXTURE_2D, GL_TEXTURE_MIN_FILTER, GL_NEAREST);
+  glFramebufferTexture2D(GL_FRAMEBUFFER, GL_DEPTH_ATTACHMENT, GL_TEXTURE_2D, *texDepth, 0);
+}
+
+void SolidPipeline::initialize()
+{
+  initializeFramebuffer(&m_renderFBO, &m_renderTexture, &m_depthTexture);
+  resize(1024, 768);
```
ghutchis, line 41: Do we want this to be adjustable from the widget sizes or for AA?

avogadro/rendering/solidpipeline.cpp
```diff
@@ -0,0 +1,132 @@
...
+  GLenum drawBuffersList[1] = {GL_COLOR_ATTACHMENT0};
+  glDrawBuffers(1, drawBuffersList);
+
+  GLfloat tmp[5];
+  glGetFloatv(GL_COLOR_CLEAR_VALUE, tmp);
+  glGetFloatv(GL_DEPTH_CLEAR_VALUE, tmp + 4);
+  glClearColor(0.0, 0.0, 0.0, 0.0);
+  glClear(GL_COLOR_BUFFER_BIT | GL_DEPTH_BUFFER_BIT);
+  glClearColor(tmp[0], tmp[1], tmp[2], tmp[3]);
+  glClearDepthf(tmp[4]);
```
ghutchis, line 76: This line crashes on Mac - it looks like it's de-referencing a null pointer. If it's commented out, it otherwise seems to run. (This is perhaps because we're not using the 3.2 core profile.)

avogadro/rendering/solidpipeline.cpp
aerkiaga: I'm not sure if that line is needed, I might remove it as well...

avogadro/rendering/solidpipeline.cpp
aerkiaga: Yep, sorry, `glClearDepthf` is GL 4.1+ only, should use `glClearDepth` instead.

ghutchis:
This didn't actually work, because the `SolidPipeline::setPixelRatio()` is called *after* `SolidPipeline::resize()`

```
 solid pipeline resize 1
 resize ratio  2
```

The key is to re-order the events in `GLWidget::resizeGL(int width_, int height_)` - set the ratio and then resize.

ghutchis:
You can also disable the "VdW AO" plugin in `qtplugins/CMakeLists.txt`

```
# redundant since there's universal AO
# add_subdirectory(vanderwaalsao)
```

ghutchis:
All the compiler failures are coming from VTK enabled builds.
```
error: #error gl.h included before glew.h
```

Seems to be because in `vtk/vtkglwidget.h` the `<avogadro/rendering/glrenderer.h>` is `#included` on line 16 (i.e., after VTK imports `gl.h`)

If you move these to line 10 or 11, you should be okay:
```
#include "avogadrovtkexport.h"

#include <avogadro/qtgui/scenepluginmodel.h>
#include <avogadro/rendering/glrenderer.h>
```

ghutchis:
Looks like code that includes rendering / GL needs to update their CMakeLists to check for GLEW as well.

[link.patch.gz](https://github.com/OpenChemistry/avogadrolibs/files/9403492/link.patch.gz)

ghutchis:
Oops, missed one more VTK-related patch.

[vtk-com.patch.gz](https://github.com/OpenChemistry/avogadrolibs/files/9409547/vtk-com.patch.gz)

ghutchis:
I don't want to squish all of this into one commit, so I opened #1135 in which I squished a few of the commits (e.g., minor fixups).

PR #1228, opened 2023-04-02
CMake VTK updates

cryos:


avogadro/vtk/vtkglwidget.cpp
```diff
@@ -50,7 +50,7 @@ vtkImageData* createCubeImageData(Core::Cube* cube)
 
   data->AllocateScalars(VTK_FLOAT, 1);
 
-  auto* dataPtr = static_cast<double*>(data->GetScalarPointer());
+  auto* dataPtr = static_cast<float*>(data->GetScalarPointer());
```
cryos, line 53: I see commit 826904ae7391b842a167a474f5003587e5c6027c broke this where the committer changed two of the three types, it likely never worked and simply was not tested before committing in July 2022.

cryos:
I found 826904ae7391b842a167a474f5003587e5c6027c where the breakage happened, sweeping change switching to float from double where I guess this was never tested (July 2022). Fixed now, and it explains how it happened, `git log -p path/to/file` helped me see ;-)

Issue #1107, opened 2022-08-16
VTK plots do not work


ghutchis:
Open a crystal (either File => Import => Crystal or Open a CIF, etc.)

Go to Crystal => Plot PDF

No plot window appears .. the console prints:
2022-08-16 19:31:32.826 ( 137.243s) [         3F11AFB] vtkContextDevice2D.cxx:32    WARN| Error: no override found for 'vtkContextDevice2D'.

cryos:
I just see a vtkMath issue now,

2023-04-02 16:08:50.516 ( 181.482s) [        4F5E4D40]            vtkMath.cxx:799   WARN| vtkMath::Jacobi: Error extracting eigenfunctions

That is clicking on Crystal -> Plot Pair Distribution Function and accepting the defaults. I am able to view vtkCharts in the 3D rendering view when editing color maps etc. So I think it is built correctly with my latest updates.

ghutchis:
I can probably track that part down .. but I don't see anything when I choose the Plot PDF command. Shouldn't there be a new window?

cryos:
I can take another look at where that is failing, I am not very familiar with the PDF plotting. Let me dig into that part of it and see what I can discover. I would like to revisit some of this code, and am looking at VTK-based plots in C++/Qt for another project too.

cryos:
The no override warning you saw happens when the implementation library wasn't linked properly, I don't see that warning and it is more at the build level. Seeing charts in the other dialog means that VTK is built and linked correctly, at least in that plugin. I just need to look into what this plugin is up to.

cryos:
OK, I spent a while looking at this, it is really the fact that we are using a widget as a window, and relying on the render window width/height. I will make a PR that is just a few lines of code so that you can test on Mac, but I would like to address this properly soon.

cryos:
The immediate issue has been resolved, marking this as closed. More to come with charts.

PR #1231, opened 2023-04-04
Ensure the VTK plot widget has a minimum size

cryos:
This is very hackish, but it allows you to see the plot. The main issue is that this should have always been in a dialog or similar with a layout. The way that VTK interacts with Qt has changed and so the size being set does not appear to act as it once did. Issue #1107.


ghutchis:
I'll take a look in the morning - makes sense. It would probably be good to build up a full window class (e.g., save data to CSV / TSV, save images, etc.) anyway.

cryos:
It is a horrible fix but I am glad it gets things working again. I will see about getting a nicer version soon!

PR #1248, opened 2023-04-15
Port the OpenBabel plugin to work with Qt 6

cryos:
This is only a partial success, there is quite a bit of QRegExp use that will need more work and so this relies on Qt6::Core5Compat to allow the continued use of the QRegExp in places. Some trivial use had to migrate to work with the QString class.

There are also some changes in signal names in QProcess, and the migration from QMap to QMultiMap. The toSet() function was also removed, so I added a simple lambda to replace it in the places it was used. Tested on Qt 5.15 and Qt 6.5 where both worked in my testing.


ghutchis:
I think it looks good - we need to run `clang-format` over everything .. there are a bunch of files with changes that need fixes, but that's not critical.

ghutchis (approved):

PR #1250, opened 2023-04-16
ImportPQR should not delete the molecule!

cryos:


cryos:
This is a simple removal of a couple of things that shouldn't have been deleted. Looks good, merging as it fixes a segfault on close once the plugin dtors are actually called.

PR #1251, opened 2023-04-16
Ignore false positives from overflow multiplication

ghutchis:


ghutchis:
Closing this, there's something weird and it's not worth the time to fix the minor issue.

PR #1256, opened 2023-04-16
Remove the MongoChem plugin from the project

cryos:
This is no longer relevant, there were some great ideas that we worked on but there is no point to keeping this in the repository. It will be in git history if anyone wants to revisit old ideas.


ghutchis:
If we get a repository server again...

PR #1264, opened 2023-04-18
CMake modernization

cryos:


cryos:
OK - these are failing on the application failing to link when I spot check, looks like the libraries built fine. Merging and then moving the superbuild forward.

PR #1267, opened 2023-04-19
Modernize Eigen3, GLEW and OpenGL use as imported targets

cryos:
Eigen3 is a header only template library, adopt an updated FindEigen.cmake that creates an imported target. Note that include directories from imported targets are automatically treated as system includes on compilers that support the concept.

The core header only Avogadro library exposes it as part of its interface, and then we are able to remove a huge amount of duplicated code. Still need to look at the other side to ensure the exported targets find all targets like Eigen they depend upon.


ghutchis:
Looks like the wheel builds are failing due to use of older Cmake:
```
    CMake Error at CMakeLists.txt:1 (cmake_minimum_required):
      CMake 3.23 or higher is required.  You are running version 3.19.4
```

cryos:
Thanks, I am going to look at the wheels in a separate PR, but I wanted to separate that work out. The cibuildwheel needs updating quite a bit as do the Python versions.

PR #1268, opened 2023-04-20
Imported targets

cryos:


cryos:
The MSVC failure looks spurious as I don't even see the avogadrolibs build, I will check on follow up PRs but merging this. There are some great simplications in the build system but it is possible I missed something, Windows static GLEW is not something I think we use/support and that is the main thing that likely would not work (or static GLEW in general) with the new find code as I punted on the static imported library question for now.

PR #1270, opened 2023-04-22
Python build and wheels

cryos:
Just to note right now I am fixing up the Python build, but the wheels will need more updates to the manylinux base images, there was a custom Linux image, just digging into some of those details. This does allow a development build to compile/work with Python now and supports build in a subdirectory of the source tree.

### DIFF
--- a/.github/workflows/build_wheels.yml
+++ b/.github/workflows/build_wheels.yml
@@ -4,16 +4,15 @@ on: [push, pull_request]
 
 env:
   # Don't build python 2.7, pypy, or 32-bit wheels
-  CIBW_SKIP: "cp27-* pp* *-manylinux_i686 *-win32"
-
-  # This has some of the software we need pre-installed on it
-  CIBW_MANYLINUX_X86_64_IMAGE: openchemistry/avogadro_manylinux2010_x86_64
+  CIBW_SKIP: "cp36-* cp37-* cp311-* pp* *-musllinux_* *-manylinux_i686 *-win32"
 
   # Need to do some setup before repairing the wheel on linux...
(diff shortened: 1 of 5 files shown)


ghutchis:
You're getting errors from the CMake version on the build servers:
```
    CMake Error at CMakeLists.txt:1 (cmake_minimum_required):
      CMake 3.23 or higher is required.  You are running version 3.19.4
```

ghutchis:
Considering that was a release in March 2022, it's not on Ubuntu 20.04. It looks like Ubuntu 22.04 only has CMake 3.22.

If you need 3.23, you'll need to make sure it's installed on the build servers.

ghutchis:
Windows build complains:
```
      Could NOT find Eigen3 (missing: Eigen3_INCLUDE_DIR)
```

cryos:
I wasn't trying to fix the wheels build here, the actual build happens inside the manylinux container, not the host OS. It is that container that needs an updated CMake. I was more interested in verifying no other side effects.

cryos:
https://github.com/OpenChemistry/avogadrolibs/blob/master/scripts/github-actions/docker/Dockerfile is where the CMake version is coming from, just using `act` to test out simply installing the deps in the 2014 base image, it is also not determining the version in the base image due to git missing possibly. I have it building locally, iterating locally is faster that waiting on the actions here.

cryos:
I got Linux working, I think Windows is just the update to be consistent on case for Eigen but there could be other things lurking. Mac looks like it built but it wasn't finding the libraries. So two out of three ain't bad? Out of tokens and everything is queued...

PR #1286, opened 2023-05-01
CMake: Allow using externally-provided libraries

### LINKED ISSUE #402: Please add cmake variables USE_EXTERNAL_xx for projects (xx) that may exist as separate packages
Particularly, ```jsoncpp``` has a package.
```libgwavi``` also has a package on Arch Linux. I can easily create one on FreeBSD.
In general, external package use should be preferred.

berquist:
Closes #402

A partial solution is to keep the source code in `thirdparty/`.  A more complete solution is to switch to external projects, provided the local copies aren't patched for features.

---

Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.


ghutchis:
Want to do `nlohmann` json?

The tinycolormap is locally patched right now, though.

ghutchis (approved):
I definitely appreciate taking `#include <file.cpp>` to `#include <file.hpp>`

berquist:
> Want to try `nlohmann` json too? [json.nlohmann.me/integration/cmake/#supporting-both](https://json.nlohmann.me/integration/cmake/#supporting-both)
> 
> The tinycolormap is locally patched right now, though.

Yes, I have it half-done.

I'll avoid tinycolormap then.

Do you know where `struct` is from?

ghutchis:
As per `thirdparty/struct/README.md`:
https://github.com/svperbeast/struct

ghutchis:
Your current builds are failing because there are missing references to the FindLIBARCHIVE, e.g. CMakeLists.txt line 118:`     "${CMAKE_CURRENT_SOURCE_DIR}/cmake/FindLIBARCHIVE.cmake"`

ghutchis:
At the moment, all the builds fail with `importpqr/pqrrequest.cpp` looking for `#include <nlohmann/json.hpp>` (based on the `include_directories(SYSTEM "${AvogadroLibs_SOURCE_DIR}/thirdparty")` which you removed).

Similarly `plugindownloader/downloaderwidget.h` has `#include <nlohmann/json.hpp>`

If you roll back that last commit, I think it should build.

berquist:
This is a mixture of externally-provided libraries and promoting all TPLs to full targets, doing away with system-level includes.

I got rid of the `AddFoo.cmake` approach because there simply aren't enough dependencies to warrant it; there were too many small files.  If you decide that all unpatched vendored code can go away by using `ExternalProject`/`FetchContent`, it will look cleaner than it does now.

The additional `nlohmann` nesting is to avoid including `thirdparty` directly in a target.  That could allow one target to accidentally allow for includes to be covered by another target.

Ironically, the actual library mentioned in the issue is a plain Makefile project; I checked the Arch Linux package and it doesn't even provide a pkg-config file.  I don't want to write the package-finding code for it unless someone really wants it.

berquist:
My global JSON install in `/usr/include/nlohmann/...` must still be infecting my local build.

berquist:
I'm going to force push again and it will still fail clang-format, which I don't have set up locally quite yet.

ghutchis:
I can help with the `clang-format` - am I allowed to push to your branch as a maintainer?

berquist:
You should be able to, though I would rather fix/configure my setup.

ghutchis:
Looks good to me - @cryos any feedback or should we merge this?

cryos:
What prompted the bump to CMake 3.24? I don't see a commit message describing why, new feature you needed?

cryos:
The rest looks great to me, thanks for doing it! I just wonder if we really need to bump the CMake minimum.

cryos (changes requested):
I looked through the commit where the CMake minimum was bumped, everything looks safe with 3.23. I would prefer to stick with that unless there is a feature we need. Everything else looks great to me, ready to merge once we document why the bump in minimum or revert it.

berquist:
I could have sworn I needed 3.24 for the `GLOBAL` keyword on an imported library, but now I can't find it.

berquist:
Not `add_library`, `find_package`: https://cmake.org/cmake/help/latest/release/3.24.html#commands

cryos (approved):
Fair enough, it feels like a shame for a rarely needed thing. In future it would be good to remember to note when making dependency bumps like this why in the commit message for future us/reviewers.

ghutchis:
I don't know what happened here, but it seems like the compile cache isn't working quite right. I'm checking tonight and then if the Linux builds are still failing, I'll cut the cache and merge again.

PR #1288, opened 2023-05-03
A variety of clang-tidy fixes

ghutchis:
- Use .empty() instead of .size() comparisons
- Header cleanups
- Remove some unused "using" decl
- Use qualified auto (i.e., show pointers)
- Add const decl where possible


ghutchis:
Not sure why the graph tests are failing - the changes look innocuous and the tests work on my Mac. I'll check later.

PR #1311, opened 2023-06-14
Integration with 3Dconnexion input devices

pskowronskiTDx:
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

The contribution was created in whole or in part by me and I
indicated in the file.

I understand and agree that this project and the contribution


avogadro/rendering/camera.cpp
ghutchis, line 1: Please don't commit files with whitespace changes, unless these come from running `clang-format`

avogadro/rendering/camera.h
```diff
@@ -120,7 +120,21 @@ class AVOGADRORENDERING_EXPORT Camera
    * @param zFar is the distance from the viewer to the far clipping plane.
    */
   void calculatePerspective(float fieldOfView, float zNear, float zFar);
-
+#ifdef _3DCONNEXION
```
ghutchis, line 123: This seems like a perfectly reasonable API addition. In that case, please do not `#ifdef` the code and put the implementation properly in `camera.cpp`.

avogadro/rendering/geometryvisitor.cpp
```diff
@@ -9,6 +9,9 @@
 #include "curvegeometry.h"
 #include "linestripgeometry.h"
 #include "spheregeometry.h"
+#ifdef _3DCONNEXION
+#include "cylindergeometry.h"
```
ghutchis, line 13: Why is `cylindergeometry.h` needed?

avogadro/rendering/geometryvisitor.cpp
```diff
@@ -27,6 +30,9 @@ void GeometryVisitor::visit(Drawable&)
 
 void GeometryVisitor::visit(SphereGeometry& geometry)
```
ghutchis, line 31: What are you trying to do here?

avogadro/rendering/geometryvisitor.h
```diff
@@ -61,7 +71,34 @@ class GeometryVisitor : public Visitor
    * Get the radius of the scene.
    */
   float radius();
+#ifdef _3DCONNEXION
```
ghutchis, line 74: Again, this seems like a perfectly reasonable API addition without the `#ifdef` - and thus the implementation should be in the relevant `.cpp` file.

avogadro/rendering/geometryvisitor.h
```diff
@@ -41,7 +47,11 @@ class GeometryVisitor : public Visitor
   void visit(SphereGeometry&) override;
   void visit(AmbientOcclusionSphereGeometry&) override;
   void visit(CurveGeometry&) override;
+#ifdef _3DCONNEXION
+  void visit(CylinderGeometry&) override;
```
ghutchis, line 51: If you want to replace this `visit()` method, you should do that (and whatever else you want to do in the implementation rather than an `ifdef` in the header.

avogadro/rendering/geometryvisitor.h
```diff
@@ -25,6 +25,12 @@ namespace Rendering {
  * notably the center and radius of the bounding sphere.
  */
 
+#ifdef _3DCONNEXION
```
ghutchis, line 28: Why are these needed here?

avogadro/rendering/geometryvisitor.h
```diff
@@ -61,7 +71,34 @@ class GeometryVisitor : public Visitor
...
+   * @param maxY [out] maximum Y coordinate of the box diagonal
+   * @param maxZ [out] maximum Z coordinate of the box diagonal
+   * @param flags [in] flags informing which atoms will be included
+   * in the bounding box.
+   */
+  void boundingBox(double& minX, double& minY, double& minZ, double& maxX,
+                   double& maxY, double& maxZ,
+                   const std::vector<bool>& flags) const;
 
+  /**
```
ghutchis, line 91: Again. This seems like a reasonable API addition. Add it with no `#ifdef` and then the appropriate implementation.

avogadro/rendering/geometryvisitor.h
```diff
@@ -74,6 +111,11 @@ class GeometryVisitor : public Visitor
 
   std::vector<Vector3f> m_centers;
   std::vector<float> m_radii;
+#ifdef _3DCONNEXION
```
ghutchis, line 114: Why are you saving all of these? This seems like a really bad idea when someone has 10^5 or 10^6 atoms and bonds.

avogadro/rendering/glrenderer.cpp
```diff
@@ -27,8 +27,25 @@ using Core::Array;
...
   : m_valid(false), m_textRenderStrategy(nullptr), m_center(Vector3f::Zero()),
     m_radius(20.0)
+#ifdef _3DCONNEXION
+    ,
+    m_drawIcon(false), m_iconData(nullptr), m_iconWidth(0u), m_iconHeight(0u),
+    m_iconPosition(Eigen::Vector3f::Zero())
+#endif
 {
   m_overlayCamera.setIdentity();
+#ifdef _3DCONNEXION
```
ghutchis, line 37: I would be willing to see how this works in general. It doesn't seem specific to your code.

avogadro/rendering/glrenderer.cpp
```diff
@@ -169,16 +204,43 @@ void GLRenderer::setTextRenderStrategy(TextRenderStrategy* tren)
 void GLRenderer::applyProjection()
 {
   float distance = m_camera.distance(m_center);
+#ifdef _3DCONNEXION
```
ghutchis, line 207: I think I would rather test this as a general change rather than a bunch of `#ifdef` here.

avogadro/rendering/scene.h
```diff
@@ -130,11 +130,39 @@ class AVOGADRORENDERING_EXPORT Scene
 
   /** Clear the scene of all elements. */
   void clear();
-
+#ifdef _3DCONNEXION
```
ghutchis, line 133: Again, if you want to add API, just add API and associated implementation.

avogadro/rendering/tdxextensions.cpp
ghutchis, line 1: This file should not exist. Methods belong in their appropriate `.cpp` files so people can find them. Thanks.

ghutchis (changes requested):
In general, these seem reasonable. I'd much, much rather see new API methods and implementations in their associated files than a `tdxextensions.cpp` with some miscellaneous methods that belong elsewhere.

The main `ifdef` pieces I'd want to see are relative to drawing icons, etc.

ghutchis:
@pskowronskiTDx - any comments / concerns about the requested changes? Happy to discuss over e-mail if that's easier. Thanks!

avogadro/rendering/camera.cpp
pskowronskiTDx: Sure, I believe I have added them accidentaly.

avogadro/rendering/geometryvisitor.cpp
pskowronskiTDx: Before I moved part of the implementation to tdxextensions.cpp there were an overriden "visit" method for cylinder geometry here. I forgot to remove this include.

avogadro/rendering/geometryvisitor.h
pskowronskiTDx: They were added to cache scene geometry for hit testing and it's bounding box calculation.

avogadro/rendering/geometryvisitor.cpp
pskowronskiTDx: Cache the scene geometry.

avogadro/rendering/geometryvisitor.h
pskowronskiTDx: I agree, the hit testing part introduced by me is redundant and it can be memory consuming as you noted. I have reworked it today.

avogadro/rendering/glrenderer.cpp
pskowronskiTDx: Ok, I will remove mentioned #ifdef's from GLRenderer making my changes to this class general.

avogadro/rendering/geometryvisitor.h
pskowronskiTDx: It seems to be an another leftover. Removed.

pskowronskiTDx:
@ghutchis - Here is a summary of changes I did according to your review. I hope you wil be happy with them,

- The definitions of new methods have been putted back to relevant source files, tdxextensions.cpp has been removed.
- The #ifdef conditions got removed for Scene, Camera and GLRenderer (except for part responsible for drawing the 3Dx icon) classes.
- Saving the geometry got reduced to spheres. The point of doing so is to access the current collection of the spheres visible on scene. It is needed to calculate bounding box of the whole molecule or it's part consisting of selected atoms. I couldn't come up with more efficient solution to this problem so far. Maybe there is another approach you can suggest to follow?
- Sources were cleaned-up of whitespace changes and unnecessary includes.

ghutchis:
I'm traveling today, but will take a look. @cryos - do you have a moment to look as well?

pskowronskiTDx:
Hi Geoff, do you managed to take a look on the changes I have proposed to both PRs? Thanks.

avogadro/rendering/camera.cpp
```diff
@@ -23,7 +23,7 @@ Camera::Camera()
 Camera::Camera(const Camera& o)
   : m_width(o.m_width), m_height(o.m_height),
     m_projectionType(o.m_projectionType),
-    m_orthographicScale(o.m_orthographicScale), m_data(new EigenData(*o.m_data))
+    m_orthographicScale(o.m_orthographicScale), m_data(new EigenData(*o.m_data))
```
cryos, line 26: This just adds a trailing space and should be fixed.

avogadro/rendering/CMakeLists.txt
```diff
@@ -144,3 +145,7 @@ target_link_libraries(Rendering
     Avogadro::Core # Needed for an emum in residue.h that we should move, maybe others.
     GLEW::GLEW
     OpenGL::GL)
+
+if(USE_3DCONNEXION AND (WIN32 OR APPLE))
+  target_compile_definitions(Rendering PUBLIC _3DCONNEXION)
```
cryos, line 150: Does this need to be public?

avogadro/rendering/glrenderer.h
```diff
@@ -136,7 +152,7 @@ class AVOGADRORENDERING_EXPORT GLRenderer
   Scene m_scene;
   TextRenderStrategy* m_textRenderStrategy;
   SolidPipeline m_solidPipeline;
-
+
```
cryos, line 155: Just trailing space, please remove.

ghutchis:
> * Saving the geometry got reduced to spheres. The point of doing so is to access the current collection of the spheres visible on scene. It is needed to calculate bounding box of the whole molecule or it's part consisting of selected atoms. I couldn't come up with more efficient solution to this problem so far. Maybe there is another approach you can suggest to follow?

Can't you calculate the bounding box (or selected atom bounding) in `core/molecule.*` or `qtgui/molecule.*` instead? There are a few places that use similar code.

ghutchis:
Please just run `clang-format` on your modified files, thanks.

avogadro/rendering/camera.h
```diff
@@ -121,6 +121,20 @@ class AVOGADRORENDERING_EXPORT Camera
    */
   void calculatePerspective(float fieldOfView, float zNear, float zFar);
 
+  /**
+   * << API Extension for TDX >>
```
ghutchis, line 125: Please omit this. You may have added it, but there's plenty of reason to use this in other code.

avogadro/rendering/geometryvisitor.cpp
```diff
@@ -187,4 +189,55 @@ void GeometryVisitor::average()
   }
 }
 
+void GeometryVisitor::boundingBox(double& minX, double& minY, double& minZ,
```
ghutchis, line 192: As I mentioned, it seems easier to put this into `core/molecule.*`. If there's some reason you can't access the Molecule class, this is okay.

avogadro/rendering/geometryvisitor.h
```diff
@@ -62,6 +64,22 @@ class GeometryVisitor : public Visitor
    */
   float radius();
 
+  /**
+   * <<API Extension for TDX>>
```
ghutchis, line 68: Please, please omit "API Extension." I'm more than happy to provide credit / acknowledgement for your work and TDX, but we don't need to clutter the documentation with who wrote particular methods.

avogadro/rendering/scene.cpp
```diff
@@ -53,4 +53,14 @@ void Scene::clear()
   m_dirty = true;
 }
 
+void Scene::getBoundingBox(double& minX, double& minY, double& minZ,
+                           double& maxX, double& maxY, double& maxZ,
+                           const std::vector<bool>& flags)
+{
+  GeometryVisitor visitor;
```
ghutchis, line 60: As above - does this need to be in the `Scene` class?

avogadro/rendering/scene.h
```diff
@@ -131,10 +131,26 @@ class AVOGADRORENDERING_EXPORT Scene
   /** Clear the scene of all elements. */
   void clear();
 
+  /**
+   * <<API Extension for TDX>>
```
ghutchis, line 135: Please remove this line.

avogadro/rendering/scene.h
```diff
@@ -131,10 +131,26 @@ class AVOGADRORENDERING_EXPORT Scene
...
+   * @param minX [out] minimum X coordinate of the box diagonal
+   * @param minY [out] minimum Y coordinate of the box diagonal
+   * @param minZ [out] minimum Z coordinate of the box diagonal
+   * @param maxX [out] maximum X coordinate of the box diagonal
+   * @param maxY [out] maximum Y coordinate of the box diagonal
+   * @param maxZ [out] maximum Z coordinate of the box diagonal
+   * @param flags [in] flags informing which atoms will be included
+   * in the bounding box
+   */
+  void getBoundingBox(double& minX, double& minY, double& minZ, double& maxX,
```
ghutchis, line 146: Again, I think it would be easier to access the bounding box directly from the molecule (which could adjust as needed for isosurface meshes, etc.)

ghutchis (changes requested):
This looks much better, but please avoid using "API Extension" comments, which aren't needed and let's consider putting the bounding box into the `Molecule` class instead.

cryos:
I have been really busy the last month, sorry. I can take a look, you covered a lot of my initial concerns with comments and where the additional API is best placed @ghutchis

avogadro/core/molecule.cpp
```diff
@@ -1324,4 +1324,53 @@ std::list<Index> Molecule::getAtomsAtLayer(size_t layer)
...
 }
 
+void Molecule::boundingBox(Vector3& boxMin, Vector3& boxMax) const
+{
+  boxMin.setConstant(std::numeric_limits<double>::max());
+  boxMax.setConstant(-std::numeric_limits<double>::max());
+
+  bool noSelection = true;
+
+  for (uint32_t i = 0; i < m_selectedAtoms.size(); i++) {
```
ghutchis, line 1334: You can get this from `isSelectionEmpty()` but it's not a big deal either way.

avogadro/core/molecule.cpp
```diff
@@ -1324,4 +1324,53 @@ std::list<Index> Molecule::getAtomsAtLayer(size_t layer)
...
+    if (m_selectedAtoms[i]) {
+      noSelection = false;
+      break;
+    }
+  }
+
+  for (uint32_t i = 0; i < atomCount(); i++) {
+    if (noSelection || m_selectedAtoms[i]) {
+      double radius = 1.0;
```
ghutchis, line 1343: This seems small. I'd suggest adding it as a parameter with a default of 1.0?

avogadro/core/molecule.cpp
```diff
@@ -1324,4 +1324,53 @@ std::list<Index> Molecule::getAtomsAtLayer(size_t layer)
...
+
+      Vector3 boxMinBuffer;
+      Vector3 boxMaxBuffer;
+
+      boxMinBuffer.x() = atom(i).position3d().x() - radius;
```
ghutchis, line 1348: boxMinBuffer = atom(i).position3d().array() - radius;

https://stackoverflow.com/a/35693493/131896

avogadro/core/molecule.cpp
```diff
@@ -1324,4 +1324,53 @@ std::list<Index> Molecule::getAtomsAtLayer(size_t layer)
...
+      boxMinBuffer.y() = atom(i).position3d().y() - radius;
+      boxMinBuffer.z() = atom(i).position3d().z() - radius;
+      boxMaxBuffer.x() = atom(i).position3d().x() + radius;
```
ghutchis, line 1351: boxMaxBuffer = atom(i).position3d().array() + radius;

This way Eigen will properly perform an element-wide operation.

avogadro/core/molecule.cpp
```diff
@@ -1324,4 +1324,53 @@ std::list<Index> Molecule::getAtomsAtLayer(size_t layer)
...
+      boxMaxBuffer.y() = atom(i).position3d().y() + radius;
+      boxMaxBuffer.z() = atom(i).position3d().z() + radius;
+
+      if (boxMinBuffer.x() < boxMin.x())
+        boxMin.x() = boxMinBuffer.x();
```
ghutchis, line 1356: boxMin.x() = min(boxMin.x(), boxMinBuffer.x());

etc. .. which is more readable, since the library handles the conditional

avogadro/core/molecule.h
```diff
@@ -717,9 +718,18 @@ class AVOGADROCORE_EXPORT Molecule
   Layer& layer();
   const Layer& layer() const;
 
+  /**
+   * Calculte and return bounding box of the whole molecule or selected atoms
+   * only.
+   * @param boxMin [out] the minimum corner (first end of the box diagonal)
+   * @param boxMax [out] the maximum corner (second end of the box diagonal)
+   */
+  void boundingBox(Vector3& boxMin, Vector3& boxMax) const;
```
ghutchis, line 727: Please add  Real radius = 1.0 as an additional parameter

ghutchis (changes requested):
This all looks great .. just some minor nitpicks in the bounding box.

avogadro/core/molecule.h
pskowronskiTDx: Understood, suggested changes were introduced.

ghutchis:
Looks good to me - I'll re-run clang-format when I get a chance. Thanks for sticking with this, I'll turn on the flag for Windows and Mac builds.

Issue #1346, opened 2023-09-14
Continue Modernizing Code / Warnings


ghutchis:
There are some additional static-analysis cleanups that could be done to minimize warnings and improve the code. Some of these can be run easily from Qt Creator.

Please consider running one of the following, fix some things, and submit a pull request.

- [ ] Fix GCC / Clang warnings
- [ ] Cppcheck: https://github.com/danmar/cppcheck
- [ ] Clang static analyzers / clang-tidy: http://clang.llvm.org/extra/clang-tidy/
- [ ] Clazy (clang KDE/Qt checks): https://github.com/KDE/clazy
- [ ] Include-what-you-use: https://github.com/include-what-you-use/include-what-you-use

attharva15:
Hello,
I'm interested in working on this issue and would like to contribute to the project. Could you please assign the issue to me?
I know C/C++ Efficiently, and would love to work on the issue.
Please let me know if there are any specific guidelines or preferences for working on this issue, and I'll get started on it.

Thank you!

ghutchis:
I'm not going to assign the issue to anyone. If you'd like to work on some of the issues, please fix one thing at a time, and submit a PR for testing.

PR #1387, opened 2023-10-22
Modernized some code, willing to work on more. #1346

Kenshiro15:
#1346  **Description:**
#Used constexpr to define M_PI instead of a preprocessor macro.
#Removed #define preprocessor directive and used a more standard constexpr definition for M_PI.
#Updated loopvariables
I am also willing to change the variable types to use Eigen::Index, but i am a little unsure. I think it will ensure a better code.
I am new to opensource, and any newsuggestions would certainly be very helpful. 
Thank you!

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -1,0 +1,10 @@
+{
+    "files.associations": {
+        "array": "cpp",
+        "bitset": "cpp",
+        "initializer_list": "cpp",
+        "regex": "cpp",
+        "utility": "cpp",
+        "valarray": "cpp"
+    }
+}
--- a/avogadro/calc/chargemodel.cpp
+++ b/avogadro/calc/chargemodel.cpp
@@ -1,10 +1,9 @@
-/******************************************************************************
+ /******************************************************************************
   This source file is part of the Avogadro project.
   This source code is released under the 3-Clause BSD License, (see "LICENSE").
 ******************************************************************************/
 
 #include "chargemodel.h"
-
 #include <avogadro/core/array.h>
 #include <avogadro/core/molecule.h>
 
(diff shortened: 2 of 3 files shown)


ghutchis:
Thanks. There is already a PR for constexpr M_PI #1383 but IIRC there are a few more like DEG_TO_RAD (see `avogadro/core/avogadrocore.h`)

This commit only includes `.vscode/settings.json` which frankly needs to go into the .gitignore file: https://docs.github.com/en/get-started/getting-started-with-git/ignoring-files

ghutchis:
I'm also not sure I understand what you mean by updating loop variables.

Kenshiro15:
> Thanks. There is already a PR for constexpr M_PI #1383 but IIRC there are a few more like DEG_TO_RAD (see `avogadro/core/avogadrocore.h`)
> 
> This commit only includes `.vscode/settings.json` which frankly needs to go into the .gitignore file: https://docs.github.com/en/get-started/getting-started-with-git/ignoring-files

I guess i made some mistakes in using github, sorry as i am a little bit new to it. I will get back to this asap.

Kenshiro15:
> I'm also not sure I understand what you mean by updating loop variables.

We can use `size_t` instead of `unsigned int` (its not being used at someplaces), also we can use eigenMatrixx Eigen::Index,  in some of the loops.
Thank you for mentioning, i will get back to it.

Kenshiro15:
> Thanks. There is already a PR for constexpr M_PI #1383 but IIRC there are a few more like DEG_TO_RAD (see `avogadro/core/avogadrocore.h`)
> 
Hello,  after taking your valuable  feedback, 
I have changed the following avagadrocore.h code as required with many necessary changes. May i submit a new PR, if you may allow?

ghutchis:
Of course you can submit a new PR, thanks.

PR #1398, opened 2023-10-23
Allow for reading imaginary vibrational frequencies in ORCA outputs.

dtelsing:
When reading an ORCA output file with an imaginary mode (which have an ***imaginary mode*** comment in the frequency line), the line and subsequent lines are skipped. This leads to an out-of-bounds access when reading normal modes, the IR spectrum or the Raman spectrum, as the indices read from the file are not verified.

Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

### DIFF
--- a/avogadro/quantumio/orca.cpp
+++ b/avogadro/quantumio/orca.cpp
@@ -301,7 +301,11 @@ void ORCAOutput::processLine(std::istream& in, GaussianSet* basis)
           break;
         list = Core::split(key, ' ');
         while (!key.empty()) {
-          if (list.size() != 3) {
+          // imaginary frequencies can have an additional comment:
+          // ***imaginary mode***
+          if (list.size() != 3 &&
+              (list.size() != 5 || list[3] != "***imaginary" ||
+               list[4] != "mode***")) {
             break;
           }
           // e.g. 0:         0.00 cm**-1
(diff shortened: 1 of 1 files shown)


avogadro/quantumio/orca.cpp
```diff
@@ -397,6 +415,9 @@ void ORCAOutput::processLine(std::istream& in, GaussianSet* basis)
           // the first entry might be 5 or 6 because of removed rotations /
           // translations
           int index = Core::lexicalCast<int>(list[0]);
+          // invalid index
```
ghutchis, line 418: As above.

avogadro/quantumio/orca.cpp
```diff
@@ -332,13 +336,23 @@ void ORCAOutput::processLine(std::istream& in, GaussianSet* basis)
...
         list = Core::split(key, ' ');
         vector<int> modeIndex;
+        bool invalid_index = false;
         while (!key.empty()) {
           // first we get a set of column numbers
           // e.g. 1  2  3  4  5  6  7  8  9 10
           modeIndex.clear();
           for (unsigned int i = 0; i < list.size(); i++) {
-            modeIndex.push_back(Core::lexicalCast<int>(list[i]));
+            int index = Core::lexicalCast<int>(list[i]);
```
ghutchis, line 345: Shouldn't we just declare (index) as `std::size_t` here?

avogadro/quantumio/orca.cpp
```diff
@@ -374,6 +388,10 @@ void ORCAOutput::processLine(std::istream& in, GaussianSet* basis)
           // the first entry might be 5 or 6 because of removed rotations /
           // translations
           int index = Core::lexicalCast<int>(list[0]);
+          // invalid index
```
ghutchis, line 391: Shouldn't we just declare (index) as `std::size_t` above? In that case, `index > 0` below can be dropped too.

ghutchis:
Great, thanks! I obviously haven't dealt with many TS or imaginary modes and was largely porting the "Orca-enhanced Avogadro 1.2" code.

ghutchis:
The CMake build failure on Ubuntu / GCC seems like a problem on that build node. I'll restart in an hour or two. (There are lots of new PR, so each is taking a while.)

ghutchis:
If you can make the minor changes I mentioned, I'm happy to merge this ASAP. Thanks!

avogadro/quantumio/orca.cpp
dtelsing: Yes, I just kept it as it was before.

avogadro/quantumio/orca.cpp
dtelsing: Yes.

PR #1415, opened 2023-10-27
Dev extensions

research11111:
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.


avogadro/qtgui/CMakeLists.txt
```diff
@@ -1,3 +1,15 @@
+find_package(Eigen3 REQUIRED)
+# Add as "system headers" to avoid warnings generated by them with
+# compilers that support that notion.
+include_directories(SYSTEM ${EIGEN3_INCLUDE_DIR})
+
+cmake_minimum_required(VERSION 3.5)
```
cryos, line 6: None of this makes sense and is not the direction we want to move the CMake code in this project.

research11111:
This pull request addresses Issue #1391 and includes the following enhancements:

- Sets the Python path similarly to how the `quantumInput` extension does.
- Partial support for `virtualenv`; manual path input is required from the user.
- Enables the addition of custom plugin URLs (GitHub-hosted only).
- Allows local download of the `plugins.json` file and the ability to add entries to it.
- Supports installations both with and without Conda.
- Features auto-detection of existing installations.
- Automates the installation of dependencies, either from `requirements.txt` or `pyproject.toml` (https://github.com/research11111/avogadro-rdkit/ may be used to validate).

avogadro/qtgui/CMakeLists.txt
research11111: mistake it remove

research11111:
add CMakeFile

PR #1500, opened 2023-12-04
Fixed Menu Tab options Bugs

secretkontributer:
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

### DIFF
--- a/avogadro/qtplugins/select/select.cpp
+++ b/avogadro/qtplugins/select/select.cpp
@@ -550,7 +550,7 @@ void Select::createLayerFromSelection()
   if (!m_molecule)
     return;
 
-  if(!m_molecules->atomeCount())
+  if(!m_molecules->atomCount())
     return;
 
   QtGui::RWMolecule* rwmol = m_molecule->undoMolecule();
--- a/avogadro/qtplugins/select/select.cpp
+++ b/avogadro/qtplugins/select/select.cpp
@@ -139,7 +139,7 @@ bool Select::evalSelect(bool input, Index index) const
 
 void Select::selectAll()
 {
-  if (m_molecule) {
+  if (m_molecule && m_molecule->atomCount() > 0) {
     for (Index i = 0; i < m_molecule->atomCount(); ++i) {
       m_molecule->undoMolecule()->setAtomSelected(i, evalSelect(true, i));
     }
@@ -150,7 +150,7 @@ void Select::selectAll()
 
 void Select::selectNone()
 {
-  if (m_molecule) {
+  if (m_molecule && m_molecule->atomCount() > 0) {
     for (Index i = 0; i < m_molecule->atomCount(); ++i)
       m_molecule->undoMolecule()->setAtomSelected(i, false);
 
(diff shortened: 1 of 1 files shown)


secretkontributer:
My PR contains fixed code of both the issues i raised on discussion forum.

avogadro/qtplugins/select/select.cpp
```diff
@@ -438,11 +438,13 @@ void Select::selectAtomIndex()
     // check if it's a range
     if (item.contains('-')) {
       auto range = item.split('-');
-      if (range.size() >= 2) {
```
ghutchis, line 441: But what about the case with "1-2,5-6,8,10-12" ?

avogadro/qtplugins/select/select.cpp
```diff
@@ -544,6 +550,9 @@ void Select::createLayerFromSelection()
   if (!m_molecule)
     return;
 
+  if(!m_elements)
```
ghutchis, line 553: This is definitely not the right fix. That's a dialog if someone wants to select elements. If someone opens the "Select by Elements" the dialog will be created.

I'd check to see if the molecule has atoms?

ghutchis:
See if https://github.com/OpenChemistry/avogadrolibs/pull/1499 solves your crashes.

avogadro/qtplugins/select/select.cpp
secretkontributer: it will work sir i changed the range.size()==2 
because of this case 8-9-7 which is valid in that if loop but doesnt make sense

avogadro/qtplugins/select/select.cpp
secretkontributer: I can use m_molecule->atomCount() to check this

avogadro/qtplugins/select/select.cpp
secretkontributer: it will crash the system if we dont remove the case of 5-6-7

avogadro/qtplugins/select/select.cpp
secretkontributer: but i think we need to select atoms first to create a newLayer

secretkontributer: thats the reason i went with this solution

secretkontributer: I have made relevant changes  you can check it

secretkontributer: I have updated according to the latest PR of issue 1499

avogadro/qtplugins/select/select.cpp
secretkontributer: @ghutchis  sir can u look in this pr?

ghutchis:
I'm closing this because the tests didn't run. It's now in #1625 - the code looks good so let's make sure everything passes.

Thanks and sorry for the delay

PR #1503, opened 2023-12-05
If userOptions specifies an order, use that to sort the form

### LINKED ISSUE #1498: Allow plugins/extensions to request specific positions in menus, choose order of commands, layout of dialogs
Currently anything added to a menu or dialog by a plugin seems to be displayed in alphabetical order. Ideally the plugin could choose how these things are displayed.

If I understand correctly, in the code, menu options are given an integer as a sort of "priority" and are ordered by that, and dividers are added every 100. So ideally what would be implemented:

1. When a script specifies its location using `--menu-path`, it would be able to provide a number to use for sorting *at each level* so that it could affect both the plugin's location in the menus, but also the relative order of the plugin's commands if it has multiple.

Secondly, the options for customizing script-specific dialogs are fantastic and allow for a lot of flexibility, but the way things are displayed is not controllable:

2. The relative order of dialog widgets specified in  `"userOptions"` should be settable.

I am not sure what the mechanism is for the latter in other dialogs in the program.

ghutchis:
Fix #1498

Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

### DIFF
--- a/avogadro/qtgui/jsonwidget.cpp
+++ b/avogadro/qtgui/jsonwidget.cpp
@@ -208,10 +208,26 @@ void JsonWidget::buildOptionGui()
     }
 
     // Add remaining keys at bottom.
+    // look for "order" key to determine order
+    QStringList keys;
     for (QJsonObject::const_iterator it = userOptions.constBegin(),
                                      itEnd = userOptions.constEnd();
          it != itEnd; ++it) {
-      addOptionRow(it.key(), it.key(), it.value());
+      if (it.value().isObject()) {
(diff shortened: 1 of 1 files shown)


matterhorn103:
This has a slightly odd bug whereby if the first item in a dialog (set by `"order"`) is a field of type `"string"`, the value set by `"default"` is wiped and the text box is displayed as empty. Drop down `"stringList"`s also get reset to the value with index 0 if they are the top object. This only happens to the first object.

PR #1504, opened 2023-12-05
Fix include of avogadrocoreexport.h

### LINKED ISSUE #1476: core header file includes header without path
The core header file `coordinateblockgenerator.h` includes avogadrocoreexport.h at
https://github.com/OpenChemistry/avogadrolibs/blob/204a4098f5f13202c3795367b5f4c2a2a5f7e7ff/avogadro/core/coordinateblockgenerator.h#L9

avogadrocoreexport.h is a generated file that will be located in the same directory as coordinateblockgenerator.h, i.e. `/usr/include/avogadro/core/avogadrocoreexport.h`

At l.9 it is included using angle brackets, `#include <avogadrocoreexport.h>`. Angle brackets indicate that its path is determined relative to the system include dirs, which are `/usr/include/`.  In this case the reference should give the relative path
```
#include <avogadro/core/avogadrocoreexport.h>
```
same as done in [coordinateblockgenerator.cpp](https://github.com/OpenChemistry/avogadrolibs/blob/204a4098f5f13202c3795367b5f4c2a2a5f7e7ff/avogadro/core/coordinateblockgenerator.cpp#L9) to include other core headers.

Alternatively if avogadrocoreexport.h will always be in the same directory as coordinateblockgenerator.h then a local path could be indicated with quote marks instead of angle brackets,
```
#include "avogadrocoreexport.h"
```

The current configuration gives an error using g++ 13.2.0. It possibly did not give an error in earlier gcc versions.

It may be partially responsible for failure to build tests against the installed library reported in https://github.com/OpenChemistry/avogadrolibs/issues/1430

ghutchis:
Fix #1476

Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

### DIFF
--- a/avogadro/core/atomutilities.h
+++ b/avogadro/core/atomutilities.h
@@ -8,8 +8,6 @@
 
 #include "avogadrocoreexport.h"
 
-#include "avogadrocoreexport.h"
-
 #include <avogadro/core/molecule.h>
 
 #include <vector>
--- a/avogadro/core/coordinateblockgenerator.h
+++ b/avogadro/core/coordinateblockgenerator.h
@@ -6,7 +6,7 @@
 #ifndef AVOGADRO_CORE_COORDINATEBLOCKGENERATOR_H
 #define AVOGADRO_CORE_COORDINATEBLOCKGENERATOR_H
 
-#include <avogadrocoreexport.h>
+#include "avogadrocoreexport.h"
 
 #include <sstream>
 #include <string>


drew-parsons:
I'd suggest using this PR as an audit of all the files using avogadrocoreexport.h, and also edit the others so they're consistent, also using
```
#include "avogadrocoreexport.h"
```

ghutchis:
@drew-parsons .. I did that:

`git grep "avogadrocoreexport"`
…

Everything now uses quotes. I did find one which included it twice (evidently from a script).

Issue #567, opened 2021-05-09
Add support for Wayland


kevinsmia1939:
**Is your feature request related to a problem? Please describe.**
It will be some time before Wayland gain adoption. But since Ubuntu 21.04 and Fedora 34 switch to Wayland as default.
Avogadro in Wayland is broken at the moment. If I were to guess, it might be because broken glew on Wayland. So when glew support it, Avogadro might be.

**Describe the solution you'd like**
Run Avogadro on Wayland.

**Describe alternatives you've considered**

**Additional context**
https://discuss.avogadro.cc/t/will-avogadro-support-wayland/3499

ghutchis:
As  I mentioned on the discussion forum, while there's a potential fix for GLEW,  there are also several alternatives:
- https://github.com/anholt/libepoxy
- https://github.com/skaslev/gl3w

I'd probably lean towards gl3w when we switch to the core profile.

matterhorn103:
Noticed this issue while browsing the repo. I am running 1.98.1 on Wayland (KDE Plasma 5.27.9 on OpenSUSE Tumbleweed) just fine. Presumably this means the problem has since been resolved?

kevinsmia1939:
> Noticed this issue while browsing the repo. I am running 1.98.1 on Wayland (KDE Plasma 5.27.9 on OpenSUSE Tumbleweed) just fine. Presumably this means the problem has since been resolved?

Where did you download from?

matterhorn103:
Ok, this seems to only be true for the AppImage version, as when I build from source, I get graphical glitches.

@kevinsmia1939 The AppImage from https://two.avogadro.cc/ and the [nightly build](https://nightly.link/OpenChemistry/avogadrolibs/workflows/build_cmake/master/Avogadro2.AppImage.zip) available via the readme of the repo both work fine on my PC on the Wayland session. Are they are broken for you?

PR #1577, opened 2024-01-15
Fix for Wayland

matterhorn103:
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.


matterhorn103:
See [this issue](https://github.com/nigels-com/glew/issues/172) at the GLEW repo.

Fixes #567.

As discussed [on the forum](https://discuss.avogadro.cc/t/will-avogadro-support-wayland/3499/11), the current proposal is that this be a temporary fix for 1.99 and that for 2.0 Avogadro moves away to another bindings library.

Note that my commit message is wrong, GLEW needs to be from *2022* or newer in order to have `glewContextInit` defined. Presumably using the current state of the GitHub master branch would do the trick; I used the "recent snapshot" (Apr 2022) found in the repo's readme which is [here](https://sourceforge.net/projects/glew/files/glew/snapshots/glew-20220402.tgz/download) and works.

The URL in `cmake/projects.cmake` needs to be changed appropriately in order for it to compile correctly, and obviously use of system GLEW needs to be disabled.

It seems GLEW is a while away from any 2.2.1 release still, so I guess that means not only compiling the AppImage and flatpak with the snapshot, but also packaging our own GLEW for 1.99, unless distros have backported the fixes since the 2.2.0 release.

ghutchis:
I mentioned on the forum, but is there something in the snapshot header that indicates that `glewContextInit()` is available?

ghutchis:
Okay, here's what I think we need to do - create a `#define GLEW_CONTEXT_INIT` using cmake. That is, we try to compile a small test program with the new `glewContextInit()` call. If it works, then cmake will create the `#define`.

For example:
https://stackoverflow.com/a/25732622/131896

matterhorn103:
> I mentioned on the forum, but is there something in the snapshot header that indicates that `glewContextInit()` is available?

Well it builds correctly without complaining, whereas when I use the 2.2.0 release it tells me that `glewContextInit()` is undefined within this scope at compile time.

matterhorn103:
> Okay, here's what I think we need to do - create a `#define GLEW_CONTEXT_INIT` using cmake. That is, we try to compile a small test program with the new `glewContextInit()` call. If it works, then cmake will create the `#define`.
> 
> For example: https://stackoverflow.com/a/25732622/131896

I'm not sure I quite follow. Is your suggestion that `glewInit()` continues to be used normally and `glewContextInit()` is only used if its availability is detected? And only use it on Linux?

Is it a bad idea to just use the snapshot and `glewContextInit()` for building on all platforms then?

ghutchis:
Consider a package maintainer. GLEW is at 2.2.0 release, which does not have this function. They want to compile `avogadro2` packages against released packages. If we switch to `glewContextInit()` completely, they don't have many options.

Unfortunately, GLEW doesn't provide its own `#define` to indicate the method is present. Worse, both the release and the snapshot declare they are version 2.2.0.

If we have a `cmake` test to see if `glewContextInit()` compiles, and create our own `#define` we can do something like:

```c++
#ifdef GLEW_HAS_CONTEXT_INIT
   glewContextInit(…)
#else
   glewInit(…)
#endif
```

ghutchis:
What GLEW did is really annoying:
- new method rather than fixing `glewInit()`
- no release yet, only a snapshot
- no `#define` in the snapshot to indicate the new method is present

So for now, we'll need to do a workaround until we can remove GLEW for good.

matterhorn103:
> What GLEW did is really annoying:
> 

Yes absolutely, I get that they don't want to vouch for a new release without testing of it first but still, it's been 3 years of commits with no new release. This despite the fact that Wayland support is nowadays pretty crucial. And as you say, reporting itself as version 2.2.0 seems a little cheeky as well.

I don't know if there's any point asking them to release the snapshot as 2.2.1? Since that's been available for nearly two years, I would think the maintainer should know roughly how many/if any bugs were introduced by it.

> Consider a package maintainer. GLEW is at 2.2.0 release, which does not have this function. They want to compile avogadro2 packages against released packages. If we switch to glewContextInit() completely, they don't have many options.

I mean fair enough, but at the same time, if a maintainer compiles against the released package then Avogadro won't work for 50% of the distro's users without a command line flag, so that feels like a non-starter no?

If at least the Flatpak and/or AppImage are compiled in such a way that they work on Wayland, people will always have those options, I guess. And both of those will become de-GLEWed 2.0 soon enough.

Issue #347, opened 2018-08-03
Add depth-of-field fog or blurring rendering options


ghutchis:
Avogadro 1 included "fog" to add some simple depth-of-field effects, e.g.:
![8-sqap](https://user-images.githubusercontent.com/41128/43658215-3b3eefe4-9726-11e8-90f1-d5a63611af07.png)

VTK offers other kinds of depth-of-field effects including blurring:
![](https://blog.kitware.com/wp-content/uploads/2016/02/image001-1024x753.png)

ghutchis:
From Avo1:

```
      glFogi(GL_FOG_MODE, GL_LINEAR);
      GLfloat fogColor[4]= {d->background.redF(), d->background.greenF(),
                            d->background.blueF(), d->background.alphaF()};
      glFogfv(GL_FOG_COLOR, fogColor);
      Vector3d distance = camera()->modelview() * d->center;
      double distanceToCenter = distance.norm();
      glFogf(GL_FOG_DENSITY, 1.0);
      glHint(GL_FOG_HINT, GL_NICEST);
      glFogf(GL_FOG_START, distanceToCenter - (d->fogLevel / 8.0) * d->radius);
      glFogf(GL_FOG_END, distanceToCenter + ((10-d->fogLevel)/8.0 * 2.0) * d->radius);
      glEnable(GL_FOG);
```

ghutchis:
This doesn't work in OpenGL core AFAICT, but there are shader options, e.g.:

https://vicrucann.github.io/tutorials/osg-shader-fog/

aerkiaga:
Related to this, I'd suggest (once OpenGL 3 is in place) trying to have a set of common fragment shaders (fog, ambient occlusion...) rather than one per render type, and moving color computation, if applicable, into the geometry shader. Then all render types would be able to use any shader. Also simpler, more orthogonal code.

ghutchis:
Absolutely. See #582 which I don't think even requires switching to OpenGL 3. (It also includes an "outline" shader which would be nice for surfaces.)

IIRC 3Dmol.js uses a similar strategy and it enables a volumetric render for orbitals / electron density:
http://3dmol.csb.pitt.edu

I can imagine a "View Options" window that allows tuning the global shader options.

ghutchis:
I'm going to leave this one open, since a blurring shader would still be nice.

perminder-17:
[...]

PR #1662, opened 2024-04-11
Add depth-of-field (Blurring rendering options)

perminder-17:
issue : #347 
Signed-off by : permindersingh089@gmail.com 
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.


perminder-17:
Hi @ghutchis code is functioning correctly, but there are a couple of concerns I'd like your help with. Here's the other PR https://github.com/OpenChemistry/avogadroapp/pull/492

1. The code currently uses 64 samples for blurring, following the number of samples from Speck. However, this = reduces frame rates a bit. I'm considering reducing the number of samples, but I'm concerned about compromising the quality of the blur. Do you think decreasing the samples would be advisable?
2. I've noticed that when we zoom in or out, the blur changes, which differs from the behavior in Speck. While this deviation could mimic real-world effects, such as objects appearing blurry when they're too close or too far. Considering potential future features like VR, this behavior might actually enhance realism. What's your take on this? Should we maintain or adjust the blur behavior?
3. Your feedback is invaluable, and I'm committed to optimizing and tidying up the code. I'd greatly appreciate your input on any minor cleanups or optimizations you think are necessary.
4. I was thinking about adding blur and fog effects to a new file, but I believe combining all rendering options into one would be better.

Screenshot from 2024-04-10 06-10-04

perminder-17:
Have a look, How does it looks now: Does not blur when camera is too close

### BEFORE
Screenshot from 2024-04-11 21-50-41

### AFTER-BLUR (THE INTENSITY CAN BE INCREASED)
Screenshot from 2024-04-11 21-50-28

### BEFORE
Screenshot from 2024-04-11 21-47-44

### AFTER-BLUR
Screenshot from 2024-04-11 21-47-28

avogadro/rendering/solid_first_fs.glsl
```diff
@@ -66,6 +68,47 @@ float lerp(float a, float b, float f)
     return a + f * (b - a);
 }
 
+float rand(vec2 co){
+  return fract(sin(dot(co.xy ,vec2(12.9898,78.233))) * 43758.5453);
+}
```
vinayakjeet, line 73: float rand(vec2 co) {
  return fract(sin(dot(co.xy ,vec2(12.9898,78.233)) + cos(dot(co.xy, vec2(93.9898,50.233)))) * 43758.5453);
}
//more complex pattern produced which helps in reduing the noise. @ghutchis your insights ?

avogadro/rendering/solid_first_fs.glsl
```diff
@@ -66,6 +68,47 @@ float lerp(float a, float b, float f)
...
+  // number of samples can be optimized.
+  for (int i = 0; i < 64; i++) {
+    float t = (float(i) / float(64));
+    float angle = (t * 4.0) * 6.28319; 
+    float radius = (t * 2. - 1.); 
+    angle += 1.0 * rand(gl_FragCoord.xy);
+    vec2 offset = (vec2(cos(angle), sin(angle)) * radius * 0.05 * blurAmt) / pixelScale;
+    float z = depthToZ(texture2D(inDepthTex, texCoord + offset).x);
+    float sampleBlur = calcBlur(z, pixelScale);
+    float weight = float((z >= origZ) || (sampleBlur >= blurAmt * radius + 0.));
```
vinayakjeet, line 103: float weight = 1.0 - smoothstep(0.0, 1.0, abs(z - origZ) / blurAmt); 
// using smoothstep to soften weight transitions
This modification uses smoothstep to create a more gradual transition in weights applied to sampled colors, potentially reducing artifacts like banding. Reviews @ghutchis

vinayakjeet (commented):
**Review of Shader Code Enhancements**

The proposed changes to the shader code, including enhanced randomness in the `rand` function, dynamic sample adjustment, and smoother weight transitions using `smoothstep`, significantly improve performance and visual quality. These updates are well-aligned with best practices for rendering optimizations.

@ghutchis , could you please review these enhancements and provide your feedback?

avogadro/rendering/solid_first_fs.glsl
```diff
@@ -66,6 +68,47 @@ float lerp(float a, float b, float f)
...
+}
+
+vec4 applyBlur(vec2 texCoord) {
+  float pixelScale = max(width, height);
+  float origZ = depthToZ(texture2D(inDepthTex, texCoord).x);
+  float blurAmt = calcBlur(origZ, pixelScale);
+  
+  float total = 1.0;
+  vec4 color = texture2D(inRGBTex, texCoord);
+  // number of samples can be optimized.
```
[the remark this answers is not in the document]
perminder-17: Thank you for the review. I'll definitely make time to address the suggestions. However, regarding this issue, the loop index can't be compared with a non-constant expression. Therefore, adding `numSamples` to the for loop won't function since its value isn't constant. Could also leads to error. Are you sure about this one??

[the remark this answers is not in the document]
vinayakjeet: You're right, and I appreciate your careful consideration. While GLSL 1.2 does support dynamic looping, the behavior can indeed vary based on the GPU and driver support. Sticking with your current approach to ensure compatibility and stability is wise. Thank you for pointing this out, @guthichs, and your reviews and help are greatly appreciated!

perminder-17:
https://github.com/OpenChemistry/avogadrolibs/assets/127239756/773bd5b0-f7f1-4876-81c9-f34412da3a89

https://github.com/OpenChemistry/avogadrolibs/assets/127239756/a5157f19-c22a-4e35-8634-0a3b27713cff


https://github.com/OpenChemistry/avogadrolibs/assets/127239756/f03f5cba-5e28-47fc-a284-fe31316dfaf7


Hi, I just have pushed the recent changes, above is the result I uploaded to show how depth-of-field is working for now. If any contributors or users of avogadro software have any suggestion, I would be more than happy to consider those changes into my code. :)

avogadro/rendering/solid_first_fs.glsl
```diff
@@ -32,14 +32,20 @@ uniform float offset;
...
 // 0.0 if disabled
 uniform float inFogStrength;
 // Shadow strength for SSAO
 uniform float inAoStrength;
 // 1.0 if enabled, 0.0 if disabled
 uniform float inEdStrength;
-// offset for zoom-in and zoom-out
+// amount of offset when zoom-in or zoom-out.
 uniform float uoffset;
+// Dof strength
```
ghutchis, line 45: Much like the fog strength, I'd much rather move to one float for inDofStrength == 0.0 if disabled. (I'll be submitting a patch for the ambient occlusion in a moment.)

avogadro/rendering/solid_first_fs.glsl
ghutchis: I think the smoothstep() would be nice here, since it's also easier to read.

avogadro/rendering/solid_first_fs.glsl
```diff
@@ -136,18 +184,25 @@ float computeEdgeLuminosity(vec3 normal)
...
   float luminosity = 1.0;
   luminosity *= max(1.2 * (1.0 - inAoEnabled), computeSSAOLuminosity(getNormalNear(UV)));
   luminosity *= max(1.0 - inEdStrength, computeEdgeLuminosity(getNormalAt(UV)));
   vec4 color = texture2D(inRGBTex, UV);
-  if (inFogStrength == 0.0) {
+  if (inFogStrength == 0.0 && inDofEnabled == 0.0) {
     gl_FragColor = vec4(color.xyz * luminosity, color.w);
   }
-  else {
+  else if (inFogStrength != 0.0) { // apply fog since it's fast.
```
ghutchis, line 195: I think we may want a separate shader path with fog, but not AO or depth blur (e.g., for speed, less-powerful GPU, etc.). I'll follow up on the forum.

avogadro/rendering/solidpipeline.cpp
```diff
@@ -159,6 +161,9 @@ void SolidPipeline::end()
   d->attachStage(d->firstStageShaders, "inRGBTex", d->renderTexture, "inDepthTex",
                  d->depthTexture, m_width, m_height);
   d->firstStageShaders.setUniformValue("inAoEnabled", m_aoEnabled ? 1.0f : 0.0f);
+  d->firstStageShaders.setUniformValue("inDofEnabled", m_dofEnabled ? 1.0f : 0.0f);
+  d->firstStageShaders.setUniformValue("inDofStrength", ((m_dofStrength) * 100.0f));
```
ghutchis, line 165: As I mentioned, let's set inDofStrength to 0.0 if m_dofEnabled is false.

ghutchis (commented):
Just some minor cleanups, but it otherwise looks good.

perminder-17:
I separated the fog shader into a different file to make the process faster because the Ambient Occlusion (AO) operation was running a bit slow. Thanks to @ghutchis  for the helpful insights! 

https://github.com/user-attachments/assets/dbcdefe7-c416-4a55-9275-a7eb7f32b2e4

ghutchis:
Sorry, I guess I wasn't clear. I think the fog should be in both shaders:
- AO + DOF + fog (i.e., good GPU shader)
- just fog (i.e., faster rendering)

In other words, if AO or DOF or Edge detection is enabled, the "full" shader is used. If all three are off, then the faster fog-only shader is used for speed.

Does that make sense?

So my suggestion is to put back the fog into the main shader. That's the default. Then we can change which shader is used in the C++, e.g.
```
if (ao_enabled || dof_enabled || edge_enabled)
   // use full shader
else
   // fast fog-only shader
```

perminder-17:
> Sorry, I guess I wasn't clear. I think the fog should be in both shaders:

Ahh..I am sorry. I thought we need to separate the fog shaders, Sorry for the misunderstanding. I will fix that now.

perminder-17:
Okay, I think I have worked on the suggestion @ghutchis . Let me know if you have any other suggestions?

ghutchis:
I think the effect may need some tweaking (e.g., strength, position) but the code looks good. Thanks for all the work!

perminder-17:
> I think the effect may need some tweaking (e.g., strength, position) but the code looks good. Thanks for all the work!

Yes, it could be improved. The problem was when you zoom-out zoom-in the change in the projection matrix is not consistent or linear. That's what makes it a problem. :"). I could probably open a new PR in improving it shortly.

ghutchis:
It's a great effect. Exactly as you say, it's the change when zooming. Tweaking the strength and projection matrix are secondary now that the code is already merged.

I needed to confirm that https://github.com/OpenChemistry/avogadroapp/pull/492 built cleanly after the merge. I'll also tweak the text a bit in the labels for avogadroapp.

PR #1659, opened 2024-03-30
Enhance PDB File Import to Support Trajectory Data in Avogadro 

vinayakjeet:
This PR updates pdbformat.cpp to improve the handling of PDB trajectories, aligning with community guidelines for file processing without integrating GUI elements. Key enhancements include:

Frame Separation with ENDMDL: Implements detection of ENDMDL to identify and separate frames within PDB trajectories, enabling accurate animation and analysis of sequential models.

Frame Position Storage: Accumulates atomic positions per frame until ENDMDL, then stores these as distinct frames in the molecule, facilitating correct trajectory visualization.

Final Frame Processing: Ensures the last set of positions is added as a frame, improving robustness in handling diverse PDB files.

PR #1668, opened 2024-04-19
Added a Resetview Option Feature

### LINKED ISSUE #1666: Add a reset view option
Something that would be useful to have is a "Reset View" option. This should return the view to the default initial state: axes in their default alignment (+x to the right, +y upwards, +z out of the screen), the origin of the coordinate system in the centre of the view, and the zoom level at the default.

This seems to be exactly what happens automatically when switching between molecules in the `Molecules` pane – each is initially displayed in the default view state. If you change the view of "molecule 1" with the Manipulate Tool, then switch to another molecule and back to molecule 1, the view is (for better or worse) not the way you left it but rather completely reset. It would be good to be able to get the same result via a menu option.

This would also be necessary to implement if it was changed so that switching between molecules doesn't reset the view automatically (which would also be a good change to make, as the lack of persistent state makes comparing similar files/molecules difficult if the aspect of interest isn't easily visible in the default view).

In the `View` menu we already have options to "Centre" and "Align View to Axes", which are useful too in specific situations, but seem to be subtly different and don't achieve a reset, even in combination:

- `Centre` puts the calculated centre of the atom ensemble in the centre of the view, not the coordinates origin, and while it changes the zoom level to an appropriate one, it is not the origin that is put "in the plane" but rather the centre. It's also not clear to me how the alignment of the axes is decided, but it's not the default.

- `Align View to Axes` is very close to a reset, but doesn't have quite the same effect as switching to another molecule and back. The axes are reset to the default orientation, but again it seems the view is centred on the centre of the atom ensemble, not on the origin.

vinayakjeet:
#1666 
## Overview
This pull request introduces a "Reset View" option to Avogadro, enabling users to easily revert the view to its default state. This includes resetting the camera angles, the origin of the coordinate system, and the zoom level.

## Changes
- **GLWidget**: Added `resetViewSettings()` to reset camera and view-related settings.
- **MultiViewWidget**: Implemented `resetAllViews()` to propagate the reset across all embedded view widgets.

## Purpose
The addition of this feature enhances usability by providing a straightforward way to reset views, especially useful when switching between molecules or configurations. This mirrors the behavior seen when switching between molecules in the Molecules pane, where the view resets to a default state.

closes #1666

### DIFF
--- a/avogadro/io/pdbformat.cpp
+++ b/avogadro/io/pdbformat.cpp
@@ -58,14 +58,12 @@ bool PdbFormat::read(std::istream& in, Core::Molecule& mol)
 
   while (getline(in, buffer)) { // Read Each line one by one
 
-    if (startsWith(buffer, "ENDMDL")) {
-      if (coordSet == 0) {
-        mol.setCoordinate3d(mol.atomPositions3d(), coordSet++);
-        positions.reserve(mol.atomCount());
-      } else {
-        mol.setCoordinate3d(positions, coordSet++);
-        positions.clear();
-      }
+   if (startsWith(buffer, "ENDMDL")) {
+  // Ensure positions array is not empty before creating a new frame
+    if (!positions.empty()) {
+    mol.setCoordinate3d(positions, coordSet++);
+    positions.clear(); // Clear positions after adding them as a frame
+    }
     }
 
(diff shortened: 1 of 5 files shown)


vinayakjeet:
@ghutchis @matterhorn103 
Please note, that this PR includes changes from PR #1659. Please review and merge PR #1659  first to simplify the process.

PR #1683, opened 2024-07-02
Updating Shader to GL ES 3.0

perminder-17:
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

### DIFF
--- a/avogadro/rendering/textlabelbase.cpp
+++ b/avogadro/rendering/textlabelbase.cpp
@@ -183,7 +183,7 @@ void TextLabelBase::RenderImpl::render(const Camera& cam)
       !shaderProgram->setUniformValue("vpDims", vpDims) ||
       !shaderProgram->setUniformValue("anchor", anchor) ||
       !shaderProgram->setUniformValue("radius", radius) ||
-      !shaderProgram->setTextureSampler("texture", texture) ||
+      !shaderProgram->setTextureSampler("u_texture", texture) ||
 
       !shaderProgram->enableAttributeArray("offset") ||
       !shaderProgram->useAttributeArray("offset", PackedVertex::offsetOffset(),
--- a/avogadro/rendering/textlabelbase_fs.glsl
+++ b/avogadro/rendering/textlabelbase_fs.glsl
@@ -1,9 +1,13 @@
-uniform sampler2D texture;
-varying vec2 texc;
+#version 300 es
+precision highp float;
+uniform sampler2D u_texture;
+in vec2 texc;
+
+out vec4 outColor;
 
 void main(void)
 {
-  gl_FragColor = texture2D(texture, texc);
-  if (gl_FragColor.a == 0.)
+  outColor = texture(u_texture, texc);
+  if (outColor.a == 0.)
     discard;
 }
--- a/avogadro/rendering/textlabelbase_vs.glsl
+++ b/avogadro/rendering/textlabelbase_vs.glsl
@@ -1,3 +1,4 @@
+#version 300 es
 // Modelview/projection matrix
 uniform mat4 mv;
 uniform mat4 proj;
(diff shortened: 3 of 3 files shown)


perminder-17:
This is still a work in progress PR

ghutchis:
Please check the following patch. I get a bunch of errors from the shaders on my Mac, because texture2D is renamed to texture:
[texture.patch](https://github.com/user-attachments/files/16103347/texture.patch)

Nothing renders yet, but it also doesn't crash, which is a good start. 😄

perminder-17:
> Please check the following patch. I get a bunch of errors from the shaders on my Mac, because texture2D is renamed to texture: [texture.patch](https://github.com/user-attachments/files/16103347/texture.patch)
> 
> Nothing renders yet, but it also doesn't crash, which is a good start. 😄

Hmmm....good catch. I will fix it now.

I thought `texture2D()` is fully removed in **WEBGL** and `texture()` is only **depricated**  in OPENGL new versions. Anyways, fixing it now. Thanks

ghutchis:
Any "deprecated" OpenGL code is not allowed in OpenGL Core Profile.

So the trick is looking around for deprecated code.

perminder-17:
> Any "deprecated" OpenGL code is not allowed in OpenGL Core Profile.
> 
> So the trick is looking around for deprecated code.

Everything else in the code looks fine to me. Is there anything else that appears deprecated maybe?

ghutchis:
@cryos - any ideas? In combination with https://github.com/OpenChemistry/avogadroapp/pull/497 the shaders compile and link completely.

But we don't see anything rendered in the widget.

cryos:
> @cryos - any ideas? In combination with [OpenChemistry/avogadroapp#497](https://github.com/OpenChemistry/avogadroapp/pull/497) the shaders compile and link completely.
> 
> But we don't see anything rendered in the widget.

Shaders are finicky, it would be looking at runtime issues, quirks or errors in updating. I had something pretty close to this but yeah I wasn't getting things to render correctly over a year back. I think the `outColor` versus `gl_FragColor` needs to be handled properly, I don't really see where `outColor` gets hooked up to anything, but it has been a while since I looked at this.

I don't see any discussion of the handling of the fragment color in this PR or the commits.

ghutchis:
Thanks - @perminder-17 also got some feedback in the OpenGL community forum .. core profile requires VAO - vertex array objects: https://www.khronos.org/opengl/wiki/Tutorial2:_VAOs,_VBOs,_Vertex_and_Fragment_Shaders_(C_/_SDL)

There seems to be some other bits that need modernization (e.g., `arrow*`) but I think the VAO is the main missing piece.

perminder-17:
Thankyou so much @cryos for your thoughts. 
>  I think the outColor versus gl_FragColor needs to be handled properly, I don't really see where outColor gets hooked up to anything,

I'm really sorry if I'm not referring to the same thing, but what I feel is:

We can associate fragment shader outputs with colour attachments using a layout directive, 
For e.g.
```
layout(location=0) out vec4 outColor;
```
Also, the same thing can happen in the CPU too by `glBindFragDataLocation` which should be called prior to linking.
> If a shader statically assigns a location to a varying out variable in the shader text, that location is used and any location assigned with glBindFragDataLocation is ignored. 
Reference: https://registry.khronos.org/OpenGL-Refpages/gl4/html/glBindFragDataLocationIndexed.xhtml

But since we're using a single colour attachment (i.e. replacing gl_FragColor rather than gl_FragData), it that necessary to explicitly specify the attachment? as any output variable defaults to attachment zero?

ghutchis:
I'm going to close this since I merged it into #2548 - thanks so much for all your work on this, I'm excited to finally merge all your great work!

perminder-17:
> I'm going to close this since I merged it into #2548 - thanks so much for all your work on this, I'm excited to finally merge all your great work!

Do let me know if anything's breaking, I'll love to commit new changes to it.

ghutchis:
The shaders look great. We're sorting out a few issues mostly related to OpenGL 4.0 core in general (e.g., some Linux users, Wayland, etc.)

Hopefully I'll be able to merge most of the other patches in the next few weeks.

Main holdup was getting the VAO support in place on the C++ side. You already ported all the shaders.

PR #1688, opened 2024-07-17
WIP: ADDING SMOOTHNESS VIA TESSELLATION SHADERS

perminder-17:
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

### DIFF
--- a/avogadro/rendering/shader.cpp
+++ b/avogadro/rendering/shader.cpp
@@ -41,7 +41,17 @@ bool Shader::compile()
     m_handle = 0;
   }
 
-  GLenum type_ = m_type == Vertex ? GL_VERTEX_SHADER : GL_FRAGMENT_SHADER;
+GLenum type_;
+if (m_type == Vertex) {
+    type_ = GL_VERTEX_SHADER;
+} else if (m_type == Fragment) {
+    type_ = GL_FRAGMENT_SHADER;
+} else if(m_type == TessellationControlShader) {
+    type_ = GL_TESS_CONTROL_SHADER;
+} else if(m_type == TessellationEvaluationShader) {
+    type_ = GL_TESS_EVALUATION_SHADER;
+}
+
   GLuint handle_ = glCreateShader(type_);
   const auto* source_ = static_cast<const GLchar*>(m_source.c_str());
   glShaderSource(handle_, 1, &source_, nullptr);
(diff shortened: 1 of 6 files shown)


avogadro/rendering/shader.cpp
```diff
@@ -41,7 +41,17 @@ bool Shader::compile()
     m_handle = 0;
   }
 
-  GLenum type_ = m_type == Vertex ? GL_VERTEX_SHADER : GL_FRAGMENT_SHADER;
+GLenum type_;
```
ghutchis, line 44: This is probably better handled through a switch statement in C++

switch (m_type) {
case Vertex:
type_ = GL_VERTEX_SHADER;
break;
case Fragment:
type_ = GL_FRAGMENT_SHADER;
// etc.
}

avogadro/rendering/mesh_opaque_fs.glsl
```diff
@@ -1,4 +1,10 @@
-varying vec3 fnormal;
+#version 430
```
ghutchis, line 1: Do we need to go to version 430? I thought the general advice was to use the lowest version of GLSL that works for your needed code / features (e.g., we should use `#version 400`)

perminder-17:
All the work for this branch is completed successfully. Here's the result of it's working. 

The main thing was to implement something like Laplacian/Gaussian smoothing to remove the artifacts. But as we moved to the algorithm we noticed that these algorithm can't be implemented via tessellation shaders. I'm not sure there's a shader construct that gives the information you'd need to do Laplace Smoothing.

So, for now, we can do use both Laplacian smoothing(at CPU) and bezier smoothing(via tessellation). Since, after this algorithm, we will have more no. of triangles, so if we apply Laplacian smoothing (maybe 1 iteration) we can have a better result that we were having without bezier smoothing.

Here's the result of my work. (you can see smooth curves instead of sharp right angles)
image

here's the maximum result we can achieve with similar or maybe a better algorithm (catmull-clark)

[BEFORE]
image

[AFTER]
image


So, the result looks excatly same. If anyone want's to test it please request the version of opengl to 4.0 (can only be tested in linux/windows not in mac because we currently lack core profile)

```c++
  defaultFormat.setVersion(4,0);
  defaultFormat.setProfile(QSurfaceFormat::CompatibilityProfile);
  defaultFormat.setOptions(QSurfaceFormat::DeprecatedFunctions);
  ```

perminder-17:
Removed the debugging pieces which was previously added for testing purpose.

Issue #901, opened 2022-05-23
Consider switching to flying edges from parallel marching cubes


ghutchis:
Right now, the mesh generation uses parallel Marching Cubes.

There's a reference implementation of parallel flying edges (outside of VTK) here:
https://github.com/sandialabs/miniIsosurface/blob/master/flyingEdges/README.md

Summary paper here: https://www.kennethmoreland.com/documents/miniIsosurface.pdf

The FE implementation was generally 1.5-2x faster than MC even in serial form.

ghutchis:
It would also be nice to push some of this down into `core` .. maybe the serial version can be in `core` and a `QtConcurrent` version can be in the `qtgui` implementation that handles threading.

aerkiaga:
Interesting... The speedup even increases on machines with many cores. I've downloaded the paper by Schroeder et al. to check the algorithm out, hopefully today.

aerkiaga:
Anyway, I can't help but think how the number of actual polygons is only O(n) to the grid's resolution. If only we could exploit that... Maybe using some knowledge about the generated surface, like overall connectivity?

aerkiaga:
Marching Cubes is indeed a neat algorithm. I hadn't heard of Flying Edges, but I'll save this task in case I have enough time. I do agree that generating the surface directly is off the table... But I've seen optimizations to MC that shave off quite a few empty cubes, so that might help too.

ghutchis:
@perminder-17 - here's the issue thread I mentioned...

perminder-17:
> @perminder-17 - here's the issue thread I mentioned...

Yess, I see. Thanks.

PR #1741, opened 2024-10-17
Use flying edges for mesh generation

perminder-17:
Resolves issue : #901 
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.


avogadro/qtgui/meshgenerator.cpp
```diff
@@ -47,66 +47,532 @@ bool MeshGenerator::initialize(const Cube* cube_, Mesh* mesh_, float iso,
...
+
+
+                m_vertices.push_back(points[globalIdx]);
+                m_vertices.push_back(points[globalIdx1]);
+                m_vertices.push_back(points[globalIdx2]);
+
+                m_normals.push_back(normals[globalIdx]);
+                m_normals.push_back(normals[globalIdx1]);
+                m_normals.push_back(normals[globalIdx2]);    
+
```
perminder-17, line 529: Hi @cryos and @ghutchis , does this part looks okay to you guys. I think I have done all things right but my output comes out to be something like. Any suggestions?

image

perminder-17:
I can fix the data structures for variables once I start getting the expected output. Also tagging @matterhorn103 and @aerkiaga for their inputs. 

Let me know if you guys have any feedbacks/suggestions?

perminder-17:
> @perminder-17 I'm afraid rendering is something I know absolutely nothing about

Aah... Thanks for your reply. No problem, I figured that out. It was like, we need to pass the triangles as well.

ghutchis:
Great, I'll start a review tomorrow

perminder-17:
Hi everyone, Here's the results after the implementation of flying-edges. 
### Marching cube takes 1.5 minutes to render the mesh
Screenshot from 2024-11-07 02-04-36

### Flying edges takes only 3 seconds to render it. (could vary with different systems)
Screenshot from 2024-11-07 02-06-40


### Also it produces a little smoother mesh (left->flying edges, right->marching cubes)

Screenshot from 2024-11-07 18-56-29

### After Laplacian smoothing we have:

Screenshot from 2024-11-10 07-17-54
Screenshot from 2024-11-10 07-18-29

ghutchis:
There's definitely a rotation / orientation bug. Try this. Go to File => Import => Download by Name and pick "caffeine" or open the attached file (after you gunzip it).

Here's the surface with the molecule:
Screenshot 2024-11-09 at 10 23 55 PM

[caffeine.cjson.gz](https://github.com/user-attachments/files/17691915/caffeine.cjson.gz)

ghutchis:
It's definitely really fast -- but there's something weird with the x, y, z orientation

avogadro/core/cube.cpp
```diff
@@ -194,6 +206,165 @@ float Cube::value(int i, int j, int k) const
...
+
+
+inline float
+Cube::getData(int i, int j, int k) const
+{
+    return m_data[(k * m_points.x() * m_points.y()) + (j * m_points.x()) + i];
+}
+
+
+std::array<std::array<float, 3>, 8> Cube::getPosCube(int i, int j, int k) const
```
perminder-17, line 324: Hmm.... I just overlooked this bug. I think here's where the issue lies. You know, when you just renders a single molecule it's position comes out to be correct, but when you have more than one mesh it's orientation is really weird. I'll check that now..... And will try to fix that.

perminder-17: Like, try to render just a CH4 it will give the correct position but as you will have more than one molecule.. It will give you wrong orientation

avogadro/core/cube.cpp
```diff
@@ -23,21 +24,32 @@ Cube::~Cube()
...
+  // Calculate the maximum dimension
+  int maxPoints = std::max({points.x(), points.y(), points.z()});
+  Vector3i equalPoints(maxPoints, maxPoints, maxPoints);
+
+  // Recalculate spacing based on equal points
   Vector3 delta = max_ - min_;
-  m_spacing =
-    Vector3(delta.x() / (points.x() - 1), delta.y() / (points.y() - 1),
-            delta.z() / (points.z() - 1));
+  m_spacing = Vector3(delta.x() / (equalPoints.x() - 1),
```
perminder-17, line 42: Okay, I can recheck and could say here's the issue lies. 

When I was having different dimensions (as a non-uniform grid), lets say `m_points.x() != m_points.y() != m_points.z()` it was showing me a discrete mesh. (looking like the parts of the mesh has been seperated). 

So I thought to use the equal points as a workaround but it leads to the wrong orientation. Any tip for fixing it? @ghutchis

ghutchis: Okay, but that doesn't work. Most "cubes" are going to be rectangular solids. They have equal spacing, but different number of points along different axes. Caffeine is a good example here, since it's fairly flat, but kinda a rectangle in the other two dimensions.

CH4 or water aren't great examples because they're roughly cubic.

This code *definitely* shouldn't change. You can't compute the spacing this way because the number of points will be different along different axes.

I see the weird "mesh" when I revert the changes to this code, but that just means the bug is somewhere else.

perminder-17: Yes... I got it. Will look into the code again and try to find out the issue tomorrow. Thanks for your reply.

perminder-17:
Hi @ghutchis , there was a problem on the indexing part. It was really tough to debug the problem haha. Managing all of the offsets and indices is what makes the algorithm tricky to implement :"). Could you please recheck once again. I can do more cleanups in the code once the results looks okay to you. I  have done the testing from my side and that looks good to me. Let me know if I am missing any testcase

## Caffine

Screenshot from 2024-11-12 15-39-36

cryos:
Looks nice, one thing to note is that debug numbers are far less interesting than numbers for a release build. Busy this week, but interested to see the performance comparison for a release build. Flying edges had huge gains when using it in other projects, awesome to see it implemented independently here (I had been testing it in the VTK window previously with great performance).

ghutchis:
I'm getting 
```
ld: Undefined symbols:
  Avogadro::Core::Cube::getData(int, int, int) const, referenced from:
      Avogadro::QtGui::MeshGenerator::FlyingEdgesAlgorithmPass1() in meshgenerator.cpp.o
```

Looks like there isn't an implementation of `Cube::getData(int, int, int)` in the commit

avogadro/core/cube.cpp
```diff
@@ -193,9 +201,159 @@ float Cube::value(int i, int j, int k) const
...
+    vals[4] = getData(i, j, k + 1);
+    vals[5] = getData(i + 1, j, k + 1);
+    vals[6] = getData(i + 1, j + 1, k + 1);
+    vals[7] = getData(i, j + 1, k + 1);
+
+    return vals;
+}
+
+
+inline float Cube::getData(int i, int j, int k) const
```
perminder-17, line 302: Is that a warning or an error? @ghutchis? Because here's the implementation of getData(int, int ,int). 

My code runs and compile correctly. Can you confirm it once?

perminder-17:
> (maybe this is release - it isn't clear).

That's in build mode. 
> 
> Flying edges had huge gains when using it in other projects, awesome to see it implemented independently here (I had been testing it in the VTK window previously with great performance)

Yes, I can see noticeable speed ups in the mesh generation now.  Thanks @cryos

avogadro/core/cube.cpp
ghutchis: Yes, it's a build error. It's failing on my machine and all build actions.

If I remove the `inline` and assume the compiler will handle it, everything compiles fine.

cryos:
> > (maybe this is release - it isn't clear).
> 
> That's in build mode.

All I see is debug mode in what you posted, if that is the case my point stands - benchmarks in debug mode are not very useful and we know the STL etc is very slow in debug. Comparing release numbers (which I also expect to be good) would be more informative.

ghutchis:
I'm happy to help do some performance tests - I'm not seeing the "Time taken" code at the moment in your pull request.

I have a range of molecule sizes and resolutions to give this a good benchmark.

Doing a rough stopwatch count, I'm seeing ~2-2.5x performance improvement, but that's including the whole thing, not just mesh generation.

I can say that with caffeine as an example, the flying edges version appears instantly on my MacBook even at "Very High" resolution. With a larger molecule, it's going from ~10s to display the VdW mesh to ~4.5s.

If you want to send me the patch you used for timing, I'd be happy to perform some real benchmarks tonight in release mode.

perminder-17:
> All I see is debug mode in what you posted, if that is the case my point stands - benchmarks in debug mode are not very useful and we know the STL etc is very slow in debug. Comparing release numbers (which I also expect to be good) would be more informative.

Thanks @cryos for your insights. Really helps a lot and also thanks @ghutchis for the help in performance testing.

If I could say what code I used for checking the time, I used the chrono library https://www.geeksforgeeks.org/chrono-in-c/

If you include header `#include<chrono>` and replace your run() function with this, you can see the time taken.

```cpp
#include <chrono> // at the top (header file)



// Rest code remains the same


// Replace the run function to test
 
void MeshGenerator::run() // you can directly copy the run() function and paste it with the original ones
{

  if (!m_cube || !m_mesh) {
    qDebug() << "No mesh or cube set - nothing to find isosurface of…";
    return;
  }
  m_mesh->setStable(false);
  m_mesh->clear();

  auto start = std::chrono::high_resolution_clock::now();
  FlyingEdgesAlgorithmPass1();
  FlyingEdgesAlgorithmPass2();
  FlyingEdgesAlgorithmPass3();
  FlyingEdgesAlgorithmPass4();

  auto end = std::chrono::high_resolution_clock::now();
  std::chrono::duration<double> elapsed = end - start;
  qDebug() << "Flying Edges Time: " << elapsed.count() << " s";    

  m_mesh->setVertices(points);
  m_mesh->setNormals(normals);
  m_mesh->setTriangles(tris);
  m_mesh->smooth(m_passes);
  m_mesh->setStable(true);

  points.resize(0);
  normals.resize(0);
  tris.resize(0);

}
``` 
Screenshot from 2024-11-13 02-37-19


also, if you want to test and compare the time with marching cubes. 

here how you can check the time taken for marching cubes as well. 

### OLD CODE

```cpp

#include <chrono> // similarly include this

// just a little changes in the run function 

void MeshGenerator::run()
{

// rest code remains the same.

 auto start = std::chrono::high_resolution_clock::now(); // add this

  for (int i = 0; i < m_dim.x() - 1; ++i) {
    for (int j = 0; j < m_dim.y() - 1; ++j) {
      for (int k = 0; k < m_dim.z() - 1; ++k) {
        marchingCube(Vector3i(i, j, k));
      }
    }
    if (m_vertices.capacity() < m_vertices.size() + m_dim.y() * m_dim.x() * 3) {
      m_vertices.reserve(m_vertices.capacity() * 2);
      m_normals.reserve(m_normals.capacity() * 2);
    }
    emit progressValueChanged(i);
  }

// add this
  auto end = std::chrono::high_resolution_clock::now();
  std::chrono::duration<double> elapsed = end - start;
  qDebug() << "Marching Cube Timing: " << elapsed.count() << " s";    
}

```

perminder-17:
> With a larger molecule, it's going from ~10s to display the VdW mesh to ~4.5s.

Also, after the merge of this, I can handle the merge conflicts (if any) in this PR https://github.com/OpenChemistry/avogadrolibs/pull/1692. If we add a progressive refinement, it won't take even 4.5 seconds to render and will generate the mesh in no time. 

Do let me know if everything looks fine in the results, I could do many cleanups in the code (added too many debugging peices, unnecessary functions, variables haha). Still needs some formatting and cleanups.

perminder-17:
Btw @ghutchis I just noticed one bug (when you click on the calculate button twice for the same mesh generation it takes more time). It's because we were not clearing the memory. In the above commit I have fixed that as well. Sorry for the oversight.

ghutchis:
I'm seeing ~2x - 2.2x speedup for mesh generation on a range of 7 molecules with 0.05Å resolution. It depends a lot on the size of the molecule and the shape (e.g., smaller molecules and more spherical / cubic see the greatest speedup).

I'll do a bit of profiling, but of course this is also the single-core implementation (vs. parallel)

perminder-17:
IMG-20241113-WA0002


Correct! I again looked at the results mentioned on the paper. They also have 2-2.2x speedups. 

I'll do some cleanups in the code tomorrow.

ghutchis:
I also checked the number of vertices for marching cubes vs. flying edges. It's getting exactly 6.0 times fewer vertices, so the mesh is much smaller at the same resolution (and thus faster to render, less memory, etc.)

Even on a really big DNA chunk, I'm seeing ~0.1s for mesh creation with the "automatic" resolution. (And yes, the progressive refinement could be a help here too.)

avogadro/core/cube.h
```diff
@@ -235,6 +239,45 @@ class AVOGADROCORE_EXPORT Cube
...
+   * Compute the gradient at a specific point in the cube.
+   * @param i x index
+   * @param j y index
+   * @param k z index
+   * @return Gradient vector at the specified point.
+   */
+  std::array<float, 3> computeGradient(int i, int j, int k) const;
+
+  /**
+   * Get the values of the eight corners of a cube cell.
```
ghutchis, line 252: I'm not quite sure I understand from the docs here. I've got some index (i, j, k) into the cube. What's the "cell" of the cube? Is it just going to get me the corners of the entire cube? If so, why do I need i, j, k parameters?

avogadro/core/cube.h
```diff
@@ -235,6 +239,45 @@ class AVOGADROCORE_EXPORT Cube
...
+   * @param i x index
+   * @param j y index
+   * @param k z index
+   * @return Array of values at the eight corners.
+   */
+  std::array<float, 8> getValsCube(int i, int j, int k) const;
+
+
+  std::array<std::array<float, 3>, 8> getGradCube(int i, int j, int k) const;
```
ghutchis, line 261: Needs a documentation comment like the other methods.

avogadro/core/cube.h
```diff
@@ -235,6 +239,45 @@ class AVOGADROCORE_EXPORT Cube
...
+   * Get the data value at the specified indices.
+   * @param i x index
+   * @param j y index
+   * @param k z index
+   * @return Value at the specified indices.
+   */
+  float getData(int i, int j, int k) const;
+
+  /**
+   * Get the positions of the eight corners of a cube cell.
```
ghutchis, line 273: Again, what's the "cell" here? Are these the min, max of the corners of the cube? If so, why do you need the i, j, k?

avogadro/core/mesh.h
```diff
@@ -141,10 +145,10 @@
...
    * @return The number of normals.
    */
-  unsigned int numNormals() const
-  {
-    return static_cast<unsigned int>(m_normals.size());
-  }
+  // unsigned int numNormals() const
+  // {
+  //   return static_cast<unsigned int>(m_normals.size());
+  // }
```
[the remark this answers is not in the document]
ghutchis: Why is this commented out? Can't we keep this method?

avogadro/core/mesh.h
```diff
@@ -228,6 +232,7 @@ class AVOGADROCORE_EXPORT Mesh
   Core::Array<Vector3f> m_vertices;
   Core::Array<Vector3f> m_normals;
   Core::Array<Color3f> m_colors;
+  Core::Array<Vector3f> tris;
```
ghutchis, line 235: Please use `m_triangles` to be consistent with the other variables.

avogadro/qtgui/meshgenerator.h
```diff
@@ -81,11 +81,77 @@ class AVOGADROQTGUI_EXPORT MeshGenerator : public QThread
...
+        int xr;
+
+        // modified on pass 2
+        // set on pass 3
+        int xstart;
+        int ystart;
+        int zstart;
+    };
+
+
```
ghutchis, line 110: To help with formatting the documentation, please add this to group the methods.

```
 /**
   * @name Flying Edges
   * Methods to implement the "flying edges" method for isosurface mesh generation.
   * Flying edges: A high-performance scalable isocontouring algorithm
   * Schroeder; Maynard; Geveci; 
   * 2015 IEEE 5th Symposium on Large Data Analysis and Visualization (LDAV)
   * [10.1109/LDAV.2015.7348069](https://doi.org/10.1109/LDAV.2015.7348069)
   * Alternate (non-VTK) implementation at
   * https://github.com/sandialabs/miniIsosurface/blob/master/flyingEdges/
   * @{
```

avogadro/qtgui/meshgenerator.h
```diff
@@ -81,11 +81,77 @@ class AVOGADROQTGUI_EXPORT MeshGenerator : public QThread
...
+  * Pass3 computes cumulative offsets for triangles 
+  * and vertices and allocates memory for the mesh structures.
+  */
+  void FlyingEdgesAlgorithmPass3();
+
+  /**
+  * Calculates normals, triangles and vertices.    
+  */
+  void FlyingEdgesAlgorithmPass4();
+
```
ghutchis, line 135: Then add
```
  // end Flying Edges doxygen group
  /**@}*/
```

avogadro/qtgui/meshgenerator.h
```diff
@@ -107,34 +173,44 @@ class AVOGADROQTGUI_EXPORT MeshGenerator : public QThread
   int progressMaximum() { return m_progmax; }
 
 signals:
+
   /**
    * The current value of the calculation's progress.
    */
   void progressValueChanged(int);
 
 protected:
```
ghutchis, line 182: Can we keep this? Sometimes it's helpful to get normals for arbitrary points.

avogadro/qtgui/meshgenerator.h
```diff
@@ -107,34 +173,44 @@ class AVOGADROQTGUI_EXPORT MeshGenerator : public QThread
...
-  /**
-   * Get the normal to the supplied point. This operation is quite expensive
-   * and so should be avoided wherever possible.
-   * @param pos The position of the vertex whose normal is needed.
-   * @return The normal vector for the supplied point.
-   */
-  Vector3f normal(const Vector3f& pos);
+
+  unsigned long duplicate(const Vector3i& c, const Vector3f& pos);
```
ghutchis, line 184: Needs documentation.

avogadro/qtgui/meshgenerator.h
```diff
@@ -144,26 +220,24 @@ class AVOGADROQTGUI_EXPORT MeshGenerator : public QThread
...
   Vector3i m_dim;           /** The dimensions of the cube. */
-  Core::Array<Vector3f> m_vertices, m_normals;
-  Core::Array<unsigned int> m_indices;
+
+  Core::Array<Vector3f> points, normals;  
+  std::vector<gridEdge> gridEdges; // size (m_dim.y() * m_dim.z())
+  std::vector<unsigned char> cubeCases; //size ((m_dim.x() - 1) * (m_dim.y() - 1) * (m_dim.z() - 1))
+  std::vector<int> triCounter;  // size ((m_dim.y() - 1) * (m_dim.z() - 1))
+  std::vector<unsigned char> edgeCases; // size ((m_dim.x() - 1) * (m_dim.y()) * (m_dim.z()))
+  Core::Array<Vector3f> tris; // triangles of a mesh
```
ghutchis, line 229: Please remove any un-used variables (from Marching Cubes)

Please also use `m_triangles` (or similar) to keep consistent with member variables through the rest of the code.

avogadro/qtgui/meshgenerator.h
```diff
@@ -144,26 +220,24 @@ class AVOGADROQTGUI_EXPORT MeshGenerator : public QThread
...
-  static const int a2iTetrahedronEdgeConnection[6][2];
-  static const int a2iTetrahedronsInACube[6][4];
-  static const long aiTetrahedronEdgeFlags[16];
-  static const int a2iTetrahedronTriangles[16][7];
-  static const long aiCubeEdgeFlags[256];
-  static const int a2iTriangleConnectionTable[256][16];
+  static const unsigned char numTris[256]; 
+  static const bool isCut[256][12];  
+  static const char caseTriangles[256][16];
+  static const unsigned char edgeVertices[12][2];
```
ghutchis, line 240: The static analysis says these class members are never used?
If they're used, please keep to `m_edgeVertices` etc. for consistency.

ghutchis (changes requested):
Mostly some small things with the comments and documentation consistency.

avogadro/core/cube.cpp
```diff
@@ -194,8 +193,159 @@ float Cube::value(int i, int j, int k) const
...
+    return m_data[(i * ny * nz) + (j * nz) + k];
+}
+
+
+std::array<std::array<float, 3>, 8> Cube::getPosCube(int i, int j, int k) const
+{
+
+  std::array<std::array<float, 3>, 8> pos;
+
+  float xpos = m_min.x() + (i * m_spacing.x());
```
ghutchis, line 308: I'm sorry - I still don't understand what this method is supposed to do.

xpos, ypos, zpos are the real-space versions of i, j, k indices
.. so the method gives me a cube that's one step in each of the x, y, z directions?

If so, shouldn't that be the doc string?

avogadro/core/cube.h
```diff
@@ -23,9 +24,7 @@ class Mutex;
 /**
  * @class Cube cube.h <avogadro/core/cube.h>
  * @brief Provide a data structure for regularly spaced 3D grids.
- * @author Marcus D. Hanwell
```
ghutchis, line 26:
```
@author Marcus D. Hanwell
@author Perminder Singh
```

avogadro/core/mesh.h
```diff
@@ -138,14 +142,6 @@ class AVOGADROCORE_EXPORT Mesh
    */
   const Core::Array<Vector3f>& normals() const;
 
-  /**
```
ghutchis, line 141: I still don't understand why you're removing `numNormals()` which seems like a useful method.

ghutchis (commented):
Mostly resolved - please note the other comments, thanks.

avogadro/core/mesh.h
perminder-17: But i was unable to find where we were using numnormals() in our actual code?

ghutchis: We're not, currently. But since this is a `core` class, it's also API for future development.

avogadro/qtgui/meshgenerator.cpp
```diff
@@ -47,66 +47,460 @@ bool MeshGenerator::initialize(const Cube* cube_, Mesh* mesh_, float iso,
...
+        {
+          ge1.xstart += isCutCase[2];
+          ge1.zstart += isCutCase[10];
+        }
+        if(isZEnd)
+        {
+          ge2.xstart += isCutCase[4];
+          ge2.ystart += isCutCase[7];
+        }
+        if(isXEnd and isYEnd)
```
ghutchis, line 170: `&&` not `and` - this is the build failure on Windows

avogadro/qtgui/meshgenerator.cpp
```diff
@@ -47,66 +47,460 @@ bool MeshGenerator::initialize(const Cube* cube_, Mesh* mesh_, float iso,
...
+        {
+          ge1.zstart += isCutCase[11];
+        }
+        if(isXEnd and isZEnd)
```
ghutchis, line 174: Same - use `&&`

avogadro/qtgui/meshgenerator.cpp
```diff
@@ -47,66 +47,460 @@ bool MeshGenerator::initialize(const Cube* cube_, Mesh* mesh_, float iso,
...
+        {
+          ge2.ystart += isCutCase[5];
+        }
+        if(isYEnd and isZEnd)
```
ghutchis, line 178: `&&`

avogadro/qtgui/meshgenerator.cpp
```diff
@@ -47,66 +47,460 @@ bool MeshGenerator::initialize(const Cube* cube_, Mesh* mesh_, float iso,
...
+                    m_normals[idx] = Vector3f(interpolatedNormal[0], interpolatedNormal[1], interpolatedNormal[2]);
+                }
+                globalIdxs[7] = idx;
+                ++y2counter;
+            }
+
+            if(isCutCase[11])
+            {
+                int idx = ge1.zstart + z1counter;
+                if(isXEnd and isYEnd)
```
ghutchis, line 413: `&&` not `and`

avogadro/qtgui/meshgenerator.cpp
```diff
@@ -47,66 +47,460 @@ bool MeshGenerator::initialize(const Cube* cube_, Mesh* mesh_, float iso,
...
+                    m_vertices[idx] = Vector3f(interpolatedPoint[0], interpolatedPoint[1], interpolatedPoint[2]);
+                    m_normals[idx] = Vector3f(interpolatedNormal[0], interpolatedNormal[1], interpolatedNormal[2]);
+                }
+                globalIdxs[11] = idx;
+            }
+
+            if(isCutCase[5])
+            {
+                int idx = ge2.ystart + y2counter;
+                if(isXEnd and isZEnd)
```
ghutchis, line 428: `&&` not `and`

avogadro/qtgui/meshgenerator.cpp
```diff
@@ -47,66 +47,460 @@ bool MeshGenerator::initialize(const Cube* cube_, Mesh* mesh_, float iso,
...
+                    m_normals[idx] = Vector3f(interpolatedNormal[0], interpolatedNormal[1], interpolatedNormal[2]);
+                    // y2 counter does not need to be incremented.
+                }
+                globalIdxs[5] = idx;
+            }
+
+            if(isCutCase[6])
+            {
+                int idx = ge3.xstart + x3counter;
+                if(isYEnd and isZEnd)
```
ghutchis, line 444: `&&` not `and`

ghutchis (changes requested):
Sorry, I missed these last time. I think this fixes the compile on Windows.

perminder-17:
Done @ghutchis

PR #1746, opened 2024-10-19
ENH: Additional Secondary Structures

TactfulDeity:
# Enhancements
- added 3-10 and pi helices

Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

### DIFF
--- a/avogadro/core/secondarystructure.cpp
+++ b/avogadro/core/secondarystructure.cpp
@@ -49,14 +49,26 @@ void SecondaryStructureAssigner::assign(Molecule* mol)
       // check to see how far apart the residues are
       int separation = std::abs(int(hBond->residue - hBond->residuePair));
 
-      // just alpha for now
-      if (separation == 4) {
-        m_molecule->residue(hBond->residue)
-          .setSecondaryStructure(Residue::SecondaryStructure::alphaHelix);
+      switch (separation) {
+        case 3: { // 3-10
+          m_molecule->residue(hBond->residue)
+            .setSecondaryStructure(Residue::SecondaryStructure::helix310);
+          break;
+        }
+        case 4: { // alpha
+          m_molecule->residue(hBond->residue)
+            .setSecondaryStructure(Residue::SecondaryStructure::alphaHelix);
+          break;
+        }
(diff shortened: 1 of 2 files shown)


ghutchis:
No idea why the self-hosted runner is failing. Likely a transient since the error message is about creating the disk image, not the compile.

TactfulDeity:
I would just download the log and force an actions re-reun on that one specifically to see if it is consistently reproducible

TactfulDeity:
So this is a niche bug but I came across this forum where the exact issue is mentioned back in 2012:
https://cmake.org/pipermail/cmake-developers/2012-February/003319.html

TactfulDeity:
I noticed the error code it is reporting is `-1401`,  which is the system error code that is reported via CPack^1. So I dug around and found out this corresponds to a specific error code on Mac, [thread here](https://discussions.apple.com/thread/4940204?sortBy=rank)
`  errFSBadFSRef                 = -1401, /* FSRef parameter is bad */`

The `FSRefParam` unfortunately has no public documentation I could find aside from [this](https://developer.apple.com/documentation/coreservices/fsrefparam). Each of the listed `instance properties` seem to function as "member variables" in the Apple/cocoa/Objective-C ecosystem as demonstrated in [this stack overflow post](https://stackoverflow.com/questions/719788/property-vs-instance-variable). One of them is `nameLength` that likely coincides with the issue mentioned in the previous post of a volume name being to long. This means one of the other 21 variables could be the issue causing it, or the length issue made a comeback I am not sure. Without a ARM Mac on hand and not much of an error message to go off of I am stuck. 

^1 see lines 478-494 [here](https://github.com/Kitware/CMake/blob/master/Source/CPack/cmCPackDragNDropGenerator.cxx) in the Cmake source.

ghutchis:
I'm good to merge. I can check the runner in person on Monday, but the build before was fine -- and it clearly has nothing related to the change.

(The MacOS runners seem to have periodic transient issues with generating the DMG.)

ghutchis (approved):

Issue #1464, opened 2023-11-22
AppImage built against deprecated version of OpenSSL causing problems


matterhorn103:
Originally described in the forums in thread [#4995](https://discuss.avogadro.cc/t/challenges-with-python-extensions/4995).

**Avogadro version: (please complete the following information from the About box):**
 - Avogadrolibs: 1.98.1
 - Qt: 5.15.2

**Desktop version: (please complete the following information):**
 - OS: OpenSUSE Tumbleweed
 - Version: latest snapshot (20231117)

**Describe the bug**
When starting the AppImage, the following lines are printed to standard output:
```
qt.network.ssl: QSslSocket: cannot resolve EVP_PKEY_base_id
qt.network.ssl: QSslSocket: cannot resolve SSL_get_peer_certificate
```

The result of this is that the plugins manager doesn't work. Possibly other things too, but I didn't find any.

Some googling revealed the problem. Many distros apparently don't ship with OpenSSL 1.1 any more but instead >=3.0, and if a Qt5 program has been built against v1.1, QSsl doesn't work. See [here](https://github.com/kiwix/kiwix-desktop/issues/871) and [here](https://bugreports.qt.io/browse/QTBUG-115146).

1.1 is still available in the OpenSUSE repos, so I installed it, and the problem was then solved, the error message was gone, and plugin installation works as expected.

From what I can find, it seems like as of 22.04 OpenSSL 1.1 is not even available in the Ubuntu repos any more, as it is deprecated. Given Ubuntu's popularity this likely means plugins are currently broken for a great number of Linux users. What the situation on other distros is I do not know.

Presumably just changing the GitHub workflow to build on Ubuntu 22.04 rather than the current 20.04 would solve the issue.

**To Reproduce**
Entertainingly, uninstalling the 1.1 package didn't bring back the error. Possibly a restart would have brought it back, but I didn't check.

ghutchis:
IIRC AppImage insists upon building from 20.04. "Build on old, run on newer" but that will mean bundling OpenSSL in the build.

ghutchis:
Can you poke around in the AppImage and see if it has libssl* and libcrypto* or those need to be added to the bundle? I'm seeing lots of messages on this, but conflicting information.

(Or better yet, can I get a list of all files in the AppImage?)

matterhorn103:
[...]

PR #1749, opened 2024-10-24
Bump to Ubuntu 22.04 LTS for AppImage builds

matterhorn103:
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

### DIFF
--- a/.github/workflows/build_cmake.yml
+++ b/.github/workflows/build_cmake.yml
@@ -37,7 +37,7 @@ jobs:
           }
         - {
             name: "Ubuntu AppImage", artifact: "Avogadro2.AppImage",
-            os: ubuntu-20.04,
+            os: ubuntu-22.04,
             cc: "gcc", cxx: "g++",
             build_type: "Release",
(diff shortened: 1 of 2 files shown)


matterhorn103:
This is in line with the policy outlined [here](https://discuss.avogadro.cc/t/appimage-ssl-issues/5260/6) of building for the Ubuntu LTS release *prior* to the current one. Now that 24.04 is available, we should move to 22.04.

Of key significance is that this should fix the SSL issues on Linux (closes #1464), as 22.04 has OpenSSL 3.0 and therefore the Qt it provides will also be built against OpenSSL 3.0. Anyone who was running the AppImage on 22.04 or other modern distros won't have been able to use the plugin manager, but will now have that restored.

ghutchis:
As long as the AppImage builder doesn't complain, I'm good. (At one point, the AppImage tool was very picky about what OS it ran on.)

ghutchis:
But you also have this change in your other pull request. I can't merge both, so I'm going to close this one.

matterhorn103:
> But you also have this change in your other pull request. I can't merge both, so I'm going to close this one.

Maybe I misunderstood how git works then, I thought it'd be fine since it's the same commit in each branch. Fair enough!

PR #1772, opened 2024-11-10
Add flatpak workflow for GitHub Actions

matterhorn103:
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

### DIFF
--- a/.github/workflows/flatpak.yml
+++ b/.github/workflows/flatpak.yml
@@ -16,47 +16,48 @@ jobs:
       with:
         repository: openchemistry/openchemistry
         submodules: false
+        path: src
         
     - name: Checkout avogadroapp
       uses: actions/checkout@v4
       with:
         repository: openchemistry/avogadroapp
-        path: avogadroapp
+        path: src/avogadroapp
         
     - name: Checkout avogadrolibs
       uses: actions/checkout@v4
       with:
-        path: avogadrolibs
+        path: src/avogadrolibs
 
     - name: Checkout i18n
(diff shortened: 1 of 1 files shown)


ghutchis:
It's giving this error:
```
** (flatpak-builder:1007): ERROR **: 17:44:23.944: 
Failed to load included manifest (/__w/avogadrolibs/avogadrolibs/avogadroapp/flatpak/shared-modules/glew/glew.json): 
Failed to open file ?/__w/avogadrolibs/avogadrolibs/avogadroapp/flatpak/shared-modules/glew/glew.json?: No such file or directory
```

matterhorn103:
Yep I saw. It's looking for the shared module relative to the path of the manifest and not to the root.

I could open a PR to keep the manifest in the top level of `openchemistry` (many projects do this), but then I would have to find a way to overwrite that manifest when building the release flatpaks.

I've added a step that manually moves the manifest up to the `openchemistry` top level, and I'm hoping that does the trick, as the source tree is then set up identically to the normal flatpak builds then I think?

matterhorn103:
This is blocked by https://github.com/OpenChemistry/avogadroapp/pull/523 as it relies on the new manifest being in the correct location, so the build won't succeed just yet.

ghutchis:
Let me know when this is ready for review, thanks.

matterhorn103:
Now that the other PR has been merged, this should be ready now!

PR #1810, opened 2024-11-21
Edit molecule name, charge, and spin in properties

### LINKED ISSUE #1059: Total Charge / Spin Multiplicity
File import / export and input generators need better support for total charge / spin multiplicity.

If I read in a calculation, the program should know *throughout* that the total charge / spin is set (e.g., H3O+ or OH-, etc.)

PR #1058 gets things started
- [x] `quantumio` readers need update
- [x] input generator scripts should get total charge and spin through JSON

ghutchis:
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

### DIFF
--- a/avogadro/core/variant-inline.h
+++ b/avogadro/core/variant-inline.h
@@ -411,7 +411,8 @@ inline const Vector3& Variant::value() const
   if (m_type == Vector)
     return *m_value.vector;
 
-  return Vector3::Zero();
+  static Vector3 nullVector(0, 0, 0);
+  return nullVector;
 }
 
 inline void Variant::clear()
@@ -422,6 +423,9 @@ inline void Variant::clear()
   } else if (m_type == Matrix) {
     delete m_value.matrix;
     m_value.matrix = 0;
+  } else if (m_type == Vector) {
+    delete m_value.vector;
+    m_value.vector = 0;
   }
 
   m_type = Null;
@@ -527,6 +531,8 @@ inline Variant& Variant::operator=(const Variant& variant)
       m_value.string = new std::string(variant.toString());
     else if (m_type == Matrix)
       m_value.matrix = new MatrixX(*variant.m_value.matrix);
+    else if (m_type == Vector)
+      m_value.vector = new Vector3(*variant.m_value.vector);
     else if (m_type != Null)
       m_value = variant.m_value;
   }
--- a/avogadro/molequeue/inputgeneratorwidget.cpp
+++ b/avogadro/molequeue/inputgeneratorwidget.cpp
@@ -64,8 +64,10 @@ void InputGeneratorWidget::setMolecule(QtGui::Molecule* mol)
     m_molecule->disconnect(this);
 
   m_molecule = mol;
-
   if (mol) {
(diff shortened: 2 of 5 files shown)


ghutchis:
Thoughts @matterhorn103

matterhorn103:
Looks good to me! :)

The first thing I notice is that the spin can go lower than 1, which obviously it shouldn't.

Secondly, we know for a fact that some combinations of charge and spin are not valid based on the proton count, so is it possible that we can have the +/- arrows be a bit smarter? By that I mean:

1. Every time the charge is changed by 1 using the arrows the spin should change by 1 to match, either from 1 to 2 or 2 to 1 (i.e. the spin is always the lowest valid value)
2. When the spin is changed with the arrows it should go in steps of 2 (and the charge is not changed). That way the arrows cycle between the valid spin states. For example incrementing the spin of a singlet would make it a triplet. This halves the number of necessary clicks!
3. None of this should apply to when the numbers are typed, where the user should be able to choose freely just as they can now - this is important so they can override any errors in Avogadro's logic

If you were *particularly* keen for suggestions I'd also say that in an ideal world you would see the +/- arrows even before you click the field so it's clear they're editable, like how it is in the Manipulate Tool for example. But I suspect your choice of widgets is limited in the table?

ghutchis:
> The first thing I notice is that the spin can go lower than 1, which obviously it shouldn't.

Ugh, that was a typo. Sorry.

> When the spin is changed with the arrows it should go in steps of 2 (and the charge is not changed). That way the arrows cycle between the valid spin states. For example incrementing the spin of a singlet would make it a triplet. This halves the number of necessary clicks!

I think you mean in the input generators? At least for me, the table doesn't show any kind of arrows. There is only minimal validation at the moment.

I get your idea (i.e., to make sure the number is always valid) but I think I'd still leave it to the user. What if I have a molecule and want to run the radical cation. Do I have to change the charge before I can change the spin state? That seems non-intuitive. Some sort of color or warning that the spin state and charge are inconsistent might make sense though.

There's some code in the `Molecule` class that works like you describe. If I draw out a molecule and change the charge to +1 or -1, it will change the auto-defined multiplicity to 2 (and back to 1) according to the number of electrons.

matterhorn103:
> I think you mean in the input generators? At least for me, the table doesn't show any kind of arrows.

No, I mean in the molecular properties table. They only show up when I click to edit the fields, but they're there.

> What if I have a molecule and want to run the radical cation. Do I have to change the charge before I can change the spin state?

The way I'm suggesting, you would in this case never have to change the spin state, because it would automatically be the radical cation after increasing the charge.

Maybe the idea would be more intuitive if it was separate buttons  labelled as what they actually are: "add/remove an electron" and "increase/decrease spin".

> There's some code in the `Molecule` class that works like you describe. If I draw out a molecule and change the charge to +1 or -1, it will change the auto-defined multiplicity to 2 (and back to 1) according to the number of electrons.

Good to know :)

> But some of this is probably better on the forum than buried in the pull request comments.

Probably, I can open a thread tomorrow if you don't.

PR #1853, opened 2024-12-08
Fixing Molecular orbital rendering with Flying Edges

perminder-17:
For the second mesh, we were still using old triangles array from the first mesh.

When we switch to the second mesh, we also fetch its triangles by calling `triangles2 = mesh->triangles();`. This ensures that the second mesh is rendered with its own vertices, normals, and triangle indices, not the first mesh's data.

Here's the result:
Screenshot from 2024-12-09 04-54-06

### DIFF
--- a/avogadro/qtplugins/meshes/meshes.cpp
+++ b/avogadro/qtplugins/meshes/meshes.cpp
@@ -66,52 +66,60 @@ void Meshes::process(const QtGui::Molecule& mol, GroupNode& node)
     auto* geometry = new GeometryNode;
     node.addChild(geometry);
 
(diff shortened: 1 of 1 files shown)


ghutchis:
I can confirm this fixes the bug on anything I've tried. I'll do some more testing tomorrow to be sure.

perminder-17:
> I can confirm this fixes the bug on anything I've tried. I'll do some more testing tomorrow to be sure.

Sure.. Do let me know if anything still breaks.

ghutchis:
Looks good in further testing.

PR #1869, opened 2024-12-16
Updates to workflows

matterhorn103:
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.


matterhorn103:
This in part is continuing recent changes.

- Separate workflow matrices for Windows, macOS, Linux, Flatpak
- Got rid of the old CMake and Qt6 matrices altogether, moved the workflows we still need into the OS-based workflows
- Release builds on Linux are Qt6
- Uniformity of:
  - Workflow names
  - Artifact names
  - Environment variables
  - Build flags
  - Steps
- Identical workflows between `avogadroapp` and `avogadrolibs`

matterhorn103:
The macOS Intel build fails due to some issue with the packaging step, which I'll look into (it should run identically to the current Mac release builds so it ought to work fine).

The two builds on `macos-latest` i.e. M1 CPUs fail at the VTK step. @ghutchis does VTK not normally compile fine on your Mac?

ghutchis:
Mac builds sometimes fail when it tries to create the DMG file. It's a known bug with CI builds. You can either set up a retry loop or upgrade to a newer macOS - although even then...

I'm definitely wondering if we should just have a zipped Mac app, which is the other common way to distribute Mac apps and would definitely be easier.

I've got a bunch going on today, but I'll get back to this when I can.

ghutchis:
If you're wondering - I just don't want to change the build plan before a release. 😉

ghutchis:
Should we keep this and update or drop and start a new pull request?

matterhorn103:
Probably better to start fresh from the current state of things and assess the ideas one by one, I'll close it.

PR #1911, opened 2025-01-04
Fixes#1714 => Functionality for setting size of ghost atom

peach280:
Modified elements.cpp to allow for individual radii for each custom element.
Added methods to set the radii for custom elements.
Updated the methods that return the radii to use the vectors instead of the single values.


ghutchis:
This looks fine from a `core` perspective. At some point, it needs UI to set the radii.

Before I merge, it would be great to have a test or two in https://github.com/OpenChemistry/avogadrolibs/blob/master/tests/core/elementtest.cpp

peach280:
I am working on this but I just wanted to clarify a small doubt, does element 0 mean neutronium? Like the edge case should be for neutronium?

ghutchis:
Most chemistry packages have a element 0 "Xx" or "Du" - it's a placeholder. In Avogadro, it's used to indicate a centroid point, for example.

peach280:
I have committed some changes into this including edge case for element 0. Please review.

matterhorn103:
@ghutchis Is this worth taking another look at? From your comments, it sounds like it was about ready

PR #1939, opened 2025-01-13
ONLY FOR TESTING (VOLUMETRIC SHADERS)

perminder-17:
I'll update it into new class when the desired results would be achieved. It's a WIP PR for now.


avogadro/rendering/solidpipeline.cpp
```diff
@@ -85,101 +229,407 @@ void initializeFramebuffer(GLuint* outFBO, GLuint* texRGB, GLuint* texDepth)
...
-  // Draw to screen
+  // std::cout<<background<<std::endl;
+  // Bind default FBO again
+
+  Avogadro::Core::Cube* m_cube;
+
+  if(m_cube){
+    std::cout<<"yes I am present"<<std::endl;
+    std::cout<<"data-size" << m_cube->data()->size()<<std::endl;  
+  }
```
perminder-17, line 522: Hi @ghutchis , as we discussed on the call I think I also don't get the size of the cube's data. 

Screenshot from 2025-02-27 04-21-11

The logged value shows that I have my cube present but as soon as I try to acess the m_cube->data() or want to check it's size, my software crashes. Also, applying a condition `if(m_cube->data())` still don't works in this case, so I am unable to get it's data.

avogadro/rendering/solidpipeline.h
```diff
@@ -50,13 +63,25 @@ class SolidPipeline
    * @brief Get or set whether Ambient Occlusion is enabled.
    */
   bool getAoEnabled() { return m_aoEnabled; }
+
+  void setData(float data);
+  float getData() const;
+
+  // Corrected setCube: assign cubing to the member variable m_cube.
+  // void setCube(Avogadro::Core::Cube* cubing) { m_cube = cubing; }
+  // Avogadro::Core::Cube* cube() const { return m_cube; }
```
perminder-17, line 72: Also, I tried it with a different method where I was adding cube's data from Mainwindow.cpp file here https://github.com/OpenChemistry/avogadroapp/blob/7c2e387a5fbc0224806e37b62cf95484a8f26c6d/avogadro/mainwindow.cpp#L1869

as soon as I changes the background color's I was getting the cube's data from mainwindow, but still that doesn't works, I was getting garbage values, like I was getting negative dimensions, so I just want to access the correct cube's data which I am unable to do right now?

perminder-17:
Completing this work in #1979

PR #1979, opened 2025-03-14
Volumetric-rendering in avogadro project. 

perminder-17:
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.


perminder-17:
## I tried fixing the looping order of the cube's data, still some work is left but here's the result. Let me know how it looks @matterhorn103 @ghutchis @cryos 

# Water

[Screencast from 16-03-25 10:16:23 PM IST.webm](https://github.com/user-attachments/assets/644682ee-030d-4925-9f49-38d1f64d6bf9)

# Ammonia
[Screencast from 16-03-25 10:18:28 PM IST.webm](https://github.com/user-attachments/assets/a4e0a85c-e51e-486f-919d-ae033b980f33)

# crh206

Screenshot from 2025-03-16 22-21-45

ghutchis:
Here's the compile issue.
[cube.patch](https://github.com/user-attachments/files/19291206/cube.patch)

ghutchis:
I'm running into some crashes on Mac with the VAO in the volume render. I think it's Mac-specific, so I'll do some reading / debugging.

```
  // Box geometry for front/back pass
  glGenVertexArrays(1, &d->volumeBoxVao);
  glBindVertexArray(d->volumeBoxVao);
```

The first line crashes. I'll investigate further.

perminder-17:
No problem geoff, I can also help. But, can you undo all the changes at cube.cpp file. (Like replacing cube.cpp file from actual main branch of cube.cpp file.) This will change the looping order and will crash some of the molecules, but you could still work on small molecules. Since the compile issue is with cube.cpp file.

ghutchis:
I mean, you can grab the `cube.cpp` file yourself: https://raw.githubusercontent.com/OpenChemistry/avogadrolibs/refs/heads/master/avogadro/core/cube.cpp

Or you can `git checkout master` then copy the `cube.cpp` file to another directory, switch back to your branch and move the `cube.cpp` file, etc.

(There are other ways to do this in git, but these are easy.)

perminder-17:
[Screencast from 23-03-25 04:26:00 AM IST.webm](https://github.com/user-attachments/assets/3829006c-5441-49f4-a476-ff9cf9fba441)

This is the colorMapping I am trying @ghutchis

perminder-17:
[Screencast from 23-03-25 05:20:25 AM IST.webm](https://github.com/user-attachments/assets/3408828e-ce06-4e38-9725-6990f7f95a4a)


Here's the structure for caffeine from caffeine.cube file.

avogadro/rendering/volume_fs.glsl
```diff
@@ -0,0 +1,88 @@
+#version 330 core
```
perminder-17, line 1: Oops, I forgot to change the version. I'll fix this very soon, but @ghutchis do you think this could be the reason of compilation issue in macOS? Also, if not, I think we could include some more libraries as we did in other files 
`#include <GL/glew.h>` or maybe to fix this issue?

perminder-17:
Btw, @ghutchis I have lower the version in shaders. Can you take a look and also, I have link the colors (here's the result)

[Screencast from 23-03-25 09:44:34 AM IST.webm](https://github.com/user-attachments/assets/4d3dea6f-9278-470b-b695-a2b439c2d13a)

ghutchis:
I don't think it's needed to go back to an old shader version. The problem on the Mac isn't the shader version, it's that `glGenVertexArrays`, etc. are only available in 3.x or newer and Apple only provides 2.x in compatibility mode or 4.x in core profile.

So I'll set this up to only build for Windows and Linux right now, and push to update the remaining bits of code to use core profile.

avogadro/qtplugins/CMakeLists.txt
ghutchis, line 1: You are using some really old bits of code. Please replace this:
https://gist.github.com/ghutchis/41ab922e800a7c2f14a2cb64ad65cd20

avogadro/rendering/volume_box_fs.glsl
```diff
@@ -0,0 +1,9 @@
+#version 120
```
ghutchis, line 1: I'd probably leave all these shaders as 330 core

ghutchis (changes requested):
Just a minor thing - it would be great to fix the qtplugins/CMakeLists.txt - you're using an old version for some reason...

ghutchis:
I'm trying on an Ubuntu VM - it seems when I resize the window, the volumetric cloud gets misaligned from the molecule:

Screenshot From 2025-03-25 16-33-22

avogadro/rendering/volumegeometry.cpp
```diff
@@ -5,14 +5,744 @@
...
+  d->program->setUniformValue("negativeColor", negCol);
+
+  glBindVertexArray(d->vao);
+  glDrawArrays(GL_TRIANGLES, 0, 6);
+  glBindVertexArray(0);
+
+  d->program->release();
+}
+
+void VolumeGeometry::resize(int width, int height)
```
perminder-17, line 647: If I could get the correct width and height of the resized canvas, that will be fixed. Any tips? Where I should write this to get the correct size. 

It's only because framebuffers is at 1000X900 fixed. I want to resise the framebuffers correctly. Any tips?

perminder-17: Like I am asking in which file I should have volume->resize(actualwidth, actualHeight) so It could directly hit here and have the correct size?

perminder-17:
[Screencast from 28-03-25 06:20:00 AM IST.webm](https://github.com/user-attachments/assets/4eaa64ae-7613-4edc-8789-ce6d57df512d)

I think we were already having the correct width and height in the volumeGeometry class :"). I was thinking very complex logic to find it out but had an oversight that we already have them.

perminder-17:
Hi @ghutchis , do you mind checking it once again then I could probably work on code cleanups.

matterhorn103:
> I'm trying on an Ubuntu VM - it seems when I resize the window, the volumetric cloud gets misaligned from the molecule:
> 
> Screenshot From 2025-03-25 16-33-22

This is actually an existing problem and applies to all displayed orbitals, it's not isolated to the volumetric renderer. I noticed it a little while ago and just hadn't got round to making an issue for it yet.

(I figured there's enough more urgent issues that I didn't need to rush with the bug report.)

ghutchis:
Okay, I merged this into #2559 which is rebased on current `HEAD` and some minor updates to the shaders.

Once 2.0 is released on 06-Feb, I will merge this 🎉 .. took me too long to get OpenGL 4.0 working on the mac, but this was the incentive. 😄

PR #2065, opened 2025-06-24
style: run clang-format

e-kwsm:
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.


avogadro/core/array.h
```diff
@@ -43,15 +43,18 @@ class ArrayRefContainer
   explicit ArrayRefContainer(const size_t n,
                              const ValueType& value = ValueType())
     : m_ref(1), data(n, value)
-  {}
```
ghutchis, line 46: I definitely don't like this change. It's adding a lot of un-needed linebreaks.

`AllowShortBlocksOnASingleLine: Empty`

avogadro/qtplugins/importpqr/pqrrequest.h
```diff
@@ -107,7 +107,8 @@ private slots:
     result()
       : inchikey("Error"), name("Error"), mol2url("Error"), formula("Error"),
         mass(-1.0)
-    {}
+    {
```
ghutchis, line 110: These kind of changes seem really weird considering a bunch of other short blocks are consolidated.

If we can get rid of these, I'd appreciate it.

ghutchis:
I'll have to look more carefully. Somewhere the `#include` reordering has broken `rendering/avogadrogl.h`

```
// GLEW has to come before gl.h
#include <GL/glew.h>
 
#ifdef __APPLE__
#include <OpenGL/gl.h>
#else
#include <GL/gl.h>
#endif
```

ghutchis:
Otherwise, most of this looks good.

avogadro/core/array.h
e-kwsm: Ugh, I tried Empty but it does not work as expected.

I am using clang-format 20.1.6.

ghutchis: I guess we want `AllowShortBlocksOnASingleLine: Always` 
- https://github.com/llvm/llvm-project/issues/62117

ghutchis:
Okay, the `gl.h` issue is fortunately only in `overlayaxes.cpp`:

```
// clang-format off
// This header must come before anything else involving OpenGL
#include <avogadro/rendering/avogadrogl.h>

#include <avogadro/qtopengl/glwidget.h>
// clang-format on
```

avogadro/qtplugins/overlayaxes/overlayaxes.cpp
```diff
@@ -5,19 +5,19 @@
 
 #include "overlayaxes.h"
 
+#include <avogadro/qtopengl/glwidget.h>
```
ghutchis: This is the cause of the compile failure. We just need to move this down after `avogadrogl.h` somewhere and make sure clang-format doesn't swap them again.

e-kwsm: Thank you for your check. I reverted the sort.

Issue #1701, opened 2024-08-31
Generated CP2K input is invalid


e-kwsm:
**Avogadro version: (please complete the following information from the About box):**
 - Avogadrolibs: 1.99.0
 - Qt: 5.15.2

**Desktop version: (please complete the following information):**
 - OS: EndeavourOS
 - Version: 
 - Compiler (appimage)

**Describe the bug**
Generated CP2K input is malformed. See [ja.txt](https://github.com/user-attachments/files/16826114/ja.txt) (Japanese locale).

Regarding Energy calculation:

- GLOBAL
  - [ ] [PROJECT](https://manual.cp2k.org/cp2k-2024_2-branch/CP2K_INPUT/GLOBAL.html#CP2K_INPUT.GLOBAL.PROJECT_NAME)  
    `2:  PROJECT CO | ????? | SZV-GTH`
    - by default avogadro determines the `PROJECT` by the formula, calculation type, and basisset
    - here the calculation type is "エネルギー" (Energy in ja), correctly shown in the GUI, but saved wrongly
    - CP2K uses `PROJECT` for the prefix of intermediate files; anyway, i.e. in any locale, CP2K 2024.2 rejects the line since it contains spaces (and `|` is not suitable for filename); it must be e.g. `PROJECT CO_Energy_SVZ-GTH`
- FORCE_EVAL
  - [ ] [METHOD](https://manual.cp2k.org/cp2k-2024_2-branch/CP2K_INPUT/FORCE_EVAL.html#CP2K_INPUT.FORCE_EVAL.METHOD)  
    `8:  METHOD DFT`
    `DFT` must be `QS` or `QUICKSTEP` for electronic structure methods including DFT
  - [ ] `9:&END FORCE_EVAL` this must be put at the EOF but `58: $END` appears instead
  - [ ] SUBSYS: surrounding `&SUBSYS` and `&END SUBSYS` are missing
    - [ ] [KIND](https://manual.cp2k.org/cp2k-2024_2-branch/CP2K_INPUT/FORCE_EVAL/SUBSYS/KIND.html): closing `&END KIND` is gone
    - [ ] [COORD](https://manual.cp2k.org/cp2k-2024_2-branch/CP2K_INPUT/FORCE_EVAL/SUBSYS/COORD.html): enclosing `&COORD` and `&END COORD` are missed around the C and O coordinates
  - DFT
[...]

PR #2067, opened 2025-06-25
improve cp2kinput

e-kwsm:
#1701

Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

### DIFF
--- a/avogadro/qtplugins/cp2kinput/cp2kinputdialog.cpp
+++ b/avogadro/qtplugins/cp2kinput/cp2kinputdialog.cpp
@@ -444,7 +444,7 @@ void Cp2kInputDialog::updatePreviewText()
   std::map<QString, int> valencee;
 
   valencee["H"] = 1;
-  valencee["He"] = 2;
+  valencee["He"] = 2; // Z=2
   valencee["Li"] = 3;
   valencee["Be"] = 4;
   valencee["B"] = 3;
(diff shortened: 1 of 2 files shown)


ghutchis:
I don't know why there's one build failure - seems to be based on finding Eigen, so I'm going to ignore it for now.

avogadro/qtplugins/cp2kinput/cp2kinputdialog.cpp
```diff
@@ -230,6 +230,22 @@ void Cp2kInputDialog::restoreOptionCache()
   }
 }
 
+QString Cp2kInputDialog::toenum(CalculateOption option)
```
ghutchis: Maybe `calcKeyword` or something like that? or `fromCalcEnum`?

avogadro/qtplugins/cp2kinput/cp2kinputdialog.h
```diff
@@ -102,6 +102,8 @@ private slots:
 
   void setBasicDefaults();
 
+  /// @return valid values for CP2K RUN_TYPE
+  static QString toenum(CalculateOption option);
```
ghutchis: Does this need to be `static`? You're only using it inside the class, right? I don't think it matters either way.

ghutchis (commented):
Just a few minor thoughts to make the `toenum` method clearer.

avogadro/qtplugins/cp2kinput/cp2kinputdialog.cpp
e-kwsm: renamed to fromCalcEnum.

avogadro/qtplugins/cp2kinput/cp2kinputdialog.h
e-kwsm: The method does no depend on member variables but on private enum, fromCalcEnum, so it can be specified as static.

PR #2285, opened 2025-10-16
Fix/issue 509 add molecule button

Daniel-Aguila:
Implements the library-side changes for [OpenChemistry/avogadroapp#509](https://github.com/OpenChemistry/avogadroapp/issues/509).
This update introduces a “+” placeholder row to the Files → Molecules list, allowing users to add a new molecule directly from the UI.

### DIFF
--- a/avogadro/qtgui/moleculemodel.cpp
+++ b/avogadro/qtgui/moleculemodel.cpp
@@ -32,7 +32,7 @@ void MoleculeModel::loadIcons(bool darkMode)
   QString iconPath = ":icons/fallback/32x32/";
   QString plusIconPath = iconPath + (darkMode ? "plus-dark.png" : "plus-light.png");
 
-  m_plusIcon = QIcon::fromTheme("file-add", QIcon(plusIconPath));
+  m_plusIcon = QIcon(plusIconPath);
 }
 
 int MoleculeModel::rowCount(const QModelIndex& p) const


avogadro/qtgui/moleculemodel.cpp
```diff
@@ -16,19 +16,31 @@ namespace Avogadro::QtGui {
...
 {
   return QModelIndex();
 }
 
+void MoleculeModel::loadIcons(bool darkMode)
+{
+  QString iconPath = ":icons/fallback/32x32/";
+  QString plusIconPath = iconPath + (darkMode ? "plus-dark.png" : "plus-light.png");
+
+  m_plusIcon = QIcon::fromTheme("file-add", QIcon(plusIconPath));
```
ghutchis: I don't think "file-add" is a thing. Do you want "document-new" instead?
It would be okay to just go with the plusIconPath, but "file-add" seems like it's always going to fail.

ghutchis (commented):
I'll need to squash on the merge. I don't want to add a bunch of build files to the git history.

avogadro/qtgui/moleculemodel.cpp
Daniel-Aguila: Sounds good, I think just leaving it with plusIconPath is better then so it stays consistent.

Daniel-Aguila:
Apologies about the squashing. I thought I had ignored the build files but they went through. I'll try my best to verify and have a clean pr further down

ghutchis:
IMHO the easiest is to have the build in a completely different directory path. In this case, it's easy enough for me to squash when merging.

Daniel-Aguila (commented):
requested changes addressed - changed it to just have the plus sign
Yeah that sounds perfect, I'll build them outside for sure

Issue #2228, opened 2025-09-30
dependency on JKQTPlotter


e-kwsm:
**Is your feature request related to a problem? Please describe.**

Now avogadrolibs requires JKQTPlotter
https://github.com/OpenChemistry/avogadrolibs/blob/cc972e3d935c9f1888d7bc4cec3c7e291a5a31bc/avogadro/qtgui/CMakeLists.txt#L34
but the dependency is not automatically resolved.

**Describe the solution you'd like**

This may be redundant, but add JKQTPlotter to an external project and build it when not found, like OpenChemistry/openchemistry.

**Describe alternatives you've considered**

Add an option, say `USE_JKQTPLOTTER`, to enable (default) the feature.

**Additional context**

ghutchis:
This should be solved with HEAD in `openchemistry` - but you can disable the plotter with USE_PLOTTER.

ghutchis:
Is this still a problem for you? It's been building cleanly on GitHub and for other users.

e-kwsm:
Yes, openchemistry automatically build JKQTPlotter, but I want to build only this repository to implement and test features. I can manually set the path to openchemistry&rsquo;s JKQTPlotter but it is not handy.

ghutchis:
But you should be able to use `-DUSE_PLOTTER=OFF` and it will ignore the plot features, no?

e-kwsm:
I thought about adding `USE_JKQTPLOTTER` for finer control but it does not seem to be useful.

PR #2357, opened 2025-10-29
AvogadroLibsConfig.cmake: Find JKQTPlotter6

heirecka:
It's in the public link interface of Avogadro::QtGui, so it needs to be found as a dependency here.

### DIFF
--- a/cmake/AvogadroLibsConfig.cmake.in
+++ b/cmake/AvogadroLibsConfig.cmake.in
@@ -28,6 +28,7 @@ if (@QT_VERSION@ EQUAL 6)
   find_dependency(Qt6Gui)
   find_dependency(Qt6Network)
   find_dependency(Qt6Concurrent)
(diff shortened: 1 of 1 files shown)


ghutchis:
In principle, one can use `-DUSE_PLOTTER=OFF` to turn it off, otherwise it seems fine to me.

heirecka:
> In principle, one can use `-DUSE_PLOTTER=OFF` to turn it off, otherwise it seems fine to me.

That didn't occur to me from reading the cmake code and now that I grep for it, no such string seems to exist.

ghutchis:
I'll add in some patches. I appreciate the reminder though!

ghutchis:
I think it's better to go with this for now and I'll think about the best way to handle cases in which -DUSE_PLOTTER=OFF are used, e.g. #2228

PR #2384, opened 2025-11-05
Fixes to constraints diaglog

jcerezochem:
# Issues
The constraint dialogue does not work correctly (tried on Linux). Namely, the following problems are identified (the commit where this issue is solved in parentheses):

- The allowed range of the box to set the value of the coordinate is [0,104.9], which prevents the proper assignment of some angles and dihedrals (28997649)
- The type of coordinate to constraints is not updated with the selection (3544ed09)
- The atom ordering in the spin box differs from that in the constraint table (e8069b70)
- The maximum value for the atom index in the spin boxes is natoms-1, which prevents selecting the last atom (173acea1)
- When going from dihedral to angle to bonds, the extra indices remain active, and are passed, e.g. to the ORCA input (c3b82fbc)
- The Delete Selected button does not work (4445ac59)
- Atomic selections only set a new constraint if the dialogue is reopened. A button to generate a constraint from the selection when the dialog is open is needed (c8d804fa)
- The atomic order in the generated constraint does not generally follow the expected order according to connectivity, nor the chronological order of the selected atom (27a7e2c4)

# Notes
The last commit (27a7e2c4) attempts to introduce a minimal, self-consistent piece of code to provide the expected order based on connectivity (some options, such as out-of-plane, are not supported). But it'd be more useful to get it from the order the atoms were clicked.

# Developer Certificate of Origin
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

### DIFF
--- a/avogadro/qtplugins/constraints/constraintsdialog.cpp
+++ b/avogadro/qtplugins/constraints/constraintsdialog.cpp
@@ -131,6 +131,93 @@ void ConstraintsDialog::updateConstraints()
   else if (selectedAtoms.size() == 4)
     changeType(2);
 
+  // Each entry bonded[i] holds the atom indices bonded to selectedAtoms[i]
+  std::vector<std::vector<Index>> bonded(selectedAtoms.size());
+
+  for (Index i = 0; i < selectedAtoms.size(); ++i) {
+    Index ai = selectedAtoms[i];
+    for (Index j = i + 1; j < selectedAtoms.size(); ++j) {
+      Index aj = selectedAtoms[j];
+      if (m_molecule->bond(ai, aj).isValid()) {
+        bonded[i].push_back(j); // neighbor atom local index
+        bonded[j].push_back(i); // symmetric
+      }
+    }
+  }
+  std::vector<Index> degree(selectedAtoms.size(), 0);
+  for (Index i = 0; i < selectedAtoms.size(); ++i)
+    degree[i] = static_cast<Index>(bonded[i].size());
+
+  // Arrange selected atoms according to connectivity
+  if (selectedAtoms.size() == 3) {
+    Index i1, i2, i3;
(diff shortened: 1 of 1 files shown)


ghutchis:
Looks good overall. There's unfortunately no good way to get the atoms in the order they were selected, which is why the align and measure tools track things internally.

We ask everyone to sign their commits with a name and permanent e-mail so, for example, we can contact them about manuscripts, etc.
https://github.com/OpenChemistry/avogadrolibs/pull/2384/checks?check_run_id=54559529635

jcerezochem:
Thanks for the reply.  Sorry for the signature, I forgot to do it.

How should I proceed now? Should I fork the repo again and redo the commits with the signature? I can git rebase interactively and do this quickly, but it will rewrite the commits already used in this pull request, and mess things up.

ghutchis:
In this case I would rebase. Since it's not merged, it's okay you can use --force when pushing to the remote.

jcerezochem:
OK, done. Thanks

avogadro/qtplugins/constraints/constraintsdialog.ui
```diff
@@ -182,6 +182,25 @@
...
+           </sizepolicy>
+          </property>
+          <property name="minimumSize">
+           <size>
+            <width>100</width>
+            <height>0</height>
+           </size>
+          </property>
+          <property name="text">
+           <string>Get constraint</string>
```
ghutchis: This should be "Get Constraint"

jcerezochem: Should be fixed now

ghutchis:
Okay, assuming everything builds cleanly, I'll merge in a bit. Many thanks!

PR #2420, opened 2025-11-16
Adjust surface generation to keep winding consistent + and - mesh

### LINKED ISSUE #2417: Does anyone know what has changed with the rendering of MOs since earlier this year?
Apologies if this is not the correct place to ask this. Earlier this year I was using Avogadro2 (v1.98, 1.99, or 1.100 I believe) to render MOs from an ORCA TDDFT calculation, and the rendering looked quite nice for the inclusion in figures. I have since picked this up again for another project and have noticed that the rendering is not the same at all (have attached screenshots to show), and was wondering if anything has changed?

Image

Image

Are there any settings that can be changed in the app, or has the rendering pipeline been changed in subsequent versions? I have downloaded v1.99 and the rendering is the same as the new style, which in my opinion is far too dark.

ghutchis:
Should fix inconsistent lighting / opacity
h/t to @matterhorn103 and @callumhind2302

Fixes #2417

Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

### DIFF
--- a/avogadro/qtgui/meshgenerator.cpp
+++ b/avogadro/qtgui/meshgenerator.cpp
@@ -479,10 +479,16 @@ void MeshGenerator::FlyingEdgesAlgorithmPass4()
         // Add triangles
         const signed char* caseTri = m_caseTriangles[caseId]; // size 16
         for (int idx = 0; caseTri[idx] != -1; idx += 3) {
-
-          m_triangles[triIdx][0] = globalIdxs[caseTri[idx]];
-          m_triangles[triIdx][1] = globalIdxs[caseTri[idx + 1]];
-          m_triangles[triIdx][2] = globalIdxs[caseTri[idx + 2]];
+          if (!m_reverseWinding) {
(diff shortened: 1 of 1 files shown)


matterhorn103:
@ghutchis this has fixed the inconsistency, so both orbitals are lit from the same direction now, but unfortunately that direction is the wrong one (from below). 😕

Also, just to check you're aware that the sudden switch in lighting direction from 99% to 100% opacity is still there? I don't have the faintest idea when it comes to 3D rendering and lighting so that might just be a technical constraint or something but I assume it's not the desired behaviour.

ghutchis:
@matterhorn103 - there's definitely inconsistent lighting / shading between the transparent shader and the opaque shader, but I'm not sure I have enough GLSL expertise to fix that.

I dunno, maybe that's something that AI could help identify 🤷‍♂️

matterhorn103:
Ah I see, there's two different shaders in action, that makes a lot more sense then.

PR #2472, opened 2025-12-04
Use --as-is flag for pixi run to avoid environment updates

### LINKED ISSUE #2469: Plugins with pixi run should use --as-is
On each `pixi run` it will attempt to update the environment.

We should use `--as-is` to ignore updates, etc.
https://pixi.sh/dev/reference/cli/pixi/run/#update-options

harikrishna-au:
Fixes #2469

### Changes
This PR adds the `--as-is` flag to the `pixi run` command within `PythonScript::execute`.

### Reason
By default, `pixi run` checks for environment updates and lockfile consistency. Adding `--as-is` ensures that:
1.  The command runs using the existing environment without attempting any updates or installations.
2.  Execution is faster and avoids potential side effects from implicit environment modifications.

I have signed off on the commit as required.

### DIFF
--- a/avogadro/qtgui/pythonscript.cpp
+++ b/avogadro/qtgui/pythonscript.cpp
@@ -140,6 +140,8 @@ QByteArray PythonScript::execute(const QStringList& args,
       realArgs.prepend("python");
     } // otherwise hope pixi knows how to run this
 
+    realArgs.prepend("--as-is");
+
     // check if the script is in the plugin directory
     if (!pluginDir.isEmpty() && defaultManifest &&
         !m_scriptFilePath.startsWith(pluginDir)) {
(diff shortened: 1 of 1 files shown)


avogadro/qtgui/pythonscript.cpp
```diff
@@ -157,6 +157,7 @@ QByteArray PythonScript::execute(const QStringList& args,
     }
 
     realArgs.prepend("run");
+    realArgs.insert(1, "--as-is");
```
ghutchis: I'm pretty sure the `--as-is` needs to be after the "run" and before the task part:
https://pixi.sh/dev/reference/cli/pixi/run/#examples

ghutchis:
Just to check - was this AI-assisted?

harikrishna-au:
Yes, I used AI tools to help draft the initial changes, but I reviewed the logic myself. Thanks for catching the argument order issue—I've fixed it now so --as-is appears correctly after run.

ghutchis:
This one is fine, but our policy is not to accept AI contributions since the license is unclear. One line is not copyright-able, so this is fine.

If your other pull request is also AI-generated, I cannot accept it.

PR #2544, opened 2026-01-22
Make sure to include Qt6 icu.dll in Windows builds

### LINKED ISSUE #2495: Avogadro 2 has silently started to require Windows version >= 1903
Avogadro 2 no longer launches on Windows 10 systems older than version 1903 (e.g. LTSC 1809), because the icu.dll being used was introduced in Windows 1903. See https://learn.microsoft.com/en-us/windows/win32/intl/international-components-for-unicode--icu-#history-of-changes-to-the-icu-library-in-windows

A possible workaround would be building this library from https://github.com/microsoft/icu and bundling the resulting .dll, but failing that the documentation should be updated to reflect this new minimum system requirement of the Windows builds.

ghutchis:
Should fix #2495

Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

### DIFF
--- a/.github/workflows/build_windows.yml
+++ b/.github/workflows/build_windows.yml
@@ -127,6 +127,9 @@ jobs:
         mv pixi-x86_64-pc-windows-msvc.exe prefix/bin/pixi.exe
         # run windeployqt
         windeployqt --release prefix/bin/avogadro2.exe
+        # copy ICU DLLs from Qt installation (needed for Windows 10 1903+)
+        # windeployqt doesn't always copy these automatically
+        cp "${QT_ROOT_DIR}/bin"/icu*.dll prefix/bin/ 2>/dev/null || true
         mv prefix avogadro2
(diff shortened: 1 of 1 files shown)


.github/workflows/build_windows.yml
```diff
@@ -127,6 +127,9 @@ jobs:
         mv pixi-x86_64-pc-windows-msvc.exe prefix/bin/pixi.exe
         # run windeployqt
         windeployqt --release prefix/bin/avogadro2.exe
+        # copy ICU DLLs from Qt installation (needed for Windows 10 1903+)
```
TiborGY, line 130: Nit: this is for older versions of Win 10, so <1903 not 1903+

ghutchis: Thanks - I'll fix that up when I have a chance to log into a Windows build and see why the icu.dll wasn't included.

PR #2551, opened 2026-01-25
Some packaging fixes

ghutchis:
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

### DIFF
--- a/.github/workflows/build_windows.yml
+++ b/.github/workflows/build_windows.yml
@@ -117,10 +117,6 @@ jobs:
       shell: bash
       working-directory: ${{ runner.workspace }}/build
 
-    - name: Check build with tmate
-      uses: mxschmitt/action-tmate@c0afd6f790e3a5564914980036ebf83216678101 # v3.23
-      timeout-minutes: 60
-
     - name: Create Windows Packages
       if: runner.os == 'Windows'
       shell: bash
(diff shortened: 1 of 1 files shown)


ghutchis:
@TiborGY - looks like this one is including icu*.dll so hopefully it fixes #2495

TiborGY:
> @TiborGY - looks like this one is including icu*.dll so hopefully it fixes #2495

I think some new icu*.dll did get included but there is no `icu.dll`. Since that does not exist, A2 still cannot run.
I am getting the feeling that this might be unresolvable without downgrading the Windows SDK, or patching QT.

TiborGY:
The prerequisites for the new SDK to build compatible binaries seem to be:

1. Use the legacy import libraries, icuuc.lib and icuin.lib, not the combined library icu.lib
2. Do not use any new APIs not present in the legacy import libraries.
3. Ensure icuuc.lib and icuin.lib are listed before the umbrella libraries, like onecoreuap.lib or WindowsApp.lib, in the Additional Dependencies Linker setting of Visual Studio.
4. For Win32 apps to use ICU, they need to call [CoInitializeEx](https://learn.microsoft.com/en-us/windows/win32/api/combaseapi/nf-combaseapi-coinitializeex) first. This may or may not be be neglected by QT6, since on Windows 10 version 1903 and above, where the combined ICU library (icu.dll/icu.lib) is available, you can omit the CoInitializeEx call by using the combined library. If QT6 dropped support for older versions, of Win10 they might have stripped this call out from their codebase,

ghutchis:
Looks like it may be specific to Qt 6.10:
https://www.patreon.com/posts/new-requirement-148823382

Want me to try a build with 6.8.x or 6.9.x?

ghutchis:
Maybe it's better to go with 6.8.x since that's the last LTS. (I'll keep 6.10 for Mac because there are some specific fixes.)

Issue #1430, opened 2023-10-30
standalone tests file due to removal of Eigen configuration


drew-parsons:
**Avogadro version: (please complete the following information from the About box):**
 - Avogadrolibs: 1.98.0
 - Qt: 5.15.10+dfsg-4

**Desktop version: (please complete the following information):**
 - OS: Debian Linux
 - Version unstable
 - Compiler gcc 13.2.0-6

**Describe the bug**

PR #1267 removed Eigen configuration from CMakeLists.txt files throughout avogadrolibs.  This makes a problem when running the test standalone against an existing avogadrolibs installation (debian does that for [CI testing](https://ci.debian.net/packages/a/avogadrolibs/))

Running cmake against tests/CMakeLists.txt only and then building tests gets an error "Eigen/Dense: No such file or directory" from /usr/include/avogadro/core/vector.h:10.
This is because (on debian systems), Eigen is installed under /usr/include/eigen3.  So `EIGEN3_INCLUDE_DIR` is needed.

**To Reproduce**
1. create test dir under `tests`
```
mkdir tests/run_test 
cd tests/run_test
```
2. Configure tests, referring to installed avogadrolib headers
```
cmake -DAvogadroLibs_BINARY_DIR=/usr/include/ -DUSE_QT=ON -DUSE_OPENGL=ON -Wno-dev ..
``` 
3. `make AvogadroTests`

**Expected behavior**

Should be possible to build and run tests standalone for the purpose of CI testing of an existing avogadrolibs installation.

**Screenshots**
Error is
```
[  9%] Building CXX object core/CMakeFiles/AvogadroTests.dir/atomtest.o
cd /tmp/autopkgtest.Uz9Zw5/tree/tests/run_test/core && /usr/bin/c++  -I/usr/include/avogadro/core -DGTEST_HAS_PTHREAD=1 -MD -MT core/CMakeFiles/AvogadroTests.dir/atomtest.o -MF CMakeFiles/AvogadroTests.dir/atomtest.o.d -o CMakeFiles/AvogadroTests.dir/atomtest.o -c /tmp/autopkgtest.Uz9Zw5/tree/tests/core/atomtest.cpp
In file included from /usr/include/avogadro/core/atom.h:10,
                 from /usr/include/avogadro/core/bond.h:11,
                 from /usr/include/avogadro/core/molecule.h:14,
                 from /tmp/autopkgtest.Uz9Zw5/tree/tests/core/atomtest.cpp:19:
[...]

PR #2575, opened 2026-01-29
Enable build of standalone tests

drew-parsons:
Applies debian patch tests_find_avogadrolibs.patch https://salsa.debian.org/debichem-team/avogadrolibs/-/blob/129fc8bba9006dee7b56c103c4a0537c36fc9e02/debian/patches/tests_find_avogadrolibs.patch

Allows tests to be built and run as standalone tests using an existing avogadrolibs installation, via the AvogadroLibs cmake config files.

Useful for CI testing or confirming correct installation of the library.

Assumes Qt 6 has been used to build avogadrolibs.

cf. discussion at https://github.com/OpenChemistry/avogadrolibs/issues/1430#issuecomment-3820765654

Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

### DIFF
--- a/tests/CMakeLists.txt
+++ b/tests/CMakeLists.txt
@@ -50,7 +50,14 @@ find_path(AVOGADRO_DATA_ROOT .avogadro.data
 mark_as_advanced(AVOGADRO_DATA_ROOT)
 
 # Add the tests for each module.
-add_subdirectory(calc)
+
+if(NOT PROJECT_IS_TOP_LEVEL)
+  # calc/energycalculator.h requires cppoptlib/problem.h, which is
+  # provided in source but not currently installed with AvogadroLibs.
+  # So skip calc in standalone tests until Issue#2631 is resolved.
+  add_subdirectory(calc)
+endif()
+
 add_subdirectory(core)
 add_subdirectory(io)
 add_subdirectory(quantumio)


ghutchis:
Thanks, this looks good, but I'll need the DCO sign-off. `git rebase HEAD~1 --signoff`

https://github.com/OpenChemistry/avogadrolibs/pull/2575/checks?check_run_id=61938311225

ghutchis:
It's still failing with:
```
  By not providing "FindAvogadroLibs.cmake" in CMAKE_MODULE_PATH this project
  has asked CMake to find a package configuration file provided by
  "AvogadroLibs", but CMake did not find one.

  Could not find a package configuration file provided by "AvogadroLibs" with
  any of the following names:

    AvogadroLibsConfig.cmake
    avogadrolibs-config.cmake

  Add the installation prefix of "AvogadroLibs" to CMAKE_PREFIX_PATH or set
  "AvogadroLibs_DIR" to a directory containing one of the above files.  If
  "AvogadroLibs" provides a separate development package or SDK, be sure it
  has been installed.
```

drew-parsons:
The standalone code path is intended to be run when an AvogadroLibs installation is in place, i.e. when the cmake config files are accessible. 

It should be ignored in a normal build where the tests are built as part of processing the toplevel CMakeLists.txt.  The patch identified this state via definition of the AvogadroCore target. If AvogadroCore is being processed, it would mean a "normal" toplevel build.

I'll review the test log and see if I can check if the test context is making sense.

drew-parsons:
The test logs accessible to me in the Actions runs (https://github.com/OpenChemistry/avogadrolibs/actions/runs/21502948435/job/61955501451?pr=2575 etc) seem to be passing in the sense that this patch is not blocking builds (the tests themselves don't seem to be getting built or run). There are tmate errors in the Sanitizer checks, but that seems to be unrelated to theispatch.

These logs are not showing the AvogadroLibsConfig.cmake error.  Are there other test logs I should be inspecting?

ghutchis:
`tmate` just allows you to authenticate to the build server if you want to inspect the failure further.

The builds with tests (i.e., the sanitizers) are all failing:
```
CMake Error at tests/CMakeLists.txt:8 (find_package):
  By not providing "FindAvogadroLibs.cmake" in CMAKE_MODULE_PATH this project
  has asked CMake to find a package configuration file provided by
  "AvogadroLibs", but CMake did not find one.

  Could not find a package configuration file provided by "AvogadroLibs" with
  any of the following names:

    AvogadroLibsConfig.cmake
    avogadrolibs-config.cmake

  Add the installation prefix of "AvogadroLibs" to CMAKE_PREFIX_PATH or set
  "AvogadroLibs_DIR" to a directory containing one of the above files.  If
  "AvogadroLibs" provides a separate development package or SDK, be sure it
  has been installed.
```

https://productionresultssa12.blob.core.windows.net/actions-results/66157259-4f03-46f5-9bf3-d06ff25e8700/workflow-job-run-6599563e-af61-56e4-b96e-62ca0a5f3c1d/logs/job/job-logs.txt?rsct=text%2Fplain&se=2026-01-30T17%3A15%3A47Z&sig=fwyS6pC5FY3sXDkoQd6vJBD%2BA5Szx%2F1frE1aLwLPtVg%3D&ske=2026-01-30T18%3A52%3A02Z&skoid=ca7593d4-ee42-46cd-af88-8b886a2f84eb&sks=b&skt=2026-01-30T14%3A52%3A02Z&sktid=398a6654-997b-47e9-b12b-9515b896b4de&skv=2025-11-05&sp=r&spr=https&sr=b&st=2026-01-30T17%3A05%3A42Z&sv=2025-11-05

ghutchis:
And let me be clear, I won't merge a patch that fails the test runners. We want to run the unit tests on all our builds.

If you want to have some other `ENABLE_STANDALONE_TESTS=ON` that would be fine with me.

drew-parsons:
It's not authenticating me to the windows.net blob either,
`Server failed to authenticate the request. Make sure the value of Authorization header is formed correctly including the signature. RequestId:002f5fa7-701e-0073-6811-92cf62000000 Time:2026-01-30T17:54:46.6959249`

ghutchis:
Okay, I'm just clicking through to a run and then using the "View Raw Logs", e.g.
https://github.com/OpenChemistry/avogadrolibs/actions/runs/21502948436/job/61955501408?pr=2575

I can promise that for all the test builds, they're all failing with the same error:
```
CMake Error at tests/CMakeLists.txt:8 (find_package):
  By not providing "FindAvogadroLibs.cmake" in CMAKE_MODULE_PATH this project
  has asked CMake to find a package configuration file provided by
  "AvogadroLibs", but CMake did not find one.

  Could not find a package configuration file provided by "AvogadroLibs" with
  any of the following names:

    AvogadroLibsConfig.cmake
    avogadrolibs-config.cmake

  Add the installation prefix of "AvogadroLibs" to CMAKE_PREFIX_PATH or set
  "AvogadroLibs_DIR" to a directory containing one of the above files.  If
  "AvogadroLibs" provides a separate development package or SDK, be sure it
  has been installed.
```

drew-parsons:
Raw logs are working for me. Accessible beneath the page-setup gear icon. I'll check the rest of the raw log for context.

tests/CMakeLists.txt
```diff
@@ -1,3 +1,13 @@
+if(NOT TARGET AvogadroCore)
+  # Enable build of standalone tests.
+
+  # cmake and QT_VERSION not previously defined when building tests only
+  cmake_minimum_required(VERSION 3.24 FATAL_ERROR)
+  set(QT_VERSION 6)
```
e-kwsm: > It should be ignored in a normal build where the tests are built as part of processing the toplevel CMakeLists.txt. The patch identified this state via definition of the AvogadroCore target. If AvogadroCore is being processed, it would mean a "normal" toplevel build.

If you want to check if this is top level or not, you can use [PROJECT_IS_TOP_LEVEL](https://cmake.org/cmake/help/v3.24/variable/PROJECT_IS_TOP_LEVEL.html):

```suggestion
cmake_minimum_required(VERSION 3.24 FATAL_ERROR)
project(AvogadroLibsTest LANGUAGES CXX)
if(PROJECT_IS_TOP_LEVEL)
  # Enable build of standalone tests.

  # cmake and QT_VERSION not previously defined when building tests only
  set(QT_VERSION 6)
```

tests/CMakeLists.txt
```diff
@@ -1,3 +1,13 @@
...
+
+  find_package(AvogadroLibs REQUIRED)
```
e-kwsm, line 14: This fails unless AvogadroLibs is installed on the system.

tests/CMakeLists.txt
drew-parsons: Thanks for the review.  That's a good hint. `PROJECT_IS_TOP_LEVEL` enables the intended behaviour well

tests/CMakeLists.txt
drew-parsons: That's right.  Essentially that is the point.  The usage of this "PROJECT_IS_TOP_LEVEL" section in tests/CMakeLists.txt is intended precisely for making it possible to run the tests against AvogadroLibs installed on the system. This block would not be active in the context of a normal build where tests/CMakeLists.txt is not top-level.  And it wouldn't make sense to run the tests standalone without AvogadroLibs being installed on the system.

drew-parsons:
> If AVOGADRO_DATA_ROOT is set, build fails since cppoptlib/problem.h is missed (#2631).

Interesting problem. This is not a new code issue, curious that it hadn't been a problem before, for instance when compiling the avogadro app which uses avogadrolibs. I guess it was fine because the app uses libAvogadroCalc.so without needing to use calc/energycalculator.h directly.  

And I didn't catch it in the debian standalone tests because debian/tests was configured to only build the core, qtgui and rendering tests, so calc tests weren't being run (besides, debian is still on 1.101.0, while these calc tests were added in 1.102.0).

ghutchis:
I added new tests for `calc` because they absolutely need unit tests (and they caught a few gradient bugs). I'll address the header install issue.

drew-parsons:
> If AVOGADRO_DATA_ROOT is set, build fails since cppoptlib/problem.h is missed (#2631).

I think there are two ways to manage it.

The simplest action is to make `add_subdirectory(calc)` conditional on not having `PROJECT_IS_TOP_LEVEL`, i.e. just skip the calc tests in these standalone tests.

Alternatively a toplevel configuration flag could be added to install cppoptlib alongside the avogadrolibs headers, so cppoptlib/problem.h  becomes available and can be used to build the calc tests.

Do you have a preference for one or the other of these alternatives?  (or allowing for both at the same time could be a third option).

ghutchis:
I think for now, I'd probably suggest the conditional run of the `calc` tests as long as that's okay with Debian protocol.

I think @e-kwsm already filed #2631 so we'll make sure the headers are installed too. (And at some point I need to migrate to the new `cppoptlib` headers.)

drew-parsons:
> I think for now, I'd probably suggest the conditional run of the `calc` tests as long as that's okay with Debian protocol.

The calc test will still be run as normal at build time so ~~not too much~~ nothing is lost by skipping it in the standalone tests.
It will be a bonus extra test once the cppoptlib header installation is in place.

e-kwsm (approved):

ghutchis:
The builds are failing due to changes in avogadroapp. Looks good, thanks!

Issue #2147, opened 2025-08-17
undo does not work for copied Array?


e-kwsm:
**Avogadro version: (please complete the following information from the About box):**
 - Avogadrolibs: 1c5ebcb8bba27e301b0bb66acab90bcbb5ef2a45

**Describe the bug**

https://github.com/OpenChemistry/avogadrolibs/blob/1c5ebcb8bba27e301b0bb66acab90bcbb5ef2a45/tests/qtgui/rwmoleculetest.cpp#L579

Here *three* arguments are passed to std::equal, which is something like

```cpp
auto equal = [](auto first1, auto last1, auto first2) -> bool {
  for (; first1 != last1; ++first1, ++first2)
    if (*first1 != *first2)
      return false;
  return true;
};
```

<https://en.cppreference.com/w/cpp/algorithm/equal.html>

So, it evaluates to *true* when C1 ⊆ C2, i.e. it is not symmetric. Four-parameter version, which takes last2, checks `first2 != last2` in the loop.

Actually, the followings fail.

```cpp
  EXPECT_TRUE(std::equal(mol.bondOrders().begin(), mol.bondOrders().end(), ords.begin()));
  EXPECT_TRUE(std::equal(ords.begin(), ords.end(), mol.bondOrders().begin(),
                         mol.bondOrders().end()));
```

And

```cpp
  EXPECT_EQ(ords, mol.bondOrders());
```

shows

```
Expected equality of these values:
  ords
    Which is: {}
  mol.bondOrders()
    Which is: { '\x1' (1), '\x2' (2), '\x3' (3) }
```

Array::operator== checks the sizes of the containers.

https://github.com/OpenChemistry/avogadrolibs/blob/1c5ebcb8bba27e301b0bb66acab90bcbb5ef2a45/avogadro/core/array.h#L382-L387

`ords` is cleared but not undone.

https://github.com/OpenChemistry/avogadrolibs/blob/1c5ebcb8bba27e301b0bb66acab90bcbb5ef2a45/tests/qtgui/rwmoleculetest.cpp#L573-L576

PR #2642, opened 2026-02-24
refactor: use four-parameter version of std::equal to check size consistency

e-kwsm:
c.f. #2147

Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

### DIFF
--- a/tests/qtgui/rwmoleculetest.cpp
+++ b/tests/qtgui/rwmoleculetest.cpp
@@ -321,11 +321,11 @@ TEST(RWMoleculeTest, setAtomicNumbers)
 
   mol.setAtomicNumbers(rep);
   ASSERT_TRUE(std::equal(mol.atomicNumbers().begin(), mol.atomicNumbers().end(),
-                         rep.begin()));
+                         rep.begin(), rep.end()));
 
   mol.undoStack().undo();
   EXPECT_TRUE(std::equal(mol.atomicNumbers().begin(), mol.atomicNumbers().end(),
-                         orig.begin()));
+                         orig.begin(), orig.end()));
 }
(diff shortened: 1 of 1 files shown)


tests/qtgui/rwmoleculetest.cpp
```diff
@@ -575,7 +577,8 @@ TEST(RWMoleculeTest, clearBonds)
 
   mol.undoStack().undo();
   EXPECT_EQ(3, mol.bondCount());
-  EXPECT_TRUE(std::equal(pairs.begin(), pairs.end(), mol.bondPairs().begin()));
+  EXPECT_TRUE(std::equal(pairs.begin(), pairs.end(), mol.bondPairs().begin(),
+                         mol.bondPairs().end()));
   EXPECT_TRUE(std::equal(ords.begin(), ords.end(), mol.bondOrders().begin()));
```
e-kwsm: This line is not modified because test fails: `ords.size()` is zero whereas `mol.bondOrders().size()` is three.
Three-parameter version evaluates to true vacuously (#2147).

ghutchis: I'll see if I can figure out why `ords.size()` is zero so we can fix this before merging

ghutchis:
Okay - #2645 should address this. I'd suggest that line should not only use the 4-parameter version, but we add some explicit checks for the sizes of the arrays.

Assuming #2645 builds, I'll merge this in and these tests should all pass.

Many thanks - that was hard to spot.

ghutchis:
Thanks!

PR #2646, opened 2026-02-25
Updated "manage plugins" interface

ghutchis:
Still plenty of bugs, but mainly aiming to get UI feedback

Developer Certificate of Origin
Version 1.1

Copyright (C) 2004, 2006 The Linux Foundation and its contributors.
1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

Everyone is permitted to copy and distribute verbatim copies of this
license document, but changing it is not allowed.

Developer's Certificate of Origin 1.1

By making a contribution to this project, I certify that:

(a) The contribution was created in whole or in part by me and I
    have the right to submit it under the open source license
    indicated in the file; or

(b) The contribution is based upon previous work that, to the best
    of my knowledge, is covered under an appropriate open source
    license and I have the right under that license to submit that
    work with modifications, whether created in whole or in part
    by me, under the same open source license (unless I am
    permitted to submit under a different license), as indicated
    in the file; or

(c) The contribution was provided directly to me by some other
    person who certified (a), (b) or (c) and I have not modified
    it.

(d) I understand and agree that this project and the contribution
    are public and that a record of the contribution (including all
    personal information I submit with it, including my sign-off) is
    maintained indefinitely and may be redistributed consistent with
    this project or the open source license(s) involved.

## Summary by CodeRabbit

* **New Features**
  * New "Package Installer" extension under the Extensions menu.
  * Browse, search and refresh an online package catalog from a package manager dialog.
  * View package details and README content inline.
  * Install, update, and remove packages with single actions; updates are auto-detected.
  * Install packages from local directories and from downloadable archives.

PR #2661, opened 2026-03-03
Fix plugin installation directory and change name to more clear `plugins`

brockdyer03:
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.


brockdyer03:
Needs https://github.com/OpenChemistry/avogadroapp/pull/756

brockdyer03:
To briefly explain the reason behind renaming the plugin install directory, I just found that in general it is more obvious to a user to look for a directory called "plugins" than "packages" since "packages" is

1. More broad than just an Avogadro plugin (it could be a Python package that is a dependency of a plugin)
2. Specifically a type of Python plugin
3. Not the name that is most commonly used on the forum, where most users will be looking
4. Generally more descriptive of what the plugins are.

Also, the fix for the plugin installation directory was noticed when I found that plugins were being extracted to a directory that looked nearly correct, e.g. `packagesmatterhorn103-avo_xtb-3fb27a5` but were obviously missing a forward slash between `packages` and the plugin name.

ghutchis:
I think we should discuss on the forum. Also, keep in mind that the installer isn't merged, so if there's a bug in the extraction path, please raise that as a comment on that pull request: #2646

ghutchis:
Thanks, I'm going to merge this and the plugin manager so we can continue to test and iterate.

PR #2675, opened 2026-03-06
Carefully remove symlink packages (e.g., local installs)

### LINKED ISSUE #2674: File deletion when plugins are symlinked from a local directory
## Problem

This is with the AppImage built in https://github.com/OpenChemistry/avogadroapp/pull/760

The problem arises when the following chain of events occurs:

- A local directory is installed as a plugin from the plugin manager using **Install from Directory...**, creating a symlink in the plugins folder to the source directory
- The plugin is uninstalled via the plugin manager using **Remove Selected**
- "Keep files" is selected, leaving the symlink in the plugins folder
- The plugin is reinstalled in the plugin manager
- Avogadro can't create a new symlink, because a symlink to the same source, with the same name, already exists
- All the files in the source location get deleted!?
- The symlink remains in the plugins folder, but the contents are gone

(I guess what happens is that Avogadro tries to create the new symlink at `Avogadro/plugins/<plugin>`, that fails, so it instead copies the source into `Avogadro/plugins/<plugin>`, which is an attempt to copy it into itself, which just goes bang.)

The "Remove files" option does remove the symlink, but also all the files in the source directory.

## Proposed solution

Ideally I think Avogadro would, when asked to delete a local-only plugin:

- Check if the item in `Avogadro/plugins/` is actually a symlink rather than a real directory
- If it's a symlink, don't even offer the Remove Files/Keep Files choice
- If it's a symlink, unlink it, but don't delete it (as deleting the link = deleting the directory)

Essentially, use `unlink` for symlinks rather than `rm`.

ghutchis:
Fixes #2674

Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

### DIFF
--- a/avogadro/qtplugins/packageinstaller/packagemanagerdialog.cpp
+++ b/avogadro/qtplugins/packageinstaller/packagemanagerdialog.cpp
@@ -495,6 +495,20 @@ void PackageManagerDialog::removeSelected()
     }
   }
 
+  // Symlinks are created for local-disk installs — just remove the link
+  QMutableListIterator<int> it(toRemove);
+  while (it.hasNext()) {
+    int row = it.next();
+    const PackageModel::PackageEntry& e = m_model->entry(row);
+    QFileInfo fi(e.installedDir);
+    if (fi.isSymLink()) {
+      QFile::remove(e.installedDir);
+      const QString packageKey = e.packageKey.isEmpty() ? e.name : e.packageKey;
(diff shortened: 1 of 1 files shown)


ghutchis:
@matterhorn103 - this should do exactly what you propose - check for a symlink, if it exists, just remove that silently. Otherwise, ask the user about deleting files.

PR #2681, opened 2026-03-06
Create gltfvisitor.cpp

Thunder420Rock:
### Includes
```cpp
#include "gltfvisitor.h"
#include "ambientocclusionspheregeometry.h"
#include "cylindergeometry.h"
...
```
Each geometry header gives access to the actual rendered scene objects. This is the core reason for using the Visitor pattern — you're reading from the **live scene objects** (with all their user-set colors, sizes, and opacities), not from a static file format.

---

### Constructor & Destructor
```cpp
GltfVisitor::GltfVisitor(const Camera& c)
  : m_camera(c), m_opacity(1.0f)
```
Takes a reference to the scene `Camera`. The camera is needed if you later want to embed a viewpoint in the glTF (like `VRMLVisitor` does with its `Viewpoint` node). `m_opacity` defaults to 1.0 (fully opaque) and can be used later for transparency support.

---

### `begin()`
```cpp
m_spheres.clear();
m_cylinders.clear();
m_meshVertices.clear();
...
```
Resets all internal buffers so the visitor is safe to reuse across multiple export calls without leftover geometry from a previous run. Always call this before passing the visitor to the scene.

---

### `visit(SphereGeometry&)` — Atoms
```cpp
for (const auto& s : geometry.spheres()) {
  sd.center = s.center;
  sd.radius = s.radius;
  sd.color  = s.color;
```
Iterates over every sphere in the geometry object and stores its center position, radius, and **color as it appears in the rendered scene**. This is the critical improvement over `FileFormat` — if a user has changed an atom's color, size, or opacity in the Avogadro UI, `s.color` and `s.radius` already reflect those changes. A `FileFormat` reader would only ever see the raw molecular data, not the display customizations.

---

### `visit(AmbientOcclusionSphereGeometry&)` — AO Atoms Same logic as `visit(SphereGeometry&)`. AO spheres are a separate geometry class used when ambient occlusion shading is enabled. Treating them identically ensures atoms look correct regardless of which rendering mode the user has active.

---

### `visit(CylinderGeometry&)` — Bonds
```cpp
for (const auto& c : geometry.cylinders()) {
  cd.end1   = c.end1;
  cd.end2   = c.end2;
  cd.radius = c.radius;
  cd.color  = c.color;
```
Captures both endpoints, the radius, and the bond color. Bonds in Avogadro are often split into two half-cylinders of different colors (one per atom). Because you're iterating the geometry directly, both halves are captured automatically with their correct individual colors — something impossible to reconstruct from a `FileFormat` alone.

---

### `visit(MeshGeometry&)` — Orbitals & Surfaces
```cpp
const auto& verts = geometry.vertices();
for (const auto& v : verts) {
  m_meshVertices.push_back(v.vertex);
  m_meshColors.push_back(v.color);
}
```
This is the most important addition over the original `gltfformat` approach. `MeshGeometry` represents molecular orbitals, electron density surfaces, and other volumetric visualizations. The reviewer specifically mentioned *"even including meshes for orbitals"* — this `visit()` method is exactly what captures that. Each vertex carries both position and color (e.g. the +/− phase colors of an orbital lobe).

The index buffer is built in groups of 3 since all meshes are triangle lists:
```cpp
for (unsigned int i = 0; i < verts.size(); i += 3) {
  m_meshIndices.push_back(indexOffset + i);
  m_meshIndices.push_back(indexOffset + i + 1);
  m_meshIndices.push_back(indexOffset + i + 2);
}
```
`indexOffset` is added so that when multiple `MeshGeometry` objects are visited in sequence, their indices don't collide in the shared buffer.

---

### `visit(LineStripGeometry&)`
Left empty intentionally. Line strips are used for things like trajectory paths and unit cell wireframes. glTF supports line primitives (`mode: 1`) but that's a future enhancement — leaving the method as a no-op means it compiles cleanly and can be filled in later.

---

### `tessellateSphere()` — Icosphere Generation
```cpp
const float t = (1.0f + std::sqrt(5.0f)) / 2.0f; // golden ratio
```
glTF has no native "sphere" primitive — everything must be explicit triangle meshes. This function approximates each atom as a 20-face icosahedron (a shape based on the golden ratio that gives a good sphere approximation with few triangles). Each of the 12 vertices is placed at `center + direction * radius`, and 20 triangles connect them. For a higher-quality export you could subdivide further, but 20 faces is a good balance between file size and visual quality.

---

### `tesselateCylinder()` — Prism Generation
```cpp
Vector3f u = axis.cross(up).normalized();
Vector3f v = axis.cross(u);
```
Similarly, glTF has no cylinder primitive. This builds an orthonormal basis around the bond axis using cross products, then places 8 evenly-spaced vertices around each end cap. The 8 side quads are each split into 2 triangles, giving 16 triangles per cylinder. The `sides = 8` default is a parameter you can increase for smoother bonds.

---

### `buildJson()` — glTF 2.0 Assembly
This is where everything comes together into a valid glTF 2.0 JSON structure. The key components are:

**Accessors** describe the *type* of each data chunk:
- Accessor 0: `POSITION` — `VEC3` of `FLOAT` (x, y, z per vertex)
- Accessor 1: `COLOR_0` — `VEC3` of `FLOAT` (r, g, b per vertex, normalized 0–1)
- Accessor 2: `indices` — `SCALAR` of `UNSIGNED_INT`

**BufferViews** describe *where* in the binary buffer each accessor's data lives (byte offset + byte length).

**Buffer** is a single binary blob containing all three data arrays packed end-to-end: positions → colors → indices. The `.bin` file itself is written separately by the plugin that owns the file I/O — `buildJson()` only outputs the JSON side and records the expected `byteLength`.

**Mesh primitive mode `4`** means `TRIANGLES` in the glTF spec — every 3 indices form one triangle.

The result is a self-consistent glTF JSON that a viewer like Blender, Three.js, or any glTF-compatible renderer can load directly, with all atom colors, bond colors, and orbital surface colors exactly as they appeared in Avogadro.

Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

### DIFF
--- a/avogadro/rendering/gltfvisitor.cpp
+++ b/avogadro/rendering/gltfvisitor.cpp
@@ -1,0 +1,339 @@
+/******************************************************************************
+ This source file is part of the Avogadro project.
+******************************************************************************/
+
+#include "gltfvisitor.h"
+
+#include "ambientocclusionspheregeometry.h"
+#include "cylindergeometry.h"
+#include "linestripgeometry.h"
+#include "meshgeometry.h"
+#include "spheregeometry.h"
+
+#include <Eigen/Geometry>
+
+#include <cmath>
+#include <iostream>
+#include <sstream>
+
+namespace Avogadro::Rendering {
+
+using std::ostringstream;
+using std::string;
+
+GltfVisitor::GltfVisitor(const Camera& c)
+  : m_camera(c), m_opacity(1.0f)
+{
+}
+
+GltfVisitor::~GltfVisitor() {}
+
+void GltfVisitor::begin()
+{
+  // Clear all buffers for a fresh export
+  m_spheres.clear();
+  m_cylinders.clear();
+  m_meshVertices.clear();
+  m_meshColors.clear();
+  m_meshIndices.clear();
+}
+
+string GltfVisitor::end()
+{
+  // Build the final glTF JSON from collected geometry data
+  return buildJson();
+}
+
+void GltfVisitor::visit(Drawable&) {}
+
+void GltfVisitor::visit(SphereGeometry& geometry)
+{
+  // Collect all sphere data directly from the rendered scene geometry.
+  // This captures live color, radius, opacity, and position — including
+  // any user customizations — which FileFormat-based export misses.
+  for (const auto& s : geometry.spheres()) {
+    SphereData sd;
+    sd.center = s.center;
+    sd.radius = s.radius;
+    sd.color  = s.color;
+    m_spheres.push_back(sd);
+  }
+}
+
+void GltfVisitor::visit(AmbientOcclusionSphereGeometry& geometry)
+{
+  // Treat AO spheres the same as regular spheres for export purposes
+  for (const auto& s : geometry.spheres()) {
+    SphereData sd;
+    sd.center = s.center;
+    sd.radius = s.radius;
+    sd.color  = s.color;
+    m_spheres.push_back(sd);
+  }
+}
+
+void GltfVisitor::visit(CylinderGeometry& geometry)
+{
+  // Collect cylinder (bond) data with full scene color information
+  for (const auto& c : geometry.cylinders()) {
+    CylinderData cd;
+    cd.end1   = c.end1;
+    cd.end2   = c.end2;
+    cd.radius = c.radius;
+    cd.color  = c.color;
+    m_cylinders.push_back(cd);
+  }
+}
+
+void GltfVisitor::visit(MeshGeometry& geometry)
+{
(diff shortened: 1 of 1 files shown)


ghutchis:
These are not separate pull requests. I cannot review them independently. Please include the header as part of this PR so they can be reviewed together, thanks.

Thunder420Rock:
ok

ghutchis:
Also, you need to complete the DCO signoff with your name and permanent e-mail:
https://github.com/OpenChemistry/avogadrolibs/pull/2681/checks?check_run_id=66042690901

Thanks

PR #2695, opened 2026-03-09
Add support for reading from a filename (e.g., hd5 file)

ghutchis:
Currently used as:
file-mode = "read" or file-mode = ["read", "write"]

Default is stream-mode

Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

### DIFF
--- a/avogadro/io/fileformat.h
+++ b/avogadro/io/fileformat.h
@@ -142,7 +142,7 @@ public:
    * @param molecule The molecule the data will be read into.
    * @return True on success, false on failure.
    */
-  bool readFile(const std::string& fileName, Core::Molecule& molecule);
+  virtual bool readFile(const std::string& fileName, Core::Molecule& molecule);
 
   /**
    * @brief Write to the given @p fileName the contents of @p molecule.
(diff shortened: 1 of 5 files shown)


ghutchis:
For example: https://github.com/ghutchis/avogadro-mdtraj

Needs a bunch more testing, but I can at least see coordinates. 😄

matterhorn103:
I feel that that syntax is the wrong way round? 

How about making the possible options `"string" | "file"`, because then we can use the same set of options for all plugin feature types?

To riff on your "mode" idea, the keys could be `input-mode` and `output-mode`. Or for file formats they could be `read-mode` and `write-mode`, since that's maybe clearer.

But with the latter, I think then you'd want to change `input-format` to `write-format`, and `output-format` to `read-format`. Which I think would actually be a good thing, as I find the current terminology *super* confusing

ghutchis:
I don't know that I want to change `input-format` because that's used across all the plugin types.

But if you want to switch to something like:
```
input-mode: file
output-mode: string
```

That seems okay

PR #2703, opened 2026-03-12
Allow user-options metadata to be "dynamic" and call the script

ghutchis:
Script call adds `--user-options` and returns the JSON

Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

## Summary by CodeRabbit

* **New Features**
  * Dynamic user options: packages can load options from either scripts or static files; resulting options are applied cleanly.
  * Improved UI messaging: package scripts can surface warnings and errors that are shown to users.

* **Bug Fixes**
  * More robust loading: better handling of timeouts, execution failures and JSON parsing.
  * Avoids redundant nesting of userOptions when merging sources.

### DIFF
--- a/avogadro/qtgui/interfacescript.cpp
+++ b/avogadro/qtgui/interfacescript.cpp
@@ -308,21 +308,40 @@ bool InterfaceScript::processCommand(Core::Molecule* mol)
     }
 
     // check if there are messages for the user
-    if (obj.contains("message")) {
-      QString message;
-
-      if (obj["message"].isString())
-        message = obj["message"].toString();
-      else if (obj["message"].isArray()) {
-        QJsonArray messageList = obj["message"].toArray();
-        for (int i = 0; i < messageList.size(); ++i) {
-          if (messageList[i].isString())
-            message += messageList[i].toString() + "\n";
+    auto extractMessage = [](const QJsonValue& val) -> QString {
+      if (val.isString())
+        return val.toString();
+      if (val.isArray()) {
+        QString result;
+        const QJsonArray arr = val.toArray();
+        for (const auto& item : arr) {
+          if (item.isString())
+            result += item.toString() + '\n';
         }
+        return result;
       }
+      return {};
+    };
+
+    if (obj.contains("message")) {
+      QString message = extractMessage(obj["message"]);
       if (!message.isEmpty()) {
         QMessageBox::information(qobject_cast<QWidget*>(parent()),
                                  tr("%1 Message").arg(m_displayName), message);
+      }
+    }
+    if (obj.contains("warning")) {
+      QString message = extractMessage(obj["warning"]);
+      if (!message.isEmpty()) {
+        QMessageBox::warning(qobject_cast<QWidget*>(parent()),
+                             tr("%1 Warning").arg(m_displayName), message);
+      }
+    }
+    if (obj.contains("error")) {
+      QString message = extractMessage(obj["error"]);
+      if (!message.isEmpty()) {
+        QMessageBox::critical(qobject_cast<QWidget*>(parent()),
+                              tr("%1 Error").arg(m_displayName), message);
       }
     }
   }
--- a/avogadro/qtgui/packagemanager.cpp
+++ b/avogadro/qtgui/packagemanager.cpp
@@ -101,6 +101,30 @@ void PackageManager::mergeOptionsFromFile(QJsonObject& opts,
     opts.insert(it.key(), it.value());
 }
 
+// Locate an installed console script inside a pixi or venv environment.
+static QString findInstalledScript(const QString& packageDir,
+                                   const QString& scriptName, bool isPixi)
+{
+#ifdef Q_OS_WIN
+  const QString binDir =
+    packageDir + (isPixi ? QStringLiteral("/.pixi/envs/default/Scripts")
+                         : QStringLiteral("/.venv/Scripts"));
+  const QStringList exeSuffixes = { QStringLiteral(".exe"), QString() };
+#else
+  const QString binDir =
+    packageDir + (isPixi ? QStringLiteral("/.pixi/envs/default/bin")
+                         : QStringLiteral("/.venv/bin"));
+  const QStringList exeSuffixes = { QString() };
+#endif
+
+  for (const QString& suffix : exeSuffixes) {
+    const QString candidate = binDir + QLatin1Char('/') + scriptName + suffix;
+    if (QFileInfo(candidate).isExecutable())
+      return candidate;
+  }
+  return {};
+}
+
 QJsonObject PackageManager::loadOptionsFromScript(const QString& packageDir,
                                                   const QString& command)
 {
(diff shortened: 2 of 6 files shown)


ghutchis:
@coderabbitai review

ghutchis:
@coderabbitai review

avogadro/qtgui/interfacescript.cpp
```diff
@@ -308,23 +308,42 @@ bool InterfaceScript::processCommand(Core::Molecule* mol)
...
-        message = obj["message"].toString();
-      else if (obj["message"].isArray()) {
-        QJsonArray messageList = obj["message"].toArray();
-        for (int i = 0; i < messageList.size(); ++i) {
-          if (messageList[i].isString())
-            message += messageList[i].toString() + "\n";
+    auto extractMessage = [](const QJsonValue& val) -> QString {
+      if (val.isString())
+        return val.toString();
+      if (val.isArray()) {
```
matterhorn103, line 314: So I'm somewhat confused – this means a plugin can pass a message as either a string or an array of strings? What's the reason for this?

avogadro/qtgui/interfacescript.cpp
```diff
@@ -308,23 +308,42 @@ bool InterfaceScript::processCommand(Core::Molecule* mol)
...
+    };
+
+    if (obj.contains("message")) {
+      QString message = extractMessage(obj["message"]);
       if (!message.isEmpty()) {
         QMessageBox::information(qobject_cast<QWidget*>(parent()),
                                  tr("%1 Message").arg(m_displayName), message);
       }
     }
+    if (obj.contains("warning")) {
```
matterhorn103, line 333: Is there a risk that supporting a `warning` key would cause confusion with the `warnings` that input generators emit?

avogadro/qtgui/packagemanager.cpp
```diff
@@ -101,6 +101,118 @@ void PackageManager::mergeOptionsFromFile(QJsonObject& opts,
...
+  const QByteArray output = proc.readAllStandardOutput();
+  QJsonParseError err;
+  const QJsonDocument doc = QJsonDocument::fromJson(output, &err);
+  if (err.error != QJsonParseError::NoError) {
+    qWarning() << "PackageManager: failed to parse --user-options JSON from"
+               << command << ":" << err.errorString();
+    return {};
+  }
+
+  if (doc.isArray()) {
```
matterhorn103, line 186: Am I understanding correctly that this is just being tolerant of plugins returning incorrect input? So that if they return one of these:
```json
{
  "option1": {...}, "option2": {...}
}
# or, with tabs
[
  {"tabName": "tabA", "optionA1": {...}, "optionA2": {...}},
  {"tabName": "tabB", "optionB1": {...}}
]
```
it's treated as if they had returned the correct, nested form like:
```json
{
  "userOptions": {
    "option1": {...}, "option2": {...}
  }
}
# or, with tabs
{
  "userOptions": [
    {"tabName": "tabA", "optionA1": {...}, "optionA2": {...}},
    {"tabName": "tabB", "optionB1": {...}}
  ]
}
```

avogadro/qtgui/packagemanager.cpp
```diff
@@ -101,6 +101,118 @@ void PackageManager::mergeOptionsFromFile(QJsonObject& opts,
...
+  if (userOptionsValue == QLatin1String("dynamic"))
+    result = loadOptionsFromScript(packageDir, command);
+  else
+    result = loadOptionsFromFile(packageDir + '/' + userOptionsValue);
+
+  // Both loadOptionsFromFile() and loadOptionsFromScript() may wrap a bare
+  // array under "userOptions".  Unwrap so callers can insert under their own
+  // key without double-nesting.
+  if (result.contains(QStringLiteral("userOptions")))
+    return result.value(QStringLiteral("userOptions")).toObject();
```
matterhorn103: If the options specify a tabbed interface then `userOptions` will be an array, not an object – what will happen when `toObject()` is called in that case?

avogadro/qtgui/packagemanager.cpp
```diff
@@ -101,6 +101,118 @@ void PackageManager::mergeOptionsFromFile(QJsonObject& opts,
...
+  QJsonObject result;
+  if (userOptionsValue == QLatin1String("dynamic"))
+    result = loadOptionsFromScript(packageDir, command);
+  else
+    result = loadOptionsFromFile(packageDir + '/' + userOptionsValue);
+
+  // Both loadOptionsFromFile() and loadOptionsFromScript() may wrap a bare
+  // array under "userOptions".  Unwrap so callers can insert under their own
+  // key without double-nesting.
+  if (result.contains(QStringLiteral("userOptions")))
```
matterhorn103: I'd have thought rather than doing this here it'd be better for the two `loadOptions*` functions to always return the unnested version?

avogadro/qtgui/packagemanager.cpp
```diff
@@ -101,6 +101,118 @@ void PackageManager::mergeOptionsFromFile(QJsonObject& opts,
...
+                                                  const QString& command)
+{
+  // Locate pixi or the venv-installed script.
+  QString pixiExe = QStandardPaths::findExecutable(QStringLiteral("pixi"));
+  QProcess proc;
+  proc.setWorkingDirectory(packageDir);
+
+  if (!pixiExe.isEmpty()) {
+    proc.start(pixiExe, { QStringLiteral("run"), QStringLiteral("--as-is"),
+                          command, QStringLiteral("--user-options") });
```
matterhorn103: This is running the plugin without an argument for the feature!

matterhorn103 (commented):
Some things that stand out to me

matterhorn103:
I built from source using your branch and I've spent the past while trying to get it working with `avogadro-xtb`. It's not loading options properly when done dynamically, unfortunately.

I have verified the plugin is working correctly by running it directly on the command line independently of Avogadro.

The main thing that's going wrong is that Avogadro is attempting to get the user options by passing the flag for the plugin as a whole, running `avogadro-xtb --user-options`, without the feature as an argument. It should be running `avogadro-xtb config --user-options`.

Since this is an unrecognised argument, the plugin errors.

avogadro/qtgui/interfacescript.cpp
ghutchis: This is what the code has had for ages. The use case would be that the script doesn't just return after a message, but wants to put a few things together for the user.

avogadro/qtgui/interfacescript.cpp
ghutchis: I can remove it. I just figured if we have "message" and "error" that something intermediate would be useful.

avogadro/qtgui/packagemanager.cpp
ghutchis: Yes, this is fault-tolerant behavior.

I think you mean "we have changed the syntax a few times and some plugins may be using various forms of the syntax?"

IIRC I added it when you felt we should have non-nested syntax a few weeks ago.

avogadro/qtgui/packagemanager.cpp
ghutchis: That was the comment from CodeRabbit above. I'll submit a commit for this.

avogadro/qtgui/packagemanager.cpp
ghutchis: Less code duplication.

avogadro/qtgui/packagemanager.cpp
ghutchis: Sorry, I'll fix this in the next commit. Didn't notice it skipped the feature / identifier.

ghutchis:
I think this latest commit should solve the missing identifier - I forgot to pass it through all the way.

avogadro/qtgui/interfacescript.cpp
matterhorn103: I don't find it a bad idea, tbh. How about we make it `warnings` for all plugin types, make it an array, and it's just that they get presented differently in the case of input generators but otherwise work identically?

avogadro/qtgui/interfacescript.cpp
matterhorn103: Hmm OK. How about making it `messages` then and it's always an array, never a single string?

matterhorn103:
Based on your responses, I'm wondering if `messages`, `warnings`, `errors` would be better, and have them all be arrays of strings?

matterhorn103:
On second thoughts, maybe best to just merge as-is for now and we can revisit allowing the plural forms in future.

PR #2709, opened 2026-03-14
Add more syntax highlighting options and add support for dark mode.

brockdyer03:
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

## Summary by CodeRabbit

* **New Features**
  * Added dark mode-aware highlighting with adaptive color schemes that automatically adjust to your theme preference.
  * Enhanced highlighting colors with improved visual distinction across light and dark modes.

* **Refactor**
  * Modernized the highlighting system for more consistent and flexible color handling.

### DIFF
--- a/avogadro/molequeue/inputgenerator.cpp
+++ b/avogadro/molequeue/inputgenerator.cpp
@@ -659,45 +659,38 @@ bool InputGenerator::parseFormat(const QJsonObject& json,
 
     HighlightPreset presetType = mapStringToPreset[preset];
 
+    format.setFontFamily(QStringLiteral("mono"));
+
     /// @todo Store presets in a singleton that can be configured in the GUI,
     /// rather than hardcoding them.
     switch (presetType) {
       case HighlightPreset::Title:
-        format.setFontFamily(QStringLiteral("mono"));
         format.setFontWeight(QFont::Bold);
         format.setForeground(highlightColor("green", isDarkMode));
         return true;
       case HighlightPreset::Keyword:
-        format.setFontFamily(QStringLiteral("mono"));
         format.setForeground(highlightColor("blue", isDarkMode));
         return true;
       case HighlightPreset::Property:
-        format.setFontFamily(QStringLiteral("mono"));
         format.setForeground(highlightColor("cyan", isDarkMode));
         return true;
       case HighlightPreset::NumLiteral:
-        format.setFontFamily(QStringLiteral("mono"));
         format.setForeground(highlightColor("orange", isDarkMode));
         return true;
       case HighlightPreset::StrLiteral:
-        format.setFontFamily(QStringLiteral("mono"));
         format.setForeground(highlightColor("red", isDarkMode));
         return true;
       case HighlightPreset::Comment:
-        format.setFontFamily(QStringLiteral("mono"));
         format.setFontItalic(true);
         format.setForeground(highlightColor("gray", isDarkMode));
         return true;
       case HighlightPreset::Method:
-        format.setFontFamily(QStringLiteral("mono"));
         format.setForeground(highlightColor("purple", isDarkMode));
         return true;
       case HighlightPreset::Basis:
-        format.setFontFamily(QStringLiteral("mono"));
         format.setForeground(highlightColor("pink", isDarkMode));
         return true;
       case HighlightPreset::Block:
-        format.setFontFamily(QStringLiteral("mono"));
         format.setForeground(highlightColor("yellow", isDarkMode));
         return true;
     }


brockdyer03:
Requires https://github.com/OpenChemistry/avogenerators/pull/30
and https://github.com/OpenChemistry/avogadrolibs/pull/2710

avogadro/molequeue/inputgenerator.cpp
```diff
@@ -581,30 +650,78 @@ bool InputGenerator::parseFormat(const QJsonObject& json,
...
+      { QString("property"), HighlightPreset::Property },
+      { QString("num_literal"), HighlightPreset::NumLiteral },
+      { QString("str_literal"), HighlightPreset::StrLiteral },
+      { QString("comment"), HighlightPreset::Comment },
+      { QString("method"), HighlightPreset::Method },
+      { QString("basis"), HighlightPreset::Basis },
+    };
+
+    bool isDarkMode =
+      QGuiApplication::styleHints()->colorScheme() == Qt::ColorScheme::Dark;
```
ghutchis: I think I'd prefer if you switch to:
```
    bool darkMode = (defaultPalette.color(QPalette::WindowText).lightness() >
                     defaultPalette.color(QPalette::Window).lightness());
```

In principle, colorScheme() is okay, since Qt 6.5 is LTS but we haven't set a particular minimum Qt6 dependency previously.

The check for text lightness > window lightness is used pretty much everywhere in the codebase right now.

avogadro/molequeue/inputgenerator.cpp
```diff
@@ -570,6 +575,70 @@ bool InputGenerator::parseRules(const QJsonArray& json,
...
+};
+
+std::unordered_map<std::string, HighlightColor> mapStringToColor{
+  { "green", HighlightColor::green },   { "cyan", HighlightColor::cyan },
+  { "blue", HighlightColor::blue },     { "purple", HighlightColor::purple },
+  { "pink", HighlightColor::pink },     { "red", HighlightColor::red },
+  { "orange", HighlightColor::orange }, { "yellow", HighlightColor::yellow },
+};
+
+QColor textColor(const std::string colorName, bool darkMode = false)
```
ghutchis: Wouldn't it make more sense to have two map<string, QColor> structures, one for dark mode and one for light node? Then you don't even need this textColor() method?

ghutchis:
@coderabbitai review

avogadro/molequeue/inputgenerator.cpp
brockdyer03: The idea with this is that you don't need to have duplicate code anywhere. If I went with two maps I'd have to duplicate at least some level of code.

With this, it always passes through the same spot, and the color names are only ever detailed in one spot. Then the actual call just requires passing the color name and the `isDarkMode` boolean.

avogadro/molequeue/inputgenerator.cpp
brockdyer03: I can make that switch, but I think that this might be a good opportunity to set the minimum Qt6 dependency?

ghutchis: Right now, I'd rather do that after the 2.0 release. I haven't had time to see what kinds of Qt packages are out there on Linux distros.

avogadro/molequeue/inputgenerator.cpp
ghutchis: I'm definitely suggesting less code, e.g.
```
// Dark-mode palette: brighter / higher-value colors for contrast on dark
// backgrounds.
static const std::map<QString, QColor> darkModeColors = {
  { "red",    QColorConstants::Svg::crimson     },
  { "orange", QColorConstants::Svg::coral       },
  { "yellow", QColorConstants::Svg::goldenrod   },
  { "green",  QColorConstants::Svg::springgreen },
  { "cyan",   QColorConstants::Svg::cyan        },
  { "blue",   QColorConstants::Svg::dodgerblue  },
  { "purple", QColorConstants::Svg::mediumpurple},
  { "pink",   QColorConstants::Svg::orchid      },
  { "white",  QColorConstants::Svg::whitesmoke  },
  { "gray",   QColorConstants::Svg::darkgray    },
};
 
// Light-mode palette: deeper / more saturated colors for contrast on light
// backgrounds.
static const std::map<QString, QColor> lightModeColors = {
  { "red",    QColorConstants::Svg::firebrick    },
  { "orange", QColorConstants::Svg::darkorange   },
  { "yellow", QColorConstants::Svg::darkgoldenrod },
  { "green",  QColorConstants::Svg::limegreen    },
  { "cyan",   QColorConstants::Svg::darkcyan     },
  { "blue",   QColorConstants::Svg::blue         },
  { "purple", QColorConstants::Svg::blueviolet   },
  { "pink",   QColorConstants::Svg::fuchsia      },
  { "white",  QColorConstants::Svg::black        },
  { "gray",   QColorConstants::Svg::dimgray      },
};
 
/// Return the appropriate color map for the current mode.
inline const std::map<QString, QColor>& colorScheme(bool isDarkMode)
{
  return isDarkMode ? darkModeColors : lightModeColors;
}

// Or something like this:
inline const QColor highlightColor(const string& name, bool isDarkMode)
{
  return isDarkMode ? darkModeColors[name] : lightModeColors[name];
}
```

ghutchis: For something like this, I don't think there's a meaningful difference between std::map, QMap, and std::unordered_map.

ghutchis:
It's not a big thing, but I hope you'd switch to the string -> constant QColor map. Right now you're going through a few levels of indirection when you simply want to map from a string to a color. Thanks.

brockdyer03:
Is there any particular reason that the lightmode colors were changed?

brockdyer03:
Since I switched the `comment` preset to use gray instead of yellow, I realized it might be good to add a preset for `block` types. In ORCA this would be the equivalent of
```
%method
    keyword str_literal
end
```
where `%method` and `end` are `block`.

ghutchis:
> Is there any particular reason that the lightmode colors were changed?

Not particularly. I organized by ROYGBIV and then auto-complete was filling in some color names. So if they didn't match exactly, I'll blame that.

ghutchis:
The one thing that might be useful - after this is merged - is to check some of the preset colors against a colorblind-safe palette. Mainly to be sure that in dark-mode nothing will end up being hard to read for anyone colorblind. But that's definitely a separate PR.

avogadro/molequeue/inputgenerator.cpp
```diff
@@ -581,30 +622,85 @@ bool InputGenerator::parseFormat(const QJsonObject& json,
...
     }
-    return true;
+
+    HighlightPreset presetType = mapStringToPreset[preset];
+
+    /// @todo Store presets in a singleton that can be configured in the GUI,
+    /// rather than hardcoding them.
+    switch (presetType) {
+      case HighlightPreset::Title:
+        format.setFontFamily(QStringLiteral("mono"));
```
ghutchis: If they're all "mono" you could pull this out of the switch statement.

brockdyer03:
> The one thing that might be useful - after this is merged - is to check some of the preset colors against a colorblind-safe palette.

I actually have already done this:

Full color vision:
light_background

Protanopia:

light_bg_protanopia

Deuteranopia:

light_bg_deuteranopia

Tritanopia:
light_bg_tritanopia


Full color vision:
dark_background

Protanopia:
dark_bg_protanopia

Deuteranopia:
dark_bg_deuteranopia

Tritanopia:
dark_bg_tritanopia

ghutchis (approved):

PR #2714, opened 2026-03-15
Change the way in which tabs are specified in user options

matterhorn103:
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

### DIFF
--- a/avogadro/qtgui/jsonwidget.cpp
+++ b/avogadro/qtgui/jsonwidget.cpp
@@ -71,7 +71,8 @@ void JsonWidget::setMolecule(QtGui::Molecule* mol)
         QJsonObject inputParameters = doc.object();
         // check for a few known keys
         if (inputParameters.contains(u"processors"_s))
-          setOption(u"Processor Cores"_s, inputParameters[u"processors"_s].toInt());
+          setOption(u"Processor Cores"_s,
+                    inputParameters[u"processors"_s].toInt());
         else if (inputParameters.contains(u"memory"_s))
           setOption(u"Memory"_s, inputParameters[u"memory"_s].toInt());
         else if (inputParameters.contains(u"basis"_s))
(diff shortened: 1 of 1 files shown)


matterhorn103:
In the end using schema 3 (see [forum thread](https://discuss.avogadro.cc/t/useroptions-and-highlightstyles/7317/16) led to a much simpler implementation, as then the options are always stored in a flat array of the options, and it matches the way that the `JsonWidget` class works much better.

However, the penultimate commit, 33fe5f0d0fca458307dcfa9c870a07e684a9fe9c, implemented schema 2, and it also works.

ghutchis:
Just to check - did you run clang-format?

avogadro/qtgui/jsonwidget.cpp
```diff
@@ -664,54 +653,22 @@ QWidget* JsonWidget::createTableWidget(const QJsonObject& obj)
 
 void JsonWidget::setOptionDefaults()
 {
-  if (!m_options.contains(QStringLiteral("userOptions"))) {
+  if (!m_options.contains("userOptions")) {
```
ghutchis: No, you definitely want to keep QStringLiteral as much as possible. Otherwise the "" bits end up being created on the fly.

ghutchis: If anything, we should be using *more* QStringLiteral throughout the code.

matterhorn103:
Oops, no, sorry. I'll get round to that in a bit.

I'd appreciate it if you looked at what I've done somewhat carefully – it works but I don't trust it to be robust.

Could you maybe get the CodeRabbit review thing to run?

ghutchis (commented):
Minor comment - looks good.

avogadro/qtgui/jsonwidget.cpp
matterhorn103: Happy to oblige 😉

ghutchis:
Looks good, thanks.

PR #2716, opened 2026-03-16
Dynamic user options: Pass {} as input, and fix hang by sending EOF

matterhorn103:
Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

### DIFF
--- a/avogadro/qtgui/packagemanager.cpp
+++ b/avogadro/qtgui/packagemanager.cpp
@@ -20,6 +20,8 @@
 #include <QtCore/QJsonParseError>
 #include <QtCore/QSettings>
 
+using namespace Qt::StringLiterals;
+
 namespace Avogadro::QtGui {
 
 PackageManager::PackageManager(QObject* parent) : QObject(parent) {}
@@ -158,6 +160,13 @@ QJsonObject PackageManager::loadOptionsFromScript(const QString& packageDir,
     }
     proc.start(scriptExe, userOptsArgs);
   }
+
+  // Plugins may always expect some valid JSON as input over stdin, even if it's
+  // just an empty object
+  QByteArray scriptStdin = "{}"_ba;
+  proc.write(scriptStdin);
+  // Python
+  proc.closeWriteChannel();
 
   constexpr int timeoutMs = 30000; // 30 seconds
   if (!proc.waitForStarted(timeoutMs)) {


matterhorn103:
This fixes the remaining bug that I have in the port of `avogadro-xtb`.

There's two parts to this, despite the small diff:

### Make sure to always pass an input JSON

It's been agreed that, for consistency, Avogadro should _always_ pass some JSON object to a plugin on `stdin`, even if that's just an empty `{}`. This makes life much more convenient for plugin authors, as they can then just do:

```py
# Read (initial) input from Avogadro
avo_input = json.loads(sys.stdin.read())
```

when run, regardless of what the feature/request is. Otherwise, because `json.loads()` errors when it's passed nothing, they'd have to make the read conditional on either the arguments:

```py
if not(args.user_options or args.whatever or ...):
    avo_input = json.loads(sys.stdin.read())
```

or the content received on `stdin`:

```py
# Read (initial) input from Avogadro, if there is any
stdin_str = sys.stdin.read()
if stdin_str:
    avo_input =  json.loads(stdin_str)
else:
    avo_input = {}
```

That seems to me to be a subtle requirement for a plugin to implement that'd be bound to trip up many plugin authors.

### Make sure to send EOF

`sys.stdin.read()` will continue to read until Avogadro signals EOF. `PythonScript::asyncExecute` does this after writing the input by running `m_process.closeWriteChannel()` at the end (though only if it has been told to via a boolean argument). `PackageManager::loadOptionsFromScript` didn't do this, so when run with `--user-options` the plugin hangs at the input reading line until Avogadro terminates it.

The PR fixes it simply by adding a `closeWriteChannel()` call.

Something to consider is whether it'd make more sense to send a newline character to plugins immediately after the JSON input, and we then encourage plugins to call `sys.stdin.readline()` rather than `sys.stdin.read()`. Then the plugins aren't dependent on Avogadro sending EOF. I'm assuming that energy calculator plugins, that need to send more on `stdin` after the initial payload, muck about with opening, closing, and reopening the write channel? Or do they just require plugins to use `readline()` for the initial JSON rather than `read()`?

_**The other solution**_ is to say that plugins should always use `readline()` to read the input *anyway*, which will work regardless of whether EOF or `\n` follows the JSON, and to say that part of the plugin contract is that Avogadro guarantees to send *either* EOF *or* `\n` after the input JSON but that it can't guarantee which.

In that case this bug could also be fixed by adding `\n` to the input rather than calling `closeWriteChannel()`, but one of the two is necessary in any case.

matterhorn103:
It also isn't clear to me why `PackageManager` runs the plugin itself in this case rather than reusing e.g. `PythonScript::asyncExecute`?

ghutchis:
> It also isn't clear to me why `PackageManager` runs the plugin itself in this case rather than reusing e.g. `PythonScript::asyncExecute`?

If you want to send it to `PythonScript` that's okay. To me, it seemed to make more sense for `PackageManager` to handle "track all the metadata needed for the script`.

matterhorn103:
I meant just for the actual Python process execution step. I don't have a strong opinion at all, it just surprised me, but I don't really have a great understanding of how all the different parts fit together and what relies on what.

ghutchis:
It'd be fine. If this is blocking the `xtb` plugin, let's merge this and we can talk about whether it should pass off the invocation to PythonScript.

PR #2735, opened 2026-03-18
If a generator doesn't output a highlight style, use "default"

### LINKED ISSUE #2727: Use `default` highlight style if it is not specified in the plugin output
### Proposal

Currently if a plugin author doesn't include the `highlightStyles` key in the output of their plugin there is no syntax highlighting in their dialog box. To me this seems a bit redundant, since the majority of plugins only actually have 1 highlight style, and it conveniently is labeled `default` in every plugin.

It would be nice if there was some logic in the syntax highlighting that was along the lines of

1. Check plugin output JSON for `highlightStyles`.
2. If there is a key for that, use whatever it specifies.
3. Otherwise, if there is nothing specified but there is a highlight style called `default` in the `syntax.(?:json|toml)` file, then use that as the default.

ghutchis:
Fix #2727 

Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

### DIFF
--- a/avogadro/molequeue/inputgenerator.cpp
+++ b/avogadro/molequeue/inputgenerator.cpp
@@ -258,6 +258,13 @@ bool InputGenerator::generateInput(const QJsonObject& options_,
                   m_fileHighlighters[fileName] = highlighter;
                 else
                   highlighter->deleteLater();
+              } else if (m_highlightStyles.contains("default")) {
+                auto* highlighter(new GenericHighlighter(this));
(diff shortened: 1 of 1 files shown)


ghutchis:
@brockdyer03 - let me know if this works for you...

brockdyer03:
Tested, this does work.
LGTM.

PR #2868, opened 2026-05-31
Initial WebAssembly build

timvdm:


timvdm:
See https://github.com/OpenChemistry/openchemistry/pull/190

PR #2894, opened 2026-07-12
Force UTF-8 mode for Windows Python scripts

### LINKED ISSUE #1618: Plugin output to console doesn't use Unicode on Windows, Avogadro expects UTF-8
See [this issue](https://github.com/matterhorn103/avo_xtb/issues/2) on the avo_xtb plugin repo, helpfully raised by @Ineluki80.

It seems that on Windows, in at least some cases, the Python stdout stream isn't being encoded in UTF-8 but rather that of the system locale/extended ASCII, which means Unicode characters (e.g. the ellipsis for menu options) are being garbled.

The Python `sys` [docs](https://docs.python.org/3/library/sys.html) suggest that actually in many cases, UTF-8 is the default, so I'm not quite sure what's going on. In any case, it looks like there are various ways to enforce Unicode when calling the interpreter e.g. via an environment variable or a command line flag.

aouxwoux:
For Windows launches, set `PYTHONUTF8=1` on the process environment in the shared `PythonScript` launch path. This makes direct Python and pixi-launched plugin scripts use UTF-8 consistently, avoiding system-locale decoding of plugin output.

Adds a Windows regression test that launches a temporary Python script and verifies `sys.flags.utf8_mode` is enabled.

Validation:
- `git diff --check`
- Confirmed locally that `PYTHONUTF8=1` enables Python UTF-8 mode
- Full CMake/Qt test suite was not run locally because this environment lacks CMake and Qt

Fixes #1618.

### DIFF
--- a/avogadro/qtgui/pythonscript.cpp
+++ b/avogadro/qtgui/pythonscript.cpp
@@ -13,6 +13,7 @@
 #include <QtCore/QFileInfo>
 #include <QtCore/QLocale>
 #include <QtCore/QProcess>
+#include <QtCore/QProcessEnvironment>
 #include <QtCore/QSettings>
 #include <QtCore/QStandardPaths>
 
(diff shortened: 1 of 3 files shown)


ghutchis:
Hi - as part of our policy, we require commits to be signed by full names and permanent e-mail addresses. This allows us to contact contributors, e.g. for manuscripts about Avogadro development. This commit is currently signed as:

aouxwoux <81102668+aouxwoux@users.noreply.github.com>

Which does neither. Please include a full name and permanent email address. Thanks.

ghutchis:
Great - apologies for the delay. LGTM - many thanks.

PR #2993, opened 2026-09-11
Allow drag-and-drop to reorder atoms in the property table

### LINKED ISSUE #2901: Reordering Atom Index Number
**Is your feature request related to a problem? Please describe.**
I think it would be helpful if Avogadro had a feature to allow the user to renumber the atoms, i.e. change/reorder the index of an atom. The order/index number of an atom becomes important for things like transition state calculations when the reactant atoms order must match the product atoms order. But if you have to separately optimize parts of a molecule that is separated in a reaction, then put them back together for the product file, sometimes the atom index changes. Currently to correct this you need to manually reorder the atoms lines in the .xyz or .inp files.

**Describe the solution you'd like**
Since there is already separate labels for Unique ID and Index, you could use this for a unique identifier that wouldn't change and the value that does change when you apply this feature. The feature could be implemented as part of the 'atom properties' window or a similar window that would show unique ID, Index, Element, X, Y, and Z columns (columns for valence, formal charge, label, color, etc. may also be helpful so the user can identify they are changing the correct atom) and highlight the atom when the row is selected. Then you could drag the selected row up or down to reorder. The Unique ID would stay the same, but the Index would change to reflect the new order the rows are in (dragged row takes on the new position in the table, all rows below it increase index number by 1). Since you want to keep the same total number of atoms before and after the change, I suggest a reordering instead of a write in number. 

**Describe alternatives you've considered**
An alternative could be that feature is a write in edit of the index number when looking at the molecule with index label on. However, you would have to put restrictions on what input was possible, e.g. </= max index number, and if the index number is repeated, the other atom label gets an * on the label so the user knows they have to change it, or like the reordering all atoms with index number >/= the change would be increased by 1. This might get complicated for the user depending on how you implement the restriction on index numbers repeating. If all numbers change, the user might find that annoying if they just had to switch a few things around. 

**Additional context**
This is mostly a problem when I remove atoms from the frame to optimize a part of a molecule, then add atoms back for the final complete frame before a NEB-TS calculation. Sometimes if I don't completely remove the atoms where a bond is broken, they will be attached somewhere else after a geometry optimization. So I have found it easier to remove completely so I can get a converged geometry.

I can open a topic on the discussion forum if you think it is necessary.

ghutchis:
Shows an indicator for where the row will "drop"
Fixes #2901

Version 1.1

1 Letterman Drive
Suite D4700
San Francisco, CA, 94129

    indicated in the file; or

    in the file; or

    it.

## Summary by CodeRabbit

* **New Features**
  * Added drag-and-drop reordering for atom rows in the property table.
  * Reordering is available when the table uses its natural, unfiltered order.
  * Atom moves preserve the current selection and support undo.
  * Added automatic table refresh after atoms are reordered.

### DIFF
--- a/avogadro/qtgui/molecule.h
+++ b/avogadro/qtgui/molecule.h
@@ -91,7 +91,14 @@ public:
     Added = 1024,
     Removed = 2048,
     Modified = 4096,
-    Moved = 8192
+    Moved = 8192,
+    /**
+     * Indicates atom indices were permuted (see RWMolecule::reorderAtoms).
+     * Same atoms, same bonds, same counts -- only the numbering changed, so
+     * listeners that detect structural change by comparing counts will miss
+     * it unless they check for this flag.
+     */
+    Reordered = 16384
   };
   Q_DECLARE_FLAGS(MoleculeChanges, MoleculeChange)
 
--- a/avogadro/qtgui/rwmolecule_undo.h
+++ b/avogadro/qtgui/rwmolecule_undo.h
@@ -154,6 +154,11 @@ namespace {
 // permutation rather than as the permutation itself. A transposition is its
 // own inverse, so undoing is the same sequence walked backwards, and there is
 // no second permutation to keep consistent with the first.
+//
+// Unlike the other commands in this file, this one notifies directly instead
+// of relying on the blanket Atoms | Added that MainWindow emits after an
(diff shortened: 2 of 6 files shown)


ghutchis:
@coderabbitai review

PR #3018, opened 2026-09-18
Fix the clang-tidy workflow and drop the duplicate build

ghutchis:
The clang-tidy job has been failing at the first step. It uses the local composite action `./.github/actions/checkout-repositories` without ever running `actions/checkout`, so the action file is not on disk yet:

> Can't find 'action.yml', 'action.yaml' or 'Dockerfile' under `/home/runner/work/avogadrolibs/avogadrolibs/.github/actions/checkout-repositories`. Did you forget to run actions/checkout before running your local action?

(The 4h07m run time on the last attempt was the un-timeouted `action-tmate` step idling until it was cancelled.)

## Sync with `build_linux.yml`

- Check out the workflow repository before using the local composite action.
- Configure against `$GITHUB_WORKSPACE/openchemistry` — `checkout-repositories` nests every repo under `openchemistry/`, so the old `$GITHUB_WORKSPACE` pointed at avogadrolibs rather than the superbuild.
- Fix the Analyze working directory, which was `${{ runner.workspace }}/avogadrolibs` — a path that is never created — plus a stray extra `cd avogadrolibs`.
- Install `zlib1g-dev libbz2-dev liblzma-dev libzstd-dev` (needed since the compressed I/O work) and add `-DUSE_SYSTEM_EIGEN=TRUE` to match Linux.
- Drop `-DUSE_VTK=ON` from `FEATURES`; VTK is never installed on the runner.
- Use `cmake --build` rather than bare `ninja`, matching the other workflows.

## Build avogadrolibs once instead of twice

`CMAKE_EXPORT_COMPILE_COMMANDS` is not forwarded to the avogadrolibs `ExternalProject` — `cmake/External_avogadrolibs.cmake` passes only `OpenChemistry_DEFAULT_ARGS` / `OpenChemistry_THIRDPARTYLIBS_ARGS`, which carry just `BUILD_SHARED_LIBS`, the prefix paths and `CMAKE_BUILD_TYPE`. So the `-DCMAKE_EXPORT_COMPILE_COMMANDS=1` cache argument on the superbuild did nothing, which is why the job needed `bear` plus a `ninja clean` and a second full compile of avogadrolibs at `-O2`.

CMake reads that variable from the *environment* as well (3.17+), and `ExternalProject` inherits it, so setting it job-level reaches the sub-project. A single build now writes `build/avogadrolibs/compile_commands.json`, `bear` is gone, and the clean-and-rebuild is gone with it. The generated file is also already scoped to avogadrolibs, which is what the clean/rebuild was approximating. Verified with a minimal superbuild that the environment variable does propagate through `ExternalProject_Add`.

## Analysis scope and hygiene

- Restrict `run-clang-tidy` to `avogadro/` and set a matching `-header-filter`, so the generated patch can never touch the bundled thirdparty sources (`pugixml`, `struct`, `libgwavi` all compile into `compile_commands.json`) or AUTOMOC output.
- Tolerate `run-clang-tidy`'s non-zero exit, which it returns whenever any check fires — the normal case here, and it would otherwise abort before the patch is written.
- Gate the PR metadata step on `pull_request`; on `workflow_dispatch` it wrote three empty files.
- Give `action-tmate` a `timeout-minutes`, and add a `concurrency` group.

Note that the build type is nearly irrelevant to the analysis itself — clang-tidy never runs the backend, so `-O2`/`-g` do not affect diagnostics. The build is required only because the sources will not parse without the generated ones (AUTOMOC output, `ui_*.h`, `generate_export_header` headers) and the installed thirdparty headers. `RelWithDebInfo` is kept so `NDEBUG` matches what ships.

## Before merging

`on:` is temporarily `[push, pull_request, workflow_dispatch]` to validate the workflow here. **This must go back to `workflow_dispatch` only before merge** — there is a `TODO` on the line.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


ghutchis:
If there's any question of whether a pull request is generated via Claude... you can see the message it generated.

In this case, I'm going to argue it's okay because:
- it's fixing a GitHub workflow, not anything involving chemistry
- I've reviewed the actual changes and it looks okay

I'm still *very* reluctant to allow fully-generated pull requests.

PR #3025, opened 2026-09-19
Read Molden Occup as a double and lround into the electron count

### LINKED ISSUE #3005: Molden reader truncates fractional Occup values — electron count (and HOMO/LUMO, occupied set, density) wrong for localized/natural-orbital files
**Summary.** For a Molden file whose `Occup` values are fractional — localized-orbital exports (NBO/JANPA CLPO/LHO/LMO), MP2 natural orbitals, any non-Aufbau density — the Molden reader reports the wrong electron count. `avogadro/quantumio/molden.cpp:242` accumulates each occupancy as an **integer**:

```cpp
if (Core::contains(line, "Occup"))
  m_electrons += Core::lexicalCast<int>(list.back()).value_or(0);
```

so `Occup= 1.986000` contributes 1 and `Occup= 0.500000` contributes 0. The truncated sum is committed as the electron count at `molden.cpp:537` (`basis->setElectronCount(m_electrons)`).

This is distinct from #2722 (no reader populates `m_moOccupancy` yet): the *count itself* is wrong, and it feeds code paths beyond the orbital table.

**Symptom.** A 10-electron water file with five occupied MOs at `Occup= 1.997086` reports **5 electrons**. User-visible knock-on effects:

- `BasisSet::lumo()` (`avogadro/core/basisset.h:245`) finds no occupancy data (the Molden reader never fills `m_moOccupancy`), so it falls back to `m_electrons[0] / 2 + …` (`basisset.h:268`); `homo()` is defined as `lumo() - 1` (`basisset.h:112`) — so HOMO/LUMO labels are wrong.
- `avogadro/core/gaussiansettools.cpp` builds the occupied-MO list for surfaces/cubes by preferring occupancy data (empty for these files, `:802`) and otherwise taking the first `electronCount() / 2` MOs at occupancy 2.0 (`:814`) — a wrong occupied set.
- `avogadro/core/gaussianset.cpp:575` builds the RHF density matrix over the first `m_electrons[0] / 2` MOs — with the truncated count only 2 of water's 5 occupied orbitals contribute, so electron-density output from such files is built from the wrong occupied set as well.
- The orbital table (`avogadro/qtplugins/surfaces/orbitaltablemodel.cpp:351`) reads `moOccupancy(Paired)`, gets nothing, and falls back to energy/position heuristics — which cannot help files whose `Ene=` values are placeholders (raw JANPA exports write 0, 1, 2, …): "occupied" there is pure file position. The attached repro carries real `Ene=` values, which isolates the count bug itself.
- Canonical SCF files (`Occup= 2.000000` exactly) are unaffected, and `tests/quantumio/moldentest.cpp` only uses integral occupancies — which is why this isn't caught by the suite.

**Repro.** Open a localized-orbital Molden export with fractional occupations. Attached in a comment: [`water_CLPO_Alpha.molden.txt`](https://github.com/user-attachments/files/32241973/water_CLPO_Alpha.molden.txt) (the `.molden` extension renamed `.molden.txt` for GitHub's upload filter) — exactly what our ORCA → JANPA pipeline emits by default (`orca_to_janpa.py water --clpo`; HF/def2-SVP, 10 electrons, five occupied MOs at `Occup= 1.997…`): cartesian d/f, `Spin= Alpha`, real `Ene=` values, and the `Occup` values exactly as JANPA computed them. The electron count reads **5 instead of 10**, and the occupied set is mislabelled. Same behaviour for any JANPA export — all six orbital families our pipeline produces (CLPO/LHO/AHO/LPO/NAO/PNAO) have fractional occupations; e.g. ethene, 16 electrons → reports 8. We currently work around this by rewriting `Occup` to integers 2/0 behind an optional flag — a display-level hack that discards the true occupancies and must be re-applied per file; this report asks for the real fix.

**Related:** #2722 (File import for QM codes need to read orbital occupations — the Molden checkbox), #2627 (closed; ORCA open-shell occupancy display).

**Fix direction (for discussion).**

1. Parse `Occup` as a real number (`lexicalCast<double>`) and accumulate that — this alone fixes `electronCount()` and every consumer above.
2. To complete the Molden item of #2722, populate `m_moOccupancy` from the file. Note `BasisSet::m_moOccupancy` is `std::vector<unsigned char>` (`basisset.h:242`), so fractional occupations cannot be represented; a real-valued type (float/double) would be needed for localized/natural-orbital files. If the integer type is a deliberate choice for the table's occupation indicators, then at minimum round instead of truncate, and keep the double-precision total separately.

Happy to test a patch; can supply more example files (same molecule as fractional / integer-workaround / raw JANPA exports).

shaneraphel:
Molden `Occup` is accumulated with `lexicalCast<int>`. Fractional occupations (localized NBO/JANPA, MP2 natural orbitals) truncate: 1.986 → 1, 0.500 → 0. That truncated sum is passed to `setElectronCount`, so HOMO/LUMO and the occupied set (`electrons/2`) are built from the wrong integer. A 10-electron water CLPO file with five `Occup= 1.997` MOs reports 5 electrons.

Occup is now read as `lexicalCast<double>` (same as `Ene`) and added with `lround` into `m_electrons`. Canonical `Occup= 2.000` is unchanged. `m_moOccupancy` stays `unsigned char`.

Fixes #3005


ghutchis:
I don't understand this at all. The commit message seems purely LLM-generated.

ghutchis:
You have not signed off the DCO correctly with a full legal name and permanent e-mail address: https://github.com/OpenChemistry/avogadrolibs/pull/3025/checks?check_run_id=105961992271

And I do not understand why the PR was created by a different GitHub user than the contribution.

avogadro/quantumio/molden.cpp
```diff
@@ -242,7 +243,8 @@ void MoldenFile::processLine(std::istream& in)
         // Parse the occupation, spin, energy, etc (Occup, Spin, Ene).
         while (!line.empty() && Core::contains(line, "=")) {
           if (Core::contains(line, "Occup"))
-            m_electrons += Core::lexicalCast<int>(list.back()).value_or(0);
+            m_electrons += static_cast<int>(std::lround(
```
ghutchis: Why would you use `lround` and then cast to an int?

ghutchis (changes requested):
This should be a relatively trivial patch, but a lot of this doesn't make much sense. Please confirm you're a human and take care of the proper rounding.

shaneraphel:
thanks for review,i am human,this lround is my coding system bug and not LLM-generated.i will fix the DCO.

shaneraphel:
I use the git/local system push.

avogadro/quantumio/molden.cpp
shaneraphel: You're right.

shaneraphel:
fixed.thanks for review.

